PR #151, opened 2014-01-08
bugfix: notebook list editor works with extended name notation

aestrivex:
ListEditor in notebook style crashes with the following error whenever the model object is given in extended name notation:

traits.trait_errors.TraitError: Trait notification dispatch type 'ui' is not compatible with handler signature and extended trait name notification style

A search through old emails reveals that this bug was found in 2009, but the fix was never propagated to the notebook editor.


corranwebster:
LGTM.

PR #153, opened 2014-01-17
Fix for _popEventHandlers deleting handlers not added by traitsui

robmcmullen:
Fix for problem reported in the mailing list where there was a crash in the DateEditor after a simple configure_traits test.


corranwebster:
Closed in favour of #215 because of extra commit.

PR #182, opened 2014-10-07
[WIP] Feature/advanced tuple editor

itziakos:
This PR augments the TupleEditor to support extra validation provided by the user in the Editor factory or defined in the trait (see ValidatedTuple PR #https://github.com/enthought/traits/pull/205).

example:

``` python
>>> from traits.api import Int, HasStrictTraits, Tuple
>>> from traitsui.api import View, Item, TupleEditor
>>> class Model(HasStrictTraits):
...     value_range = Tuple(Int(0), Int(1))
...     view = View(Item('value_range', editor=TupleEditor(fvalidate=lambda x: x[0] < x[1])))
... 
>>> model = Model()
>>> model.configure_traits()
>>> 
```

image
**normal values**

image
**invalid field**

image
**invalid tuple** (x[0] > x[1])

**note**: Tests require #https://github.com/enthought/traits/pull/205


corranwebster:
I think this probably needs to wait for a traits release containing the `ValidatedTuple` before its safe to merge this.

itziakos:
@corranwebster  this is now ready to review

mdickinson:
I wonder whether both fields should have a red background in the case that the pair of values fails validation. It's not obvious right now why one field is red but the other not. With the example from the initial comment (and `validation` replaced by `fvalidate`), I can achieve all three of the following states (the first text box is active in all three cases).

screen shot 2016-06-07 at 09 40 36
screen shot 2016-06-07 at 09 40 52
screen shot 2016-06-07 at 09 41 01

traitsui/editors/tuple_editor.py
```diff
@@ -73,9 +58,10 @@ class ToolkitEditorFactory(EditorFactory):
     # 'enter_set' metadata or an editor defined.
     enter_set = Bool(False)
 
-#-------------------------------------------------------------------------
-#  'SimpleEditor' class:
-#-------------------------------------------------------------------------
+    # The validation function to use for the Tuple. If the edited trait offers
+    # already a validation function then the value of this trait will be
```
corranwebster: This should be the other way around (unless you have a particular use-case in mind).  Because one `HasTraits` class can potentially have many different views, you generally want thing specified in the editor to override things specified in traits.

Although not exactly analogous, compare with (note that the `label` in the `Item` wins):
```
from traits.api import HasTraits, Float


class Test(HasTraits):

    x = Float(label='Variable x')

    view = View(Item('x', label='Var x'))

t = Test()
t.edit_traits()
```

itziakos: As you say it is not analogous. Since there is no restriction on the label. However if the Editor validation overrides the trait validation it can potentially result in an invalid value been forced into a trait (i.e. a ValidatedTuple trait) which will throw an exception.

itziakos: Thinking about it a little more, maybe a compromise would be the union of the trait validation and editor validation

traitsui/editors/tuple_editor.py
```diff
@@ -216,21 +198,27 @@ def __init__(self, editor):
...
+        value = editor.value
         index = int(name[1:])
         value = self.editor.value
         if new != value[index]:
-            self.editor.value = tuple(
-                [getattr(self, 'f%d' % i) for i in range(self.fields)])
+            new_value = tuple(
+                getattr(self, 'f{0}'.format(i)) for i in range(self.fields))
+            if self.fvalidate is not None:
+                if self.fvalidate(new_value):
```
itziakos: I am not convinced. If you want the validation to change the value then use a VaidatedTuple trait.  A number of editor already provide a format function to support conversions.

traitsui/editors/tuple_editor.py
```diff
@@ -73,9 +58,10 @@ class ToolkitEditorFactory(EditorFactory):
     # 'enter_set' metadata or an editor defined.
     enter_set = Bool(False)
 
-#-------------------------------------------------------------------------
-#  'SimpleEditor' class:
-#-------------------------------------------------------------------------
+    # The validation function to use for the Tuple. If the edited trait offers
+    # already a validation function then the value of this trait will be
+    # ignored.
+    fvalidate = Callable
```
itziakos: It has been a while... So initially I had it a validator but then I changed it to fvalidate to be similar to the Property fget, fset, fvalidate

traitsui/editors/tuple_editor.py
itziakos: Also the fact that the validate function of some traits changes the actual changed value it an internal implementation/optimization. For example the Property trait has separate functions for get, set and validation.

corranwebster:
I am tempted to merge this as-is.  Nothing above is a show-stopper, and it would be good to have this finished.

itziakos:
@corranwebster, this PR still needs so fixes related to the ui behavior.

PR #195, opened 2015-02-24
Workaround for PySide table alignment bug.

### LINKED ISSUE #194: TabularEditor text alignment doesn't work under PySide
See https://bugreports.qt.io/browse/PYSIDE-20 for the bug and workaround.

We should probably include this workaround as long as it doesn't break under PyQt.

corranwebster:
Fixes #194 

See https://bugreports.qt.io/browse/PYSIDE-20 for a discussion of the issue and workaround.


rkern:
LGTM!

PR #196, opened 2015-02-25
DataFrameEditor

corranwebster:
This is an editor for a Pandas DataFrame built on top of the `TabularEditor` in a similar way to the `ArrayViewEditor`.


traitsui/tests/ui_editors/test_data_frame_editor.py
```diff
@@ -0,0 +1,287 @@
...
+    raise nose.SkipTest
+
+from traits.api import HasTraits, Instance
+
+from traitsui.item import Item
+from traitsui.ui_editors.data_frame_editor import (
+    DataFrameEditor, DataFrameAdapter)
+from traitsui.view import View
+
+from traitsui.tests._tools import *
```
jonathanrocher: Would replace with what is used.

traitsui/tests/ui_editors/test_data_frame_editor.py
```diff
@@ -0,0 +1,287 @@
...
+            [ 3,  4,  5],
+            [ 6,  7,  8],
+            [ 9, 10, 11]]
+    df = DataFrame(data, index=range(1,5),
+                   columns=['X', 'Y', 'Z'])
+    viewer = DataFrameViewer(data=df)
+    return viewer
+
+
+def test_adapter_get_item():
```
jonathanrocher: We should add a unit test like this one for the case where the dataframe has no rows as it should be supported. It could be the initial state of a model that this editor is viewing.

traitsui/ui_editors/data_frame_editor.py
```diff
@@ -0,0 +1,309 @@
...
+    #---- Adapter methods that are not sensitive to item type ----------------
+
+    def get_item(self, object, trait, row):
+        """ Override the base implementation to work with DataFrames
+
+        This returns a dataframe with one row, rather than a series, since
+        using a dataframe preserves dtypes.
+
+        """
+        return getattr(object, trait).iloc[row:row+1]
```
jonathanrocher: This fails with Pandas 0.12 with an `IndexError` when the dataframe being viewed has no rows. Newer versions of Pandas are fine.

traitsui/tests/ui_editors/test_data_frame_editor.py
```diff
@@ -0,0 +1,287 @@
...
+    viewer = sample_data()
+    adapter = DataFrameAdapter()
+
+    item = adapter.get_item(viewer, 'data', 0)
```
jonathanrocher: I would replace `item` by `first_row_df` to help understand what this item object is.

traitsui/tests/ui_editors/test_data_frame_editor.py
corranwebster: Done.

traitsui/ui_editors/data_frame_editor.py
corranwebster: 0.12 is 2 years old at this point.  I'm willing to not worry about support for it.

PR #198, opened 2015-04-09
Update vertical spring behavior description

johntyree:
Was poking through demos and found this inaccurate label. Presumably vertical springs used to not work. They apparently do now.


johntyree:
For reference

pankajp:
Possibly a case of wx vs Qt?

corranwebster:
Indeed.  It moves down on Qt, but not Wx.

cfarrow:
Superseded by #320.

PR #203, opened 2015-05-21
Import version correctly for development versioning

corranwebster:
Can't exec `__init__.py` any more because it uses absolute imports; instead exec `_version.py`.


corranwebster:
@itziakos Care to check that I'm not doing something dumb.

corranwebster:
Self-merging as time is of the essence.

PR #209, opened 2015-05-26
Fix installation issue 206

itziakos:


itziakos:
@cournape  and @corranwebster

corranwebster:
LGTM

PR #221, opened 2015-07-01
Use travis ci containers

itziakos:
This PR augments the testing of traitsui in travis-ci

In more details:
- We use the travis-ci containers so that PR testing starts faster
- Testing takes place using `qt`, `wx` and the `null` backends
- Coverage is calculated and uploaded at codecov.io
- minor fixes for the tests to be skipped when incompatible backend is active

**Note**: 
- Some tests are failing using the wx backend so coverage does not include these runs.
- The wx build will not count for the status updates (see issue #231)


itziakos:
Need to also check using 2.9.x and then update tests to skip known issues.

itziakos:
@corranwebster  this PR is ready for review

corranwebster:
:+1:

PR #228, opened 2015-08-03
Deprecate wx toolkit default

corranwebster:
Warn that wx toolkit will be deprecated in version 5.0.  Plus improve README a bit.


README.rst
```diff
@@ -14,9 +16,47 @@ The following GUI backends are supported:
 - PyQt
 - PySide
 
+**Warning:** In TraitsUI version 5.0 the default GUI backend will change from
```
itziakos: one can use rst admonitions http://docutils.sourceforge.net/docs/ref/rst/directives.html#admonitions), they should be rendered properly by sphinx

corranwebster: I know - see earlier revisions and the update to the docs.  Unfortunately Github doesn't render admonitions properly for the README on the project landing page, which is why I wanted this warning here.

traitsui/toolkit.py
```diff
@@ -110,6 +112,8 @@ def toolkit ( *toolkits ):
         return _toolkit
     else:
         if len( toolkits ) == 0:
+            warnings.warn(DeprecationWarning(
```
itziakos: I think that should be `warn(message, DeprecationWarning)`

itziakos: A simple test using https://docs.python.org/2/library/warnings.html#available-context-managers should be possible here.

README.rst
itziakos: fair enough (especially when it does not look that it will arrive soon https://github.com/github/markup/issues/68)

traitsui/toolkit.py
```diff
@@ -110,6 +110,12 @@ def toolkit ( *toolkits ):
         return _toolkit
     else:
         if len( toolkits ) == 0:
+            print 'here'
```
itziakos: remove print statement

itziakos:
One minor comment :+1:

PR #230, opened 2015-08-17
[WIP] Python 3 support

corranwebster:
Extending #165 to get it working without depending on private repo wheels.


.travis.yml
```diff
@@ -1,20 +1,25 @@
...
+    # For PySide
+    - 'export LD_LIBRARY_PATH=`python -c "import os,nose; print(os.path.dirname(os.path.dirname(nose.__file__)))"`/PySide'
+
 install:
   # nose is already installed
   - sudo apt-get install python-numpy
   - pip install -r dev_requirements.txt
   - python setup.py develop
+  # Test against the current master of traits
+  - pip install git+http://github.com/enthought/traits.git#egg=traits
```
itziakos: This is not required since dev_requirements contains the necessary entries

.travis.yml
```diff
@@ -1,20 +1,25 @@
...
+  # Test against the current python3 branch of pyface
+  - pip install git+http://github.com/pradyun/pyface.git@python3#egg=pyface
```
itziakos: see https://github.com/enthought/traitsui/pull/230/files#r37177945

.travis.yml
```diff
@@ -1,20 +1,25 @@
...
+  - python setup.py build
```
itziakos: This is also unnecessary

corranwebster: I think this may have been needed to invoke py2to3?

itziakos: I do not think so, the build command should run automatically when one does `python setup install`. I think it is probably a personal preference.

itziakos:
@corranwebster, we are now running tests on python 3.4 using pyside 1.2.2 and qt 4.8.5. There are still some tests failures.

corranwebster:
:fireworks: Yay!  I'll look into the failures to see what the cause looks like.

corranwebster:
Looks like the Qt failures are all around extracting strings from the clipboard.

The null backend failures are around the Pyface `null` toolkit not having an `image_cache` module.

corranwebster:
Qt4 passing on Python 3 - only remaining problem is with pyface imports of missing modules under Python 3 - structure of `ImportError` has changed so we need to deal with it.

corranwebster:
Plan for merging this and corresponding Pyface PR:

* merge Pyface first, still pointing at this PR
* change this PR to use Pyface Master
* merge this PR
* add a secondary PR to Pyface to switch to TraitsUI master
* merge it

itziakos:
Plan is ok (I added checkboxes :P)

corranwebster:
And we're good.

Merging.

PR #234, opened 2015-08-21
Various Python 3 Fixes

corranwebster:
Fixes which have come up in actual testing of examples.

### DIFF
--- a/examples/demo/Advanced/Table_editor_with_live_search_and_cell_editor.py
+++ b/examples/demo/Advanced/Table_editor_with_live_search_and_cell_editor.py
@@ -393,7 +393,7 @@ class SourceFile ( HasTraits ):
     @property_depends_on( 'full_name' )
     def _get_contents ( self ):
         try:
-            fh = open( self.full_name, 'rb' )
+            fh = open( self.full_name, 'rU' )
             contents = fh.readlines()
             fh.close()
             return contents
@@ -425,4 +425,3 @@ demo = LiveSearch()
 # Run the demo (if invoked from the command line):
 if __name__ == '__main__':
     demo.configure_traits()
-
--- a/examples/demo/Advanced/Table_editor_with_live_search_and_cell_editor.py
+++ b/examples/demo/Advanced/Table_editor_with_live_search_and_cell_editor.py
@@ -393,11 +393,10 @@ class SourceFile ( HasTraits ):
     @property_depends_on( 'full_name' )
     def _get_contents ( self ):
         try:
-            fh = open( self.full_name, 'rU' )
-            contents = fh.readlines()
-            fh.close()
-            return contents
-        except:
+            with open(self.full_name, 'rU') as fh:
+                contents = fh.readlines()
+                return contents
+        except Exception:
             return ''
 
     @property_depends_on( 'full_name, live_search.[search, case_sensitive]' )
--- a/examples/tutorials/tutor.py
+++ b/examples/tutorials/tutor.py
@@ -27,6 +27,7 @@
 #-------------------------------------------------------------------------------
 
 import sys
+from operator import itemgetter
 import os
 import re
 
@@ -187,13 +188,13 @@ MP3Template = """<html>
 #  Returns the contents of a specified text file (or None):
 #-------------------------------------------------------------------------------
 
-def read_file ( path, mode = 'rb' ):
+def read_file ( path, mode = 'rU' ):
     """ Returns the contents of a specified text file (or None).
         """
     fh = result = None
 
     try:
-        fh     = file( path, mode )
+        fh     = open( path, mode )
         result = fh.read()
     except:
         pass
@@ -607,7 +608,7 @@ class ASection ( HasPrivateTraits ):
                                       match.group(1), dir ) )
 
         # Sort the directories by their index value:
-        dirs.sort( lambda l, r: cmp( l[0], r[0] ) )
+        dirs.sort(key=itemgetter(0))
 
         # Create the appropriate type of section for each valid directory:
         self._subsections = [
@@ -1786,4 +1787,3 @@ where: tutorial_path = Path to the root of the traits tutorial.
 
 If tutorial_path is omitted, the current directory is assumed to be the root of
 the tutorial.""" % path
-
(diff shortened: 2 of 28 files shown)


traitsui/editors/api.py
```diff
@@ -7,7 +7,7 @@
 try:
     from .array_editor import ArrayEditor
 except ImportError as e:
-    if 'numpy' not in e.message:
```
itziakos: I would think that the check should try to import numpy and on success import ArrayEditor instead of this *ugly* check on the error message. Is there a reason for not doing that?

corranwebster:
At this point, all the examples which run under Python 2 & Qt also run under Python 3 & Qt (less a couple which rely on apptools and chaco)

examples/tutorials/tutor.py
```diff
@@ -187,13 +188,13 @@
...
 
-def read_file ( path, mode = 'rb' ):
+def read_file ( path, mode = 'rU' ):
     """ Returns the contents of a specified text file (or None).
         """
     fh = result = None
 
     try:
-        fh     = file( path, mode )
+        fh     = open( path, mode )
```
itziakos: Can we fix this ugly set of `try..except`  blocks with a with context. :grinning:

examples/demo/Advanced/Table_editor_with_live_search_and_cell_editor.py
```diff
@@ -393,7 +393,7 @@ def _get_ext_path ( self ):
     @property_depends_on( 'full_name' )
     def _get_contents ( self ):
         try:
-            fh = open( self.full_name, 'rb' )
+            fh = open( self.full_name, 'rU' )
```
itziakos: This could also be made a little less ugly with a context manager.

traitsui/editors/api.py
```diff
@@ -7,11 +7,15 @@
...
+    # check if failure is due to missing numpy, otherwise re-raise
+    try:
+        import numpy
+    except ImportError:
+        import warnings
+        warnings.warn('ArrayEditor is not available due to missing numpy',
+                      ImportWarning)
+    else:
```
itziakos: We do not need the `else:` here

traitsui/editors/api.py
```diff
@@ -7,11 +7,15 @@
 try:
     from .array_editor import ArrayEditor
 except ImportError as e:
-    if 'numpy' not in e.message:
```
itziakos: Why not

```
try: 
    import numpy
except ImportError:
    import warnings
    warnings.warn('ArrayEditor is not available due to missing numpy', ImportWarning)
else:
    from .array_editor import ArrayError
```

traitsui/qt4/tabular_editor.py
```diff
@@ -348,8 +353,9 @@ def _selected_changed(self, new):
         if not self._no_update:
             try:
                 selected_row = self.value.index(new)
-            except:
-                pass
+            except Exception as e:
+                if TRAITS_DEBUG:
+                    print(e)
```
itziakos: Would it be better to use a warning here?

traitsui/table_column.py
```diff
@@ -457,7 +474,9 @@ def get_raw_value ( self, object ):
         """
         try:
             return xgetattr( self.get_object( object ), self.name )
-        except:
+        except Exception as e:
+            if TRAITS_DEBUG:
+                print("Can't get raw value:", e)
```
itziakos: see https://github.com/enthought/traitsui/pull/234/files#r37774653

rkern: Instead, use the existing `raise_to_debug()` function.

traitsui/editors/api.py
corranwebster: If everything goes cleanly, this puts `numpy` into the namespace of the module, which we don't want for an "api" module.

We could `import numpy as _numpy` or something, I guess.

rkern: Following it with a `del numpy` is the usual approach.

traitsui/editors/api.py
corranwebster: Yes, we do, since we can't put the `raise` in the inner `try: ... except: ...` and we want to raise if we didn't get an import error with numpy, since that implies an unexpected import error.

traitsui/editors/api.py
corranwebster: I think I slightly prefer the approach of "try and import the editor, and if it fails, then see if we can work out if it's because of numpy being missing"

corranwebster:
Assuming I haven't broken any tests, any objections to merging?

traitsui/editors/api.py
itziakos: I would be happier with a simple approach that does not require two `try...except`

itziakos: @rkern what is your opinion?

traitsui/editor_factory.py
```diff
@@ -184,9 +187,13 @@ def _get_toolkit_editor(cls, class_name):
...
                                 sys.modules[factory_class.__module__].__file__)
-                return toolkit_object(':'.join([editor_file_name.split('.')[0],
-                                             class_name]), True)
+                object_ref = ':'.join([editor_file_name.split('.')[0],
+                                       class_name])
+                return toolkit_object(object_ref, True)
             except Exception as e:
+                if TRAITS_DEBUG:
+                    msg = "Can't import toolkit_object '{}': {}"
+                    print(msg.format(object_ref, e))
```
rkern: Please use `raise_to_debug()` in all of these cases.

traitsui/editors/api.py
rkern: Two, three, even four `try`s would be perfectly fine, IMO. This is done exactly once. It can afford to be very mildly complicated.

traitsui/editor_factory.py
corranwebster: I don't think that's correct for all of these cases: sometimes the failure to load an editor is expected for particular editors.

In these cases of expected failed loads when `TRAITS_DEBUG` is `True`, would you prefer `print` or logging in these situations?

rkern: The point of turning on `TRAITS_DEBUG` is to reraise the exception. If you don't want to reraise the exception ever, just unconditionally use `logger.debug()`.

corranwebster: Fair enough.

PR #242, opened 2015-10-12
Qt5 Support (V2)

corranwebster:
This is (currently) just bringing @geggo's Qt5_V2 branch into the main repo so we can ensure that tests run correctly, at least on the old test infrastructure.

I suspect that initially we will have failures due to PyFace/TraitsUI cross-dependency, so we'll need to point to appropriate branches at least initially.


corranwebster:
Looks like there is an issue that `QItemSelection` has moved from `QtGui` to `QtCore` in the 4 to 5 shift. This is causing test failures in TraitsUI.

I think the fix is conditionally import it into `pyface.qt.QtCore` when using Qt4.

corranwebster:
Same thing with `QItemSelectionModel`

pankajp:
@corranwebster I'm fixing a couple of things with this PR and also attempting to add travis support for testing PyQt5 in pyface here

corranwebster:
Thanks - that's very useful.  I spent a little time on the Pyface version a couple of weeks ago, in particular trying to get Travis to work there, but didn't have enough time to persist with it.

corranwebster:
Closing this, as we have achieved qt5 support at a more basic level via Pyface.

PR #246, opened 2015-11-04
Fix: allow hiding of groups in split layouts. Fix auto_set=False behavior of CodeEditor

bergtholdt:
The first fix is for split layouts when using visible_when to hide groups, which caused an error.

The second fix is for auto_set=False in Qt CodeEditor which was completely ignored (works in WX). Note that this requires also my pyface pull request

### DIFF
--- a/traitsui/qt4/editor.py
+++ b/traitsui/qt4/editor.py
@@ -155,7 +155,7 @@ class Editor ( UIEditor ):
         self._visible_changed_helper(self.control, visible)
 
         page = self.control.parent()
-        if page is None or page.parent() is None or page.parent().parent() is None or page.layout().count() != 1:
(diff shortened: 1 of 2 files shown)


corranwebster:
Generally looks good.

traitsui/qt4/editor.py
```diff
@@ -155,7 +155,7 @@ def _visible_changed(self, visible):
         self._visible_changed_helper(self.control, visible)
 
         page = self.control.parent()
-        if page is None or page.parent() is None or page.parent().parent() is None or page.layout().count() != 1:
+        if page is None or page.parent() is None or page.parent().parent() is None or page.layout() is None or page.layout().count() != 1:
```
corranwebster: This line looks a bit long.  If you have time to PEP8 it, that would be good.

PR #251, opened 2015-11-12
Fix empty tree nodes not changing when adding children.

### LINKED ISSUE #250: Bug: tree editor doesn't refresh when empty lists are modified
The following code displays a class hierarchy using a tree editor. Adding new experiments or new studies to the existing ones triggers a refresh of the tree editor, so that the new elements get displayed. It doesn't happen with the simulations, which start empty. I believe this is a bug.

```
from traits.api import Button, HasTraits, Instance, List, Str
from traitsui.api import UItem, TreeEditor, View, ModelView, TreeNode

class Experiment(HasTraits):
    name = Str

class Simulation(HasTraits):
    name = Str

class Study(HasTraits):
    name = Str
    experiments = List(Instance(Experiment))
    simulations = List(Instance(Simulation))

class DataManager(HasTraits):
    studies = List(Instance(Study))

class DataManagerModelView(ModelView):
    add_study = Button
    add_simulation = Button
    add_experiment = Button

    def _add_study_fired(self):
        self.model.studies.append(Study(name='foo'))
        print self.model.studies

    def _add_experiment_fired(self):
        self.model.studies[0].experiments.append(Experiment(name='bar'))
        print self.model.studies[0].experiments

    def _add_simulation_fired(self):
        self.model.studies[0].simulations.append(Simulation(name='baz'))
        print self.model.studies[0].simulations

    traits_view = View(
        UItem('model', editor=TreeEditor(
                editable=False,
                auto_open=4,
                nodes=[
                    TreeNode(
                        node_for=[DataManager],
                        label="=Studies",
                        children='studies',
                    ),
                    TreeNode(
                        node_for=[Study],
                        label="name",
                        children='',
                    ),
                    TreeNode(
                        node_for=[Study],
                        label="=Experiments",
                        children='experiments',
                    ),
                    TreeNode(
                        node_for=[Study],
                        label="=Simulations",
                        children='simulations',
                    ),
                    TreeNode(
                        node_for=[Simulation],
                        label="name",
                    ),
                    TreeNode(
                        node_for=[Experiment],
                        label="name",
                    ),
                ]
            )
        ),
        UItem('add_study'),
        UItem('add_simulation'),
        UItem('add_experiment'),
        resizable=True,
    )


if __name__ == '__main__':
    data_manager = DataManager(
        studies=[
            Study(
                name='Study A',
                experiments=[
                    Experiment(name='Experiment A1'),
                    Experiment(name='Experiment A2'),
                    Experiment(name='Experiment A3'),
                ],
                simulations=[],
            ),
        ]
    )
    model_view = DataManagerModelView(model=data_manager)
    model_view.configure_traits()
```

corranwebster:
Fixes #250.

When going from having no children to having children, nodes were not being updated to indicate that they could be expanded.  The tree editor handles this by adding a dummy child to the unexpanded node so that Qt will flag it as expandable.  This was not happening in the handlers for replacing all children or updating children for a node in the case where they were not expanded and started with no children.

A better fix for this would be to replace the nodes in the Qt model with `QModelViewItem` subclasses instead of `QTreeWidgetItem`s, but that's a substantial amount of work.

### DIFF
--- a/traitsui/qt4/tree_editor.py
+++ b/traitsui/qt4/tree_editor.py
@@ -1521,9 +1521,14 @@ class SimpleEditor ( Editor ):
                     if child_node is not None:
                         self._append_node( nid, child_node, child )
             else:
-                # if model has children, but is not expanded, add dummy child
-                if len(children) > 0:
+                dummy = getattr(nid, '_dummy', None)
+                if dummy is None and len(children) > 0:
+                    # if model now has children add dummy child
                     nid._dummy = QtGui.QTreeWidgetItem(nid)
+                elif dummy is not None and len(children) == 0:
+                    # if model no longer has children remove dummy child
+                    nid.removeChild(dummy)
+                    del nid._dummy
 
             # Try to expand the node (if requested):
             if node.can_auto_open( object ):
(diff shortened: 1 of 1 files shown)


corranwebster:
@jonathanrocher Can you please check that this solves your issue #250 

If it is good, I will merge.

jonathanrocher:
Yep, the bug is fixed. From a functionality point of view, LGTM.

PR #253, opened 2015-11-17
BUG: Set panel state to be initialized after setting prefs

### LINKED ISSUE #197: QT4 restore_prefs for splitter groups
Hi,

I think in the qt4 toolkits SplitterGroupEditor class, the use of a wrong attribute causes the restored state to be overwritten later.
ui_panel.py:1164 (master branch)
self.splitter._resize = True
should rather look like this?
self.splitter._initialized = True

Many thanks for the great library!
Christian

senganal:
This PR applies the fix pointed out in #197.. 

This fixes the issue of the splitter prefs not being restored correctly..

Closes #197


senganal:
@corranwebster  Here is the driver code used to check manually that the preferences are indeed saved and loaded.. Run the script, make some layout changes and close the ui.. run the script again and check the layout is correct.. Can prob. make it to a testcase if you are OK with the fix.

```
from traits.api import HasStrictTraits, Int
from traitsui.api import View, HSplit, Group, Item

class TmpClass(HasStrictTraits):
    aa = Int(10)
    bb = Int(100)

    view = View(
        HSplit(
            Group(
                Item('aa', resizable=True, width=50, id='txt1'),
                show_border=True,
                id='grp1',
            ),
            Group(
                Item('bb', width=100, id='txt2'),
                show_border=True,
                id='grp2',
            ),
            id='Split1',
        ),
        id='NewTmpClass153',
        resizable=True,
        height=300,
    )

if __name__ == '__main__':
    ui = TmpClass().configure_traits()
```

corranwebster:
Looks like appveyor is having issues.

corranwebster:
Looks good.  A test case would be very good, although this may be awkward to test because of the close/reopen flow to the test.

We may need @itziakos help to work out what's gone wrong with Appveyor.

corranwebster:
@senganal I think that the appveyor tests may be failing because you are based off a very old master.  Can you try a pull on master and then merge to see if that gets the appveyor tests working.

senganal:
> Can you try a pull on master and then merge to see if that gets the appveyor tests working.

Indeed.. Thanks .. working on a testcase now..

senganal:
@corranwebster Added a testcase for the splitter prefs bug. Please take a look.

senganal:
The test failure (only on python 3) is unrelated to this issue here but due to a bug (saving prefs state seems broken in python 3) that gets triggered by the testcase..

andrewcollette:
@senganal, @corranwebster, what additional work would be required to get this merged?  Right now in the Canopy Cloud demos we have to use a custom patched egg with this fix.  I am happy to spend some time if needed on this.

corranwebster:
Overall, looks good.  I am pull in current master to re-run the tests before merging.

corranwebster:
Looks like appveyor is getting stuck.  Will merge anyway.

PR #255, opened 2015-11-17
Bug in the constructor of `Controller`

senganal:
The current constructor of `Controller` overwrites any defaults set by a subclass for the `model` attribute (pls. see the testcase)..

This PR fixes the problem by passing in the `model` kwarg to the base constructor.


senganal:
@corranwebster Pls. take a look.. I think the fix is pretty harmless..

corranwebster:
See comment on #253 about appveyor failures.

traitsui/handler.py
```diff
@@ -543,11 +543,10 @@ class Controller ( Handler ):
 
     #-- HasTraits Method Overrides ---------------------------------------------
 
-    def __init__ ( self, model = None, **metadata ):
+    def __init__ ( self, **traits):
```
corranwebster: With this modification, the `__init__` isn't needed at all.  However the original version avoids an initial trait change firing on the setting of the model, which may be intentional.

corranwebster:
Given the slight difference in the `__init__` behaviours noted above, it may be better to fix by doing:

    if model is not None:
        self.model = model

traitsui/handler.py
rkern: This breaks the API. `Controller(model)` is the intended way to instantiate this.

pberkes:
-1 passing the model in a Controller or ModelView is the natural thing to do. Also, changing it will likely break existing applications. 

If for some reason one needs to have a controller on a default model, one can subclass from Handler directly.

senganal:
>  This breaks the API. Controller(model) is the intended way to instantiate this.

oops.. didn't think of the case of passing in the `model` as a positional argument.. thanks for catching that..

> -1 passing the model in a Controller or ModelView is the natural thing to do.

Agreed.. maybe the example in the testcase is not natural.. But still, the current behavior is a bug..
I think the updated fix should fix the problem and not cause any backwards incompatibility..

@corranwebster Pls. take another look.. Feel free to close this PR if you feel the current behavior is not a bug and not worth fixing..

senganal:
>@corranwebster said:
> However the original version avoids an initial trait change firing on the setting of the model, which may be intentional

oh.. true.. updated to the fix as per your suggestion..

PR #263, opened 2016-02-18
Add toolbar in Qt UI panel (fix for Mayavi)

kitchoi:
The Mayavi2 application do not show the toolbar for the pipeline (one of the workbench panels) when the backend is Qt4.

This will allow the panel to add the toolbar to the window.  However, the toolbar would be added to the main window instead of the panel.  The full fix for mayavi would require https://github.com/enthought/pyface/pull/192 for the toolbar to appear in the panel it's associated with.

Before:
before

After:
after

### DIFF
--- a/traitsui/qt4/ui_base.py
+++ b/traitsui/qt4/ui_base.py
@@ -277,6 +277,45 @@ class BasePanel(object):
         """
         self.revert.setEnabled(state)
 
+    #---------------------------------------------------------------------------
+    #  Adds a menu item to the menu bar being constructed:
+    #---------------------------------------------------------------------------
+
+    def add_to_menu ( self, menu_item ):
+        """ Adds a menu item to the menu bar being constructed.
+        """
+        item   = menu_item.item
+        action = item.action
+
+        if action.id != '':
+            self.ui.info.bind( action.id, menu_item )
+
+        if action.style == 'radio':
+            if ((self._last_group is None) or
+                (self._last_parent is not item.parent)):
+                self._last_group = RadioGroup()
+                self._last_parent = item.parent
+            self._last_group.items.append( menu_item )
+            menu_item.group = self._last_group
+
+        if action.visible_when != '':
+            self.ui.add_visible( action.visible_when, menu_item )
+
+        if action.enabled_when != '':
+            self.ui.add_enabled( action.enabled_when, menu_item )
+
+        if action.checked_when != '':
+            self.ui.add_checked( action.checked_when, menu_item )
+
+    #---------------------------------------------------------------------------
+    #  Adds a tool bar item to the tool bar being constructed:
+    #---------------------------------------------------------------------------
+
+    def add_to_toolbar ( self, toolbar_item ):
+        """ Adds a toolbar item to the toolbar being constructed.
+        """
+        self.add_to_menu( toolbar_item )
+        
 
 class _StickyDialog(QtGui.QDialog):
     """A QDialog that will only close if the traits handler allows it."""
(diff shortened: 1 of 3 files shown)


traitsui/qt4/ui_panel.py
```diff
@@ -204,11 +204,35 @@ def __init__(self, ui, parent, is_subpanel):
...
+        """
+        toolbar = self.ui.view.toolbar
+        if toolbar is not None:
+            self._last_group = self._last_parent = None
+            qt_toolbar = toolbar.create_tool_bar(parent, self )
+            qt_toolbar.setMovable( False )
+            parent.addToolBar( qt_toolbar )
+            self._last_group = self._last_parent = None
+
+    def can_add_to_toolbar(self, action):
```
corranwebster: Is this function used anywhere?  It looks like it is needed to hide items which shouldn't be displayed based on the View's context, but I don't see it being explicitly called.

traitsui/qt4/ui_base.py
```diff
@@ -556,45 +595,6 @@ def set_status_text(text):
...
-    #---------------------------------------------------------------------------
-    #  Adds a tool bar item to the tool bar being constructed:
-    #---------------------------------------------------------------------------
-
-    def add_to_toolbar ( self, toolbar_item ):
-        """ Adds a toolbar item to the toolbar being constructed.
-        """
-        self.add_to_menu( toolbar_item )
-
     def can_add_to_menu(self, action, action_event=None):
```
corranwebster: Does this function also need to be moved?

traitsui/qt4/ui_panel.py
kitchoi: It is called here: https://github.com/enthought/pyface/blob/master/pyface/action/action_item.py#L127

Without `can_add_to_toolbar`, BasePanel cannot add a toolbar unless `controller` is not None in `pyface.action.action_item.add_to_toolbar`, which I have little idea what that is.

I am thinking of adding `can_add_to_toolbar` in BasePanel that raises NotImplemented.

traitsui/qt4/ui_base.py
kitchoi: These functions were in `BaseDialog` https://github.com/enthought/traitsui/blob/master/traitsui/qt4/ui_base.py#L563, not in `_StickyDialog`.  diff is confused.
`BaseDialog` inherits `BasePanel`.

kitchoi:
@corranwebster Have I answered your questions?  Should I proceed to write tests for this?

corranwebster:
Sorry. Yes, go ahead.

kitchoi:
Thanks @corranwebster 
I have added tests for this PR.  Can you take a look please?

corranwebster:
What difference in behaviour are you seeing between different versions of Qt?

kitchoi:
On my VM with Qt 4.8.x, when I test `test_subpanel_has_toolbar_no_buttons_qt4`,
all `dock.findChild(QtGui.QPushButton)`, `dock.widget().findChild(QtGui.QPushButton)` and `ui.control.findChild(QtGui.QPushButton)` give the expected `None` as the subpanel should not have buttons.
However, on Travis which I believe uses PyQt4 4.11.x, return a `QPushButton` for `dock.findChild(QtGui.QPushButton)`.

kitchoi:
@corranwebster I am sorry I got the package versions wrong in the previous comment:  I have tested on my own VM with PyQt 4.9.1, 4.11.x and PySide 1.2.2, both with Qt 4.8.1 and Qt 4.8.7 (6 combinations).  However I couldn't reproduce the error in the previous failed build on Travis ([here](https://s3.amazonaws.com/archive.travis-ci.org/jobs/111446005/log.txt)), which I believe uses PyQt 4.9.1 with Qt 4.8.0 (or 4.8.1?).  The python3 build for the same commit does not have that problem.

kitchoi:
In tests that ensure NO OK/Cancel buttons is shown for subpanel, `ui.control.findChild(QtGui.QPushButton)` is indeed the right thing to do.  The wrong behaviour (e.g. not checking for "subpanel" in `has_button` https://github.com/enthought/traitsui/blob/qt4-ui-panel-can-add-toolbar/traitsui/qt4/ui_panel.py#L122) would fail the test.

Previous confusion comes from the fact that `dock.findChild(QtGui.QPushButton)` should return the push button for popping up the panel.  However this does not occur all the time at least for my VM.  Sometimes `None` is returned and this behaviour is quite random.

@corranwebster can you take a look again please?  Thanks so much!

corranwebster:
I think that this is good to merge.

PR #265, opened 2016-03-08
Fix wx capture mouse issue.

RobinD42:
This PR changes ThemedWindow to use the wx HasCapture method instead of trying to track the state itself.

wxPython 3 is more strict about capturing the mouse, and no longer allows a widget to capture it if it already holds the capture, and it's also possible for the capture to be stolen by another widget.  So trying to track the state of the capture in a local attribute can sometimes have holes, resulting in assertion errors from wx.


corranwebster:
LGTM - I'll trust your expertise on the fix.

PR #266, opened 2016-03-14
Fix wx _left_up() event for themed windows

JCorson:
This is related to enthought/traitsui#265. The behavior of themed_windows was still a little buggy with some missed click events. I think this may solve the issue.


corranwebster:
LGTM

Issue #269, opened 2016-03-21
Exception during disposal of ListStrEditor with Qt4 backend


stefanoborini:
Found thanks to an investigation at enthought/mayavi#319

The problem is the following: mayavi 2 raises an exception at closing. I tracked down the problem and here I will make the story short for practical purposes.

1. When the disposal initiates, UI.reset() is called. This goes through all the editors and calls `dispose()` on them (see ui.py:313). This seems reasonable...
2. The problem happens with NotebookEditor under qt4/list_editor.py. This Editor basically represents a tabbed widget. calling dispose() on it calls close_all(), which performs disposal of the UI elements (ok) and then calls control.clear().
3. The control is the QTabWidget. This Qt widget has a model, which is a QAbstractListModel: ListStrModel. Qt clear() method, in order to perform its job, apparently needs to know how many rows this model has, so it invokes rowCount().
4. Our ListStrModel reimplements rowCount(), which tries to inquire the associated trait content. The object holding the trait object is in editor.object... (list_str_model.py:61)
5. Which unfortunately has been set to None at point 1 (see editor.py:204)

Fixing this requires that either we delay the disposal of the editors, or that we return zero in ListStrModel if there's no associated trait model to retrieve info from. I feel the second solution is not correct because there's no guarantee that the disposal will happen appropriately. I will investigate both strategies and add more to this.

stefanoborini:
After additional  investigation, it might be more delicate and potentially not the real cause.
the base class dispose() is correctly called _after_, meaning that there's no apparent reason why we should have a None at 4. But None we have nevertheless.

stefanoborini:
Closing this one. Fixed.

PR #270, opened 2016-03-22
Fixes exception occurring during disposal of ListStrEditor

stefanoborini:
Solves issue #269. The root cause was that the adapter was not able to deal with
None objects. This is already performed in the TabularAdapter, and was missing in
the ListStrAdapter. The result was that ListStrModel.rowCount was raising exception
because at deinitialization, the editor trait object can be None. It is sensible in
this case that the adapter simply returns zero.

### DIFF
--- a/traitsui/tests/editors/test_liststr_editor.py
+++ b/traitsui/tests/editors/test_liststr_editor.py
@@ -1,0 +1,39 @@
+#------------------------------------------------------------------------------
+#
+#  Copyright (c) 2013, Enthought, Inc.
+#  All rights reserved.
+#
+#  This software is provided without warranty under the terms of the BSD
+#  license included in enthought/LICENSE.txt and may be redistributed only
+#  under the conditions described in the aforementioned license.  The license
+#  is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+#  Author: Stefano Borini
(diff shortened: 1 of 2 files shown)


corranwebster:
LGTM.  Assuming all tests pass cleanly, feel free to merge.

traitsui/tests/editors/test_liststr_editor.py
```diff
@@ -0,0 +1,39 @@
...
+
+"""
+Test case for ListStrEditor and ListStrAdapter
+"""
+
+from traits.has_traits import HasTraits
+from traits.trait_types import List, Str
+from traitsui.list_str_adapter import ListStrAdapter
+
+class TraitObject(HasTraits):
```
rkern: flake8: One more blank line before this.

traitsui/tests/editors/test_liststr_editor.py
```diff
@@ -0,0 +1,39 @@
...
+
+    object = TraitObject()
+    object.list_str = ["hello"]
+
+    adapter = ListStrAdapter()
+
+    assert adapter.len(object, "list_str") == 1
+    assert adapter.len(None, "list_str") == 0
+
+
```
rkern: flake8: no trailing newlines.

Issue #260, opened 2016-01-06
TraitsUI is version 5, but docs still say "TraitsUI 4 User Manual"


pberkes:

corranwebster:
Fixed.

PR #273, opened 2016-03-25
Updating docs to say TraitsUI 5 instead of 4

rahulporuri:
Apart from changing the 4 to 5 in a few places, changes are

- used `try`, `except` to catch IOError raised when `_version.py` is missing to provide information on how the solution
- changed the ImportError error message raised when enthought-sphinx-theme is missing to show where to download the theme from
- added a brief README.rst file.

ref: issue #260


docs/source/conf.py
```diff
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
```
corranwebster: Removing this _may_ be safe, but it was auto-generated with this coding tag, so it's probably wisest to leave it there (it signals to Python the encoding of the file).

docs/source/conf.py
```diff
@@ -43,8 +43,15 @@
...
 d = {}
-execfile(os.path.join('..', '..', 'traitsui', '_version.py'), d)
-version = release = d['version']
+try:
+    execfile(os.path.join('..', '..', 'traitsui', '_version.py'), d)
+    version = release = d['version']
+except IOError as ioe:
+    import warnings
+    msg = '''_version.py seems to be missing,
+            which is autogenerated if the users runs
```
corranwebster: a clearer message would be something like "please run 'python setup.py develop' to generate this file."

docs/source/conf.py
```diff
@@ -87,7 +94,7 @@
 
 # The name for this set of Sphinx documents.  If None, it defaults to
 # "<project> v<release> documentation".
-html_title = "TraitsUI 4 User Manual"
+html_title = "TraitsUI 5 User Manual"
```
corranwebster: Can we build this using the version number which is set somewhere earlier in the file?  That would prevent this particular error from ever occurring again.

rahulporuri: i can do some string formatting and use the version information from `execfile(...)`. would that be okay?

corranwebster:
There may be some places it makes sense to use the `|version|` substitution in the changes that you have made to the .rst files (see http://www.sphinx-doc.org/en/stable/markup/inline.html#substitutions).

docs/source/conf.py
corranwebster: If by the version information you mean the value you got from https://github.com/enthought/traitsui/blob/issue_260_docs/docs/source/conf.py#L48 then yes.

docs/source/conf.py
```diff
@@ -94,7 +95,7 @@
 
 # The name for this set of Sphinx documents.  If None, it defaults to
 # "<project> v<release> documentation".
-html_title = "TraitsUI 5 User Manual"
+html_title = "TraitsUI {} User Manual".format(version.strip('.')[0])
```
corranwebster: `split` not `strip`, I think

docs/source/conf.py
rahulporuri: yes. added a new commit with the aforementioned corrections/changes. will look into using |version| now

docs/source/conf.py
rahulporuri: (ahhhhhh!!!!!! every.single.time.) sorry, corrected it!

rahulporuri:
@corranwebster 
screen shot 2016-03-30 at 12 12 44
`|version|` seems to be behaving differently than defined [here](http://www.sphinx-doc.org/en/stable/markup/inline.html#substitutions). I guess it's because we redefine version in `source/conf.py` [here](https://github.com/enthought/traitsui/blob/issue_260_docs/docs/source/conf.py#L48). I can change this to get only `5` or maybe `5.1` if necessary.

What would be better?

corranwebster:
It seems that the root problem is that we have both version and release set to the same value coupled with `version` being set to the `full_version` in non-released builds. See: https://github.com/enthought/traitsui/blob/master/docs/source/conf.py#L47

I would suggest that we go with a convention in the config file where we set `version` to something like "5.1" (ie. something like `version = '.'.join(d["version"].split('.', 2)[:2])`) and keep `release` as it is (ie. `release = d["version"]`).

That should work well for documentation, I think.

corranwebster:
LGTM

rahulporuri:
@corranwebster if you have the time, can you comment on the PR and tell me if there any more corrections/additions to make?

corranwebster:
@rahulporuri I don't see any changes from my last comment.  Have you pushed everything?  If you have, then it's all good and can be merged.

rahulporuri:
Ohh! I didn't want to take the liberty to merge!

corranwebster:
It doesn't really matter who does it once the PR is approved.

rahulporuri:
solves #272

PR #275, opened 2016-03-30
Fix TableEditor selection handling

### LINKED ISSUE #274: TableEditor doesn't show initial selection in Qt
Under Qt, if you have a Table editor listening to a selection on the model, and that selection has an initial value when the editor is created, then the selection is not highlighted correctly.

Code to reproduce:

```
from traits.api import HasTraits, Instance, List, Str
from traitsui.api import ObjectColumn, TableEditor, UItem, View

class MyItem(HasTraits):

    foo = Str

    bar = Str


class MyCollection(HasTraits):

    items = List(Instance(MyItem))

    selected_item = Instance(MyItem)

    view = View(
        UItem(
            'items',
            editor=TableEditor(
                columns=[
                    ObjectColumn(
                        name='foo',
                        label='Foo',
                    ),
                    ObjectColumn(
                        name='bar',
                        label='Bar',
                    ),
                ],
                editable=False,
                sortable=False,
                selection_mode='row',
                selected='selected_item',
            )
        ),
        resizable=True,
    )


if __name__ == '__main__':
    collection = MyCollection(
        items=[
            MyItem(foo='1', bar='a'),
            MyItem(foo='2', bar='b'),
            MyItem(foo='3', bar='c'),
        ]
    )
    collection.selected_item = collection.items[0]
    collection.configure_traits()
```

and the result is:
screen shot 2016-03-30 at 11 04 22 am
when we expect to have:
screen shot 2016-03-30 at 11 07 02 am

corranwebster:
Fixes #274 and a bunch of other stuff which came up during tests.  This change makes the behaviour of Qt match the behaviour of Wx, but may potentially break Qt code which is written expecting the bad Qt behaviour.

In particular the selected_indices now returns different things depending on the selection mode, rather than always returning the (row, col) tuples, and so code written with Qt only in mind will need to adapt appropriately.  The changes should make the code cleaner in general.

In particular, selections in Qt should now match the documentation here http://docs.enthought.com/traitsui/traitsui_user_manual/factories_advanced_extra.html#interacting-with-items


corranwebster:
There are more issues here than it initially looked like, based on what I'm seeing when writing tests.  The Qt table editor needs work on its selection handling.

jwiggins:
CI failed on Python 3. Maybe run flake8 over the changed files to catch the remaining nits?

corranwebster:
Yep - there was a stray old-style print in the code.

traitsui/qt4/table_editor.py
```diff
@@ -463,6 +465,8 @@ def set_selection(self, objects=[], notify=True):
                     row = items.index(obj)
                 except ValueError:
                     continue
+                if row == -1:
```
jwiggins: In what situation would `row == -1`?

traitsui/qt4/table_editor.py
```diff
@@ -598,16 +606,36 @@ def _get_selected_indices(self):
...
         indices = self.model.mapSelectionToSource(selection_items).indexes()
-        return [(index.row(), index.column()) for index in indices]
+        if self.factory.selection_mode.startswith('row'):
+            indices = sorted(set(index.row() for index in indices))
+        elif self.factory.selection_mode.startswith('column'):
+            indices = sorted(set(index.column() for index in indices))
+        else:
+            indices = [(index.row(), index.column()) for index in indices]
+
+        if self.factory.selection_mode.endswith('s'):
```
jwiggins: This seems brittle to me

PR #276, opened 2016-04-14
Fix segfault when closing TableEditor while editing inline

pankajp:
The main culprit is the `table_view.setModel(None)` in
`TableEditor.dispose()`. It is cause of segfault because
when the model is set while editing, the underlying widget
is not in sync with it and something bad ensues.
The segfault happens in a fairly complex app with custom
cell editor spanning multiple rows. Here's a sample traceback:

> Program received signal SIGSEGV, Segmentation fault.
> QHashData::nextNode (node=node@entry=0x5555588f3a50) at tools/qhash.cpp:294
> 294	    if (next->next)
> (gdb) bt
> #0  0x00007fffe1fd2363 in QHashData::nextNode(QHashData::Node*) (node=node@entry=0x5555588f3a50) at tools/qhash.cpp:294
> #1  0x00007fffdfe046a3 in QAbstractItemView::updateEditorGeometries() (this=<synthetic pointer>) at ../../src/corelib/tools/qhash.h:355
> #2  0x00007fffdfe046a3 in QAbstractItemView::updateEditorGeometries() (this=this@entry=0x555558f1d280) at itemviews/qabstractitemview.cpp:2625
> #3  0x00007fffe0a5b778 in QTableViewWrapper::updateEditorGeometries() (this=0x555558f1d280)
>     at /usr/src/debug/pyside-qt4.8+1.2.2/x86_64-redhat-linux-gnu/PySide/QtGui/PySide/QtGui/qtableview_wrapper.cpp:2039
> #4  0x00007fffdfdf8cad in QAbstractItemView::updateGeometries() (this=this@entry=0x555558f1d280) at itemviews/qabstractitemview.cpp:2650
> #5  0x00007fffdfe370f9 in QTableView::updateGeometries() (this=this@entry=0x555558f1d280) at itemviews/qtableview.cpp:2138
> #6  0x00007fffe0a5b8e8 in QTableViewWrapper::updateGeometries() (this=0x555558f1d280)
>     at /usr/src/debug/pyside-qt4.8+1.2.2/x86_64-redhat-linux-gnu/PySide/QtGui/PySide/QtGui/qtableview_wrapper.cpp:2060
> #7  0x00007fffdfe2bea6 in QTableView::columnCountChanged(int, int) (this=0x555558f1d280) at itemviews/qtableview.cpp:2015
[...]

PR #277, opened 2016-04-14
Fix qt tableeditor segfault on editing close

jdmarch:
From https://github.com/enthought/traitsui/pull/276, plus the cleanup suggested by @corranwebster  in comments there
@kitchoi @dpinte


jonathanrocher:
This fix also removed a segfault in one of my projects where a TableEditor was segfaulting on tearDown. So +1 for me...

PR #279, opened 2016-04-15
Update documentation for Release 5.1.0

kitchoi:
Documentation for release 5.1.0 using the Enthought Sphinx theme.


traitsui_user_manual/factories_basic.html
```diff
@@ -95,8 +117,7 @@
 </tr>
 <tr class="field-even field"><th class="field-name">Default for:</th><td class="field-body">Array, CArray (if 2-D)</td>
 </tr>
-<tr class="field-odd field"><th class="field-name" colspan="2">Optional parameter:</th></tr>
-<tr class="field-odd field"><td>&nbsp;</td><td class="field-body"><em>width</em></td>
+<tr class="field-odd field"><th class="field-name">Optional parameter:</th><td class="field-body"><em>width</em></td>
```
kitchoi: I changed my local `distutil.html4css1` "field-name-limit" to 20 in order to remove the `colspan="2"` here.  It is a hack and I decided to do so because otherwise the page would look like this under the new theme:
screen shot 2016-04-15 at 15 53 34
Now it would look like this
screen shot 2016-04-15 at 15 55 49
The old field-name-limit (and hence colspan=2) had made sense with the old theme but not with the new one.  Not sure if we want to modify the theme or add setting for `field-name-limit`.  I did some preliminary research for the latter and it does not seem trivial.

kitchoi: I meant `docutils.html4ccs1`, not `distutil.html4ccs1`

PR #283, opened 2016-04-27
Do not destroy children widgets on ui.dispose()

pankajp:
Qt automatically manages destruction of child widgets when the
parent is destroyed. We do not have to explicitly iterate over
child widgets and delete them. In fact it causes occasional
segfaults on exit when the deleted child is attempted to be
deleted again when the parent is destroyed on exit.


traitsui/ui.py
```diff
@@ -264,7 +264,7 @@ def finish ( self ):
         """
 
         # Reset the contents of the user interface
-        self.reset( destroy = True )
+        self.reset( destroy = False )
```
itziakos: Does this work with wx?

pankajp: I think it should work. The wx `DestroyChildren` doc says it is automatically called by the destructor. @RobinD42 may have a better insight

RobinD42: Yes, `wx` also destroys all children when the parent widget is destroyed.

corranwebster:
Are we clear to merge this?

corranwebster:
Still getting OS X PyQt segfaults, which is not related to this, I think.  I am getting close to making PyQt tests on OS X expected failures.

PR #287, opened 2016-05-19
TableEditor column with progress bar

corranwebster:
Probably want a little more customization, but this is the basics of `TableEditor` column which displays progress bars.

screen shot 2016-05-19 at 4 27 59 pm image

### DIFF
--- a/traitsui/extras/progress_column.py
+++ b/traitsui/extras/progress_column.py
@@ -25,7 +25,7 @@ from traitsui.table_column import ObjectColumn
 if ETSConfig.toolkit == 'qt4':
     from traitsui.qt4.extra.progress_renderer import ProgressRenderer
 else:
-    raise NotImplementedError("No checkbox renderer for backend")
+    raise NotImplementedError("No pregress column renderer for backend")
 
 
 class ProgressColumn(ObjectColumn):
--- a/traitsui/extras/progress_column.py
+++ b/traitsui/extras/progress_column.py
@@ -16,10 +16,8 @@
 
 """ A column class for for the TableEditor that displays progress bars. """
 
-from __future__ import absolute_import
-
 from traits.etsconfig.api import ETSConfig
-from traits.api import Str
+from traits.api import Bool, Int, Str
 
 from traitsui.table_column import ObjectColumn
 
@@ -31,9 +29,24 @@ else:
 
 
 class ProgressColumn(ObjectColumn):
+    """ A column which displays trait values as a progress bar
 
-    # Format string to apply to column values:
+    Progress values must be an integer value between the maximum and minimum
+    values.  By default it is assumed to be a percentage.
+    """
+
+    #: Format string to apply to column values.
     format = Str('%s%%')
+
+    #: The minimum value for a progress bar.
+    minimum = Int(0)
+
+    #: The maximum value for a progress bar.
+    maximum = Int(100)
+
+    #: Whether or not to display the text with the progress bar.
+    #: This may not display with some progress bar styles, eg. on OS X.
+    text_visible = Bool(True)
 
     def __init__(self, **traits):
         super(ProgressColumn, self).__init__(**traits)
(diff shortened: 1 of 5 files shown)


corranwebster:
Qt only, but I think that this is good for review.

corranwebster:
OK - error happening or not depends on where its run in relation to other tests, so problem with failure to clean-up gui somewhere, I think.

traitsui/qt4/extra/progress_renderer.py
```diff
@@ -0,0 +1,52 @@
...
+# This software is provided without warranty under the terms of the BSD
+# license included in enthought/LICENSE.txt and may be redistributed only
+# under the conditions described in the aforementioned license.  The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+# Thanks for using Enthought open source!
+#
+# Author: Corran Webster
+#------------------------------------------------------------------------------
+
+""" A renderer which displays a progress bar based on a 0.0 to 1.0 value.
```
senganal: IIUC the progress bar is based on a integer value..

traitsui/extras/progress_column.py
```diff
@@ -0,0 +1,74 @@
...
+from traits.etsconfig.api import ETSConfig
+from traits.api import Bool, Int, Str
+
+from traitsui.table_column import ObjectColumn
+
+
+if ETSConfig.toolkit == 'qt4':
+    from traitsui.qt4.extra.progress_renderer import ProgressRenderer
+else:
+    raise NotImplementedError("No checkbox renderer for backend")
```
senganal: s/checbox/progressbar ?

traitsui/qt4/extra/progress_renderer.py
```diff
@@ -0,0 +1,52 @@
...
+    """ A renderer which displays a progress bar for a True value and an
+        unchecked box for a false value.
+    """
+
+    #---------------------------------------------------------------------------
+    #  QAbstractItemDelegate interface
+    #---------------------------------------------------------------------------
+
+    def paint(self, painter, option, index):
+        """ Reimplemented to paint the checkbox.
```
senganal: s/checkbox/progressbar

traitsui/qt4/extra/progress_renderer.py
```diff
@@ -0,0 +1,52 @@
...
+from pyface.qt import QtCore, QtGui
+
+# ETS imports
+from traitsui.qt4.table_editor import TableDelegate
+
+
+class ProgressRenderer(TableDelegate):
+    """ A renderer which displays a progress bar for a True value and an
+        unchecked box for a false value.
+    """
```
senganal: incorrect docstring ?

traitsui/qt4/extra/progress_renderer.py
```diff
@@ -0,0 +1,52 @@
...
+
+    #---------------------------------------------------------------------------
+    #  QAbstractItemDelegate interface
+    #---------------------------------------------------------------------------
+
+    def paint(self, painter, option, index):
+        """ Reimplemented to paint the checkbox.
+        """
+        # Get the column and object
```
senganal: perhaps the baseclass method needs to be called first to set the background color etc  ?

corranwebster: Not needed, I believe - this is basically a Python version of the code here: http://doc.qt.io/qt-4.8/qabstractitemdelegate.html#details

The style's `drawControl` should handle everything.

senganal: Agreed.. i tried calling the base class method and all it does is to print the text label (can a hacky way to print the text :) ).. But, background color is still not handled but that is not important..

senganal:
One other comment: It might be useful to enforce  that the value lies in between the minimum and maximum.. perhaps raise an error/warning and truncate the value to the nearest boundary..

otherwise LGTM(e)..

corranwebster:
It looks like Qt clips the values if they are out of range (or at least displays something sensible), so I don't think we need to clip, warn or error (which would be the three possible options).

Based on that, I'm going to merge shortly.

PR #288, opened 2016-05-24
Feature: expose update refresh df editor

jonathanrocher:
Exposing the `update` and `refresh` attributes of the `TabularEditor` in the `DataFrameEditor` so that refresh of the view can be triggered when changes are made to a `DataFrame` being viewed in a TraitsUI application.

That allows to see the UI update when clicking the button in this type of application:

```
import pandas as pd
from traits.api import HasTraits, Instance, Button, Event
from traitsui.ui_editors.data_frame_editor import DataFrameEditor

class DFViewer(HasTraits):
    df = Instance(pd.DataFrame)
    button = Button("+1")
    update_view = Event

    view = View(
        Item("df", editor=DataFrameEditor(update="update_view")),
        Item("button"),
        resizable=True
    )

    def _df_default(self):
        return pd.DataFrame({"a": [1, 2, 3], "b": list("abc")})

    def _button_fired(self):
        self.df.iloc[0, 0] += 1
        self.update_view = True

DFViewer().configure_traits()
```


jonathanrocher:
In the current form, if a `ModelView` contains the `DataFrameEditor` and a model with an event `event_name` to listen to, the `update` attribute should be set to `event_name` rather than `model.event_name`. That is due to the `object.object` prefix that gets prepended to the value passed, which resolves to the model rather than the view. 

That feels wrong to me, but I am not sure what the best way to fix this is. I could hack it and remove any `model.` prefix, but it sounds too hacky for my taste. Suggestions @corranwebster ?

corranwebster:
Not sure why AppVeyor is failing - it didn't even get a good version of Pyface loaded.

You intuition seems right that it probably should be connecting to `model.event_name` if things were being consistent.  Not sure what the right fix is off the top of my head: it may mean fixing something in the `UIEditor` base class, or passing in the "correct" context to the `edit_traits` call.

Other than that, basically looks OK.  Let me play with the code a bit before I give the final blessing.

jonathanrocher:
@corranwebster Want to give a final blessing on this?

corranwebster:
Looking things over:

* it would be nice to have a test that allows for alternative adapters since this PR includes that change
* it would be good to have the tests for refresh and update actually fire the events

corranwebster:
See above for better tests, @jonathanrocher if you could merge into your branch, I think we're then good to merge the whole thing.

jonathanrocher:
@corranwebster Sorry it took me so long to update this PR. I have resolved the conflicts with master. Ready to merge then? cc @jdmarch since you pinged me about this...

corranwebster:
LGTM

jonathanrocher:
Thanks for merging quickly @corranwebster but I realize now that I didn't understand what you wanted me to do, and *your* test improvements PR to *my* branch was not merged. I only resolved conflicts with current enthought master. Sorry about that.

jonathanrocher:
The easiest way to resolve is probably for you to pull enthought:feature/expose_update_refresh_df_editor back out and make a PR to enthought master. Let me know if you prefer to proceed differently, and if I can help (like reverting the merge of my PR, and me merging your PR into mine).

corranwebster:
I've just made the PR from the branch `feature/expose_update_refresh_df_editor` against master.  Assuming tests pass, I'll self-merge shortly.

PR #291, opened 2016-06-20
Fix skip kill-focus events

### LINKED ISSUE #282: Windows: Unable to move cursor to previous field with wx as backend.
Bug reported from user on zendesk. In the example below, after moving the cursor to the second field the cursor cannot be moved back to the previous field. Using `<tab>` and `shift+<tab>` allow the fields to be modified, but the cursor remains fixed. This seems to be specific to Windows and wx 3.0. Downgrading to wx 2.8.10 fixes the bug.

Tested with traits 4.5.0 and traitsui 5.1.0 & 5.0.0

```
from traits.etsconfig.api import ETSConfig
ETSConfig.toolkit = 'wx'

from traits.api import Int, Float, Property, HasTraits


class Test(HasTraits):
    ''' This recreates an apparent bug I have observed when using Int, Float,
        and Str traits. The behavior is as follows:
            1) number trait can be changed from ui the first time
            2) interval trait can be changed
            3) number cannot be changed again because the cursor is stuck in
               the interval field
    '''
    
    #Define Traits
    number = Int(1, label='Number')
    interval = Float(0.1, label='Interval')
    _range = Property(Float, depends_on=['number', 'interval'])
    
    view = View(
                Item('number'),
                Item('interval'),
                Item('_range', style='readonly'),
                title='Test'
                )
    
    def _get__range(self):
        _range = (self.number-1) * self.interval
        return _range
    
if __name__ == '__main__':
    test = Test()
    test.configure_traits()
```

RobinD42:
`Skip()` should always be called for kill-focus events of native widgets, especially `wx.TextCtrl`s.

Not calling Skip for these events is one of those things that may accidentally work in some cases, but was never the correct thing to do unless it is a generic widget or you are overriding all the native behavior.

See https://enthought.zendesk.com/agent/tickets/63321 for an example of breakage with wxPython 3.0.

Fixes #282

### DIFF
--- a/traitsui/wx/color_editor.py
+++ b/traitsui/wx/color_editor.py
@@ -248,6 +248,7 @@ class CustomColorEditor(BaseSimpleEditor):
         """ Handles the user changing the contents of the edit control.
         """
         if not isinstance(event, wx._core.CommandEvent):
+            event.Skip()
             return
         try:
             # The TextCtrl object was saved as self._text_control in init().
--- a/traitsui/wx/editor_factory.py
+++ b/traitsui/wx/editor_factory.py
@@ -145,6 +145,8 @@ class TextEditor ( Editor ):
     def update_object ( self, event ):
         """ Handles the user changing the contents of the edit control.
         """
+        if isinstance(event, wx.FocusEvent):
+            event.Skip()
         try:
             self.value = self.control.GetValue()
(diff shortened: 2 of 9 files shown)


corranwebster:
@RobinD42  are there similar fixes needed in Pyface?

Other than that LGTM.

PR #295, opened 2016-07-22
Fix TypeError in Python3

bergtholdt:
Fixes "TypeError: list indices must be integers or slices, not float"

``incr`` is a list of floats due to its computation. In python3 floats cannot be used for indexing.


corranwebster:
Thanks for the contribution!  The Appveyor failure looks like it might be due to same issue that https://github.com/enthought/pyface/pull/226 fixed for Pyface.  Not sure that I will have time to apply fix to TraitsUI today, though.

PR #296, opened 2016-08-02
Fix Qt table model drop mime typo

jthacker:
Dragging and dropping fails for Qt tables, this clearly looks like a typo.


corranwebster:
Yep, agreed!  Good catch and thanks for the contribution.

Appveyor test failures are due to a change in their infrastructure which we haven't had a chance to adapt to yet, so don't worry about that.

I'll merge this when someone (probably me) has had a chance to fix the Appveyor test infrastructure.

PR #304, opened 2016-09-02
More Pyface in TraitsUI

corranwebster:
This is the companion PR to enthought/pyface#221 from the TraitsUI side.

For the time being we just import everything from Pyface into the modules where they previously were in TraitsUI to minimize any issues with compatibility.


itziakos:
can we add a deprecation warning for using these end points instead of pyface?

corranwebster:
For the `image` stuff, yes, but in that case we should also fix up the imports within traitsui to point in the correct place.

I don't think there's a way to deprecate the other imports since the modules still have other content which hasn't been shifted to Pyface.

corranwebster:
@itziakos I think this PR now satisfies the requested change.

PR #311, opened 2016-10-13
Fix integer division for enum editor on Python 3

### LINKED ISSUE #306: Python 3 traitsui/traitsui/qt4/enum_editor.py float index from devision
There is an integer division for the` incr` variable that is now a float in Python3 which causes an exception when trying to index into a list.  I checked the Python3 feature branch and it isn't corrected there (and doesn't have the // division that is shown below for the `rows` variable.

```
        rows    = (n + cols - 1) // cols
        if self.row_major:
            incr = [ 1 ] * cols
        else:
            incr = [ n / cols ] * cols
            rem  = n % cols
            for i in range( cols ):
                incr[i] += (rem > i)
            incr[-1] = -(reduce( lambda x, y: x + y, incr[:-1], 0 ) - 1)
```

corranwebster:
Fixes #306

### DIFF
--- a/traitsui/qt4/enum_editor.py
+++ b/traitsui/qt4/enum_editor.py
@@ -429,7 +429,7 @@ class RadioEditor(BaseEditor):
         if self.row_major:
             incr = [1] * cols
         else:
-            incr = [n / cols] * cols
+            incr = [n // cols] * cols
             rem = n % cols
             for i in range(cols):
                 incr[i] += (rem > i)


VenturaFranklin:
I see this was fixed, when will the next release be?

corranwebster:
Soon (next week or two?).  Currently working on Pyface issues, then will look at state of TraitsUI, then release.

PR #318, opened 2016-11-24
Fix string comparision in CheckListEditor

### LINKED ISSUE #316: Checklist editor throws an exception when selecting value from drop-down
When using future imports for unicode literals, selection using the drop-down fails..

Here is the line of code that causes this issue: https://github.com/enthought/traitsui/blob/8931370f194226b6936afee30b922d0bd86e7c3d/traitsui/qt4/check_list_editor.py#L306

senganal:
Closes #316


corranwebster:
LGTM

PR #320, opened 2016-12-09
Fix comment saying vertical springs don't work

achabotl:
The `using_springs` demo mentioned that using a Spring didn't move
widgets down, but it actually does. This updates the statement to say
that it works.

"Proof" that it works:

screen shot 2016-12-09 at 15 15 47

### DIFF
--- a/examples/demo/Useful/using_springs.py
+++ b/examples/demo/Useful/using_springs.py
@@ -50,8 +50,7 @@ class SpringDemo(HasTraits):
                    'with springs after the 2nd and 4th '
                    'buttons'),
             spring,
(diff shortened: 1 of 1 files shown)


corranwebster:
I seem to recall that this is backend-dependent.  Try running with `ETS_TOOLKIT=wx`.

achabotl:
What about something like "'Spring in vertical group moves widget down (does not work on all platforms when using the Wx backend)."?

corranwebster:
I think "(does not work on Wx backend)" is probably good enough - I'm pretty sure that it will also fail on windows and linux, nothing special about OS X here, although happy to be proven wrong.

PR #321, opened 2016-12-10
EDM tests

corranwebster:
Run tests using EDM to handle setting up clean environments.  This PR removes the old tests, the upshot of which is:

- we lose testing on 3.4, including any PySide backend testing under Python 3
- we gain testing on 3.5, including PyQt testing under Python 3
- we gain testing on macOS under exactly same conditions as we have for Linux
- we gain an easy way to test in a controlled environment on a developer system
- we gain tests that are run on solid, tested builds from Enthought and with a consistent set of dependencies.

The system is currently built on top of `pyinvoke` to run the EDM commands.  A developer can run a clean set of tests for a particular runtime and toolkit with:
```
invoke test_clean --runtime=... --toolkit=...
```

The following tasks still need to be done:

- [x] clean up of scripts and files from old build process
- [x] some developer documentation

At this point I'd like serious review, because if this is deemed to be a good approach, it would make sense to replicate throughout the ETS stack.  In particular, if @itziakos @mdickinson @sjagoe @jwiggins @rkern and/or @cournape have any comments or suggested improvements, this would be an opportune time to comment.


corranwebster:
This is probably worth a review as far as approach goes.  I am wondering if the `tasks.py` should be broken up into setup and teardown and called from the appropriate section of the .travis.yml.

I want to get the patterns right, as this will be a template for other projects once we have it working as desired.

Also, we have what look to be some genuine test failures probably because we are running in subtly different environments than we were previously.

jwiggins:
:heart_eyes: Glad to see this making it out into the ETS projects!

corranwebster:
No reason in practice that we couldn't do Windows via this means, but:

- I didn't have a handy Windows VM to test on
- we'd need to update the Appveyor stuff, which I haven't touched deeply ( @itziakos is master of that domain)
- there may be unix-isms in the `tasks.py` that need smoothing over for Windows.

.travis.yml
```diff
@@ -1,47 +1,51 @@
...
+    - language: generic
+      os: osx
+      env: RUNTIME=3.5 TOOLKIT=pyqt
   allow_failures:
-    - env: ETS_TOOLKIT=wx
+    - env: RUNTIME=2.7 TOOLKIT=wx
+  fast_finish: true
 cache:
   directories:
-    - $HOME/.cache
```
itziakos: I would still keep `.cache`  which is the default cache directory place for both pip and edm.

.travis.yml
```diff
@@ -1,47 +1,51 @@
...
-    - $HOME/.ccache
+    - $HOME/.edm
 before_install:
-  - ccache -s
-  - pip install --upgrade pip
-  - export PATH=/usr/lib/ccache:${PATH}
-  - if [[ ${TRAVIS_PYTHON_VERSION} == "3.4" && ${ETS_TOOLKIT} == "qt4" ]]; then ./build_pyside_wheel.sh; fi
   - export DISPLAY=:99.0
-  - sh -e /etc/init.d/xvfb start
+  - ./travis_ci_install.sh
```
itziakos: I think `travis-ci-bootstrap.sh` could be better as a name.

travis_ci_install.sh
```diff
@@ -0,0 +1,18 @@
+if [[ ${TRAVIS_OS_NAME} == "osx" ]]
+then
+    # download and install EDM
+    wget https://package-data.enthought.com/edm/osx_x86_64/1.3/edm_1.3.0.pkg
+    sudo installer -pkg edm_1.3.0.pkg -target /
+else
+    # set up XVFB for GUI tests
```
itziakos: Since we are using a separate python runtime it is good practice to deactivate the default virtual environment setup on linux build nodes.

.travis.yml
```diff
@@ -1,47 +1,51 @@
...
+      env: RUNTIME=3.5 TOOLKIT=pyqt
   allow_failures:
-    - env: ETS_TOOLKIT=wx
+    - env: RUNTIME=2.7 TOOLKIT=wx
+  fast_finish: true
 cache:
   directories:
-    - $HOME/.cache
-    - $HOME/.ccache
+    - $HOME/.edm
```
itziakos: This is not the default edm cache. With this we actually cache the edm environments which is not a good idea. Using the egg cache is fast enough

tasks.py
```diff
@@ -0,0 +1,217 @@
...
+        # environment doesn't exist, so create it and populate
+        print("Environment not found, creating...")
+        install(ctx, runtime, toolkit, environment, edm)
+
+    envvars = environment_vars.get(toolkit, {}).copy()
+
+    # run tests
+    with do_in_tempdir(files=['.coveragerc'], capture_files=['.coverage']):
+        run_in_env(ctx, environment,
+                   "coverage run -m nose.core -v traitsui.tests",
```
itziakos: Does coverage support the `u` option for unbuffered output? If not then I would suggest to set the related python environment variable.

.travis.yml
```diff
@@ -1,47 +1,38 @@
...
+    - os: osx
+      env: RUNTIME=3.5 TOOLKIT=pyqt
   allow_failures:
-    - env: ETS_TOOLKIT=wx
+    - env: RUNTIME=2.7 TOOLKIT=wx
+  fast_finish: true
 cache:
   directories:
     - $HOME/.cache
-    - $HOME/.ccache
```
itziakos: we should keep `ccache` since it is possible that some of the source despondencies might need a c compiler (e.g. cython code).

.travis.yml
```diff
@@ -1,47 +1,38 @@
...
 before_install:
-  - ccache -s
-  - pip install --upgrade pip
-  - export PATH=/usr/lib/ccache:${PATH}
-  - if [[ ${TRAVIS_PYTHON_VERSION} == "3.4" && ${ETS_TOOLKIT} == "qt4" ]]; then ./build_pyside_wheel.sh; fi
   - export DISPLAY=:99.0
-  - sh -e /etc/init.d/xvfb start
+  - ./travis_ci_bootstrap.sh
```
itziakos: A minor suggestion how about `travis_ci_bootstrap_edm`

travis_ci_bootstrap.sh
```diff
@@ -0,0 +1,18 @@
+if [[ ${TRAVIS_OS_NAME} == "osx" ]]
+then
+    # download and install EDM
+    wget https://package-data.enthought.com/edm/osx_x86_64/1.3/edm_1.3.0.pkg
+    sudo installer -pkg edm_1.3.0.pkg -target /
+else
+    # set up XVFB for GUI tests
+    sh -e /etc/init.d/xvfb start
```
itziakos: I would still keep this out of the bootstrap script. In some cases it is not necessary and I have seen cases one might want to adjust the parameters.

README.rst
```diff
@@ -68,5 +65,18 @@ Prerequisites
 
 If you want to run traitsui, you must also install:
 
-* `traits <https://github.com/enthought/traits>`_
+* `traits <https://github.com/enthought/traits>`__
```
itziakos: why two `_`

corranwebster: Because then it is not a "named link" in restructured text.  This was causing PyPI to not render the text correctly, because we had two "traits" named links in the README.

Yes, it's obscure.  Took some Googling to work out what was going wrong.  Try running `python setup.py check --restructuredtext -s` with and without to see the difference.

PR #327, opened 2016-12-17
Fix Wx toolkit test failures

corranwebster:
We should at least avoid timeouts now, thanks to https://github.com/enthought/pyface/pull/243 and in particular https://github.com/enthought/pyface/pull/243/commits/4ba608473d9314aae6620020f8eb162befc0fe54

To do:

- [ ] fix any other test failures


corranwebster:
The root of the remaining test failure is an incompatibility in Pyface Toolbars introduced by https://github.com/enthought/pyface/issues/223 see https://github.com/enthought/pyface/issues/246

corranwebster:
Remaining failures are specific to Linux and WxPython 2.x - need time to set up a linux environment to investigate what's going on.

PR #338, opened 2017-02-06
Improvements to ListEditor

corranwebster:
This PR fixes a personal pet peeve about the look and feel of the `ListEditor` and attempts to bring it kicking and screaming into the Century of the Fruitbat.

This PR provides:

* replace the menu icon with the now-standard "hamburger menu" icon
* move the menu button to the right, so it is less "front and center".
* tighten up the layout in Qt
* improve the positioning of the popup menu

This now looks like:
screen shot 2017-02-06 at 5 40 45 pm
screen shot 2017-02-06 at 5 41 06 pm


pankajp:
I somehow do not like the new "hamburger" icons. I always thought it was a "drawer" icon, because it usually shows up an sliding animated navigation widget in mobile apps, though firefox also uses it for its menu. I guess seeing multiple of them in a same view is what is causing the uneasy feeling. Anyway i am most likely in the minority here, so nothing to do here, an it is definitely better than the outdated icon it replaces

corranwebster:
I'm not a huge fan of the hamburger menu icon myself, but it is reasonably well understood.  I had considered using vertical dots ⋮ like Chrome does instead, but couldn't get the icon to look good.  If opinion favours it, I could have another go at the icon.

PR #345, opened 2017-02-10
Group builds in travis

itziakos:
This PR goups the builds in travis based on the used runtime to reduce the cache size

In more details

- There are 6 ( < 9) builds now
- The retrieve files functionality now support glob patterns
- Coverage is combined at the end of the build so that we still get the right combined coverage in codecov

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
     - env: RUNTIME=2.7 TOOLKITS="wx"
     - env: RUNTIME=3.5 TOOLKITS="null pyqt"
     - os: osx
-      env: RUNTIME=2.7 TOOLKITS="null pyqt pyside wx"
+      env: RUNTIME=2.7 TOOLKITS="null pyside wx"
     - os: osx
       env: RUNTIME=2.7 TOOLKITS="pyqt"
     - os: osx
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -2,40 +2,32 @@ language: generic
...
+      env: RUNTIME=2.7 TOOLKITS="pyqt"
     - os: osx
-      env: RUNTIME=2.7 TOOLKIT=pyside
-    - os: osx
-      env: RUNTIME=2.7 TOOLKIT=wx
-    - os: osx
-      env: RUNTIME=3.5 TOOLKIT=null
-    - os: osx
-      env: RUNTIME=3.5 TOOLKIT=pyqt
+      env: RUNTIME=3.5 TOOLKITS="null pyqt wx"
```
corranwebster: Wx is not supported under Python 3.

corranwebster:
Wx is currently an expected failure on 2.7, and not supported in 3.5, so the test matrix should be adjusted accordingly.  Otherwise the basic idea seems fine.

itziakos:
@corranwebster pr updated please have another look

corranwebster:
LGTM

PR #350, opened 2017-04-07
Fix PyQt5 compatibility

pankajp:
Move PyQt5 out of allowed failures on travis-ci and fix a few issues

### DIFF
--- a/traitsui/qt4/table_editor.py
+++ b/traitsui/qt4/table_editor.py
@@ -38,6 +38,12 @@ from .table_model import TableModel, SortFilterTableModel
 
 is_qt5 = QtCore.__version_info__ >= (5,)
 
+if is_qt5:
+    def set_qheader_section_resize_mode(header):
+        return header.setSectionResizeMode
+else:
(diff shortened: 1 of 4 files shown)


corranwebster:
Do you feel comfortable that the Qt5 tests won't segfault now?

pankajp:
Oh, is that failure intermittent? I fixed a specific segfault, and it passed twice in a row so we can try our luck and the next time pyqt5 segfaults we move it to allowed failures again.

corranwebster:
Yes, failure was intermittent, but if you've fixed a particular segfault it's probably worth a shot at seeing if it is stable.

traitsui/qt4/table_editor.py
```diff
@@ -1251,3 +1265,11 @@ def _update_filter_list(self):
         filters = self.filters
         self.filters = []
         self.filters = filters
+
+if is_qt5:
```
corranwebster: Minor style suggestion: put this at the top of the file (possibly immediately after the definition of `is_qt5`).

PR #352, opened 2017-04-26
fix for use with qt5, QPixmapCache

geggo:
Hi,

this PR fixes failure with Qt5, e.g. for examples/demo/Standard_Editors/ListEditor_demo.py

With PyQt5 (tested with 5.6.0 from conda) QPixmapCache has a slightly different API.


rkern:
Thanks! Looking at the PyQt5 history, the change at least dates back to 5.1, so I think this should be safe.

PR #363, opened 2017-06-16
Fix TabularEditor.drag_move to properly allow drag-copy 

### LINKED ISSUE #351: TabularEditor doesn't properly allow drag-copy
This looks like a bug here: https://github.com/enthought/traitsui/blob/master/traitsui/qt4/tabular_model.py#L315

I think that the fix is simple: return `QtCore.Qt.MoveAction | QtCore.Qt.CopyAction`.

kitchoi:
Fix #351 

~WIP for I would like to investigate:~
   ~(1) Adding tests~
   ~(2) Possibly same issue with TableEditor and ListStrEditor~

### DIFF
--- a/traitsui/editors/tabular_editor.py
+++ b/traitsui/editors/tabular_editor.py
@@ -152,7 +152,7 @@ class TabularEditor(BasicEditorFactory):
 
     # Are 'drag_move' operations allowed (i.e. True), or should they always be
     # treated as 'drag_copy' operations (i.e. False):
-    drag_move = Bool(False)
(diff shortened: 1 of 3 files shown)


corranwebster:
@kitchoi Would you be OK with this being merged in the next release, even though not as complete as you would like?

kitchoi:
Yes I agree this would be worth merging.  Removing WIP in the title.

kitchoi:
Yes I agree this would be worth merging.  Removing WIP in the title.

corranwebster:
Thanks!

PR #369, opened 2017-06-27
Call unittest.skip correctly

jwiggins:
This is causing CI failures for Enable currently: https://travis-ci.org/enthought/enable/jobs/247587609


jwiggins:
Thanks for reviewing!

PR #380, opened 2017-07-15
DOC: Document Item, Group, and View in the user manual

achabotl:
- Replace the tables describing the arguments to Item, Group and View, with something more definition-like. The documentation for each argument comes from our teaching documentation, as well as from the trait comments from the code. 

- Add more Item subclass definitions and reformat them as a table, like the Group subclasses.

### DIFF
--- a/docs/source/traitsui_user_manual/custom_view.rst
+++ b/docs/source/traitsui_user_manual/custom_view.rst
@@ -295,12 +295,12 @@ Window display
       Status bar items to add to the view's status bar. The value can be:
 
       - **None**: No status bar for the view (the default).
-      - string: Same as [ StatusItem( name = string ) ].
-      - StatusItem: Same as [ StatusItem ].
-      - [ [StatusItem|string], ... ]: Create a status bar with one field for
+      - string: Same as ``[StatusItem(name=string)]``.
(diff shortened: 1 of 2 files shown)


docs/source/traitsui_user_manual/custom_view.rst
```diff
@@ -256,66 +256,151 @@ Other View Attributes
...
+   scrollable: bool
+      Can the user scroll the view? If set to True, window-level scroll bars
+      appear whenever the window is too small to show all of its contents at
+      one time. If set to False, the window does not scroll, but individual
+      widgets might still contain scroll bars.
+   statusbar:
+      Status bar items to add to the view's status bar. The value can be:
+
+      - **None**: No status bar for the view (the default).
+      - string: Same as [ StatusItem( name = string ) ].
```
corranwebster: Probably want to quote `[StatusItem(name=string)]` so that it appears as code in final doc, and remove idiosyncratic code style here and below

corranwebster:
Other than the comments about some included code needing markup, looks good to me.

achabotl:
@corranwebster I fixed the code quoting and the spacing.

PR #384, opened 2017-08-09
Improve TraitsUI Action handling

### LINKED ISSUE #383: Can't use the same Handler method with multiple Actions
When an `Action` is added to a menu in a TraitsUI GUI, things are set up so that the `BaseDialog` subclass is a controller for the `ActionItem`.  This means that the control flow when activating an action in a TraitsUI view is:

- the `BaseDialog`'s `perform` method is called, which calls the `_perform` method with some support for undo/redo.
- the `_perform` method tries to find an appropriate method on the `handler` or the `object` that matches the `action` trait of the `Action` and if successful, calls that with the `ui_info` _but not the `Action`_.
- otherwise it calls the `Action`'s perform method (with the wrong signature!), and it acts like any regular Pyface `Action` would, but it has none of the TraitsUI data available to it.

This makes some code re-use difficult.  Consider a situation where you have a set of similar Actions which (for example) all want to perform some computation with the `model` (or other attribute) of a `ModelView` handler.  At the moment you have two options:

- write a separate method for each `Action` on the `ModelView`.  This works, but isn't very extensible (you can't have third-party code adding new actions very easily).
- provide each `Action` with the required context from the `ModelView` when initializing the GUI, _and keep it synchronized with any changes_.  In practice, this means passing the handler to the `Action`, but you lose access to the `ui_info`.

What you would like to have in this situation is have one method in the handler be called by all of these actions, extract the common state that the actions need, and dispatch them somehow (eg. call a method on the action with the common state as arguments).

There are three possible fixes to this (which aren't mutually exclusive):

- pass the `Action` to the `Handler` methods as a second argument after the `ui_info`.   Unfortunately this breaks all existing handlers unless we probe the handler signature.  There may be some way to temporarily attach the `Action` to the `ui_info`, but that seems hacky.
- pass the `ui_info` to the `Action`'s `perform` method.  This breaks all existing TraitsUI `Actions` which use `perform` (although there is an argument that they are already broken because they _should_ expect an `event` argument).
- have the `_perform` method in `BaseDialog` immediately call a `perform` method on the handler, which by default will follow something like the current logic that is in the `_perform` method, but which can be overridden by `Handler` subclasses to do different things.  This doesn't break anything if done correctly.

The last option also has the advantage of moving duplicated code out of the toolkit code (Wx and Qt `BaseDialog` classes) and into generic code (the `Handler` classes).

corranwebster:
This adds a `perform` method to the base `Handler` class and shifts the common `Action` perform dispatch logic out of the toolkit `BaseDialog` classes.

This should be largely backwards compatible: obviously any existing user `Handler` subclass with a `perform` will have a problem, and there are some subtle differences in what methods will be checked (eg. `_<action.name>_clicked` can be called for menu toolbar items and in addition to buttons, if `action.action` is not defined).

Fixes #383 and a TODO left open in #56.

Also cleans up a couple of other issues (eg. `ui_base.ButtonEditor` was not correctly initialised)

### DIFF
--- a/traitsui/handler.py
+++ b/traitsui/handler.py
@@ -25,6 +25,8 @@
 
 from __future__ import absolute_import
 
+from inspect import getargspec
+
 from .toolkit import toolkit
 
 from .help import on_help_call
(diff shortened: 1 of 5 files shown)


corranwebster:
@itziakos if you have time to look at this, I'd appreciate it.

traitsui/handler.py
```diff
@@ -240,6 +242,80 @@ def show_help(self, info, control=None):
...
+            The UIInfo associated with the view, or None.
+
+        Returns
+        -------
+        handlers : list
+            A list of objects that may potentially have action methods on them.
+        """
+        handlers = [self]
+        if info is not None:
+            additional_objects = ['object', 'model']
```
itziakos: Could this be a handler trait (e.g. `action_handlers=List(str)`)

traitsui/qt4/ui_base.py
```diff
@@ -106,26 +103,18 @@ class ButtonEditor(Editor):
     #-------------------------------------------------------------------------
 
     def __init__(self, **traits):
```
itziakos: Do we really need to override the init?

corranwebster: Yes - the super-superclass's `__init__` method expects things which aren't available here.  I am not convinced that this should be a subclass of `Editor`, but I haven't sat down and worked out exactly what is going on here and it is orthogonal to this PR.

traitsui/handler.py
corranwebster: It could be, but I didn't want to add a new trait to what is a fairly commonly subclassed class.  In the original code this was a module global.  I think it is better to just override `get_perform_handlers` (as is done for `Controller`) if it doesn't suit a particular case.

This is analogous with methods like `trait_context`.

traitsui/qt4/ui_base.py
itziakos: ok

traitsui/handler.py
itziakos: The main reason is that due to the global module variable being remove, legacy code will have to subclasss in order to support the original behavior instead of setting a trait value

PR #386, opened 2017-08-24
Fix value editor icons

### LINKED ISSUE #385: Pyface icons not found by the qt TreeEditor
The qt `TreeEditor` doesn't find the zipped [icons](https://github.com/enthought/pyface/blob/master/pyface/image/library/icons.zip) from pyface. Resulting in a `ValueEditor` which doesn't work properly, because it depends on those icons.

corranwebster:
This fixes #385.  Includes some drive-by clean-up.

### DIFF
--- a/traitsui/qt4/tree_editor.py
+++ b/traitsui/qt4/tree_editor.py
@@ -23,7 +23,8 @@ import logging
 
 from pyface.qt import QtCore, QtGui
 
-from pyface.resource_manager import resource_manager
+from pyface.api import ImageResource
+from pyface.ui_traits import convert_image
 from pyface.timer.api import do_later
 from traits.api import Any, Event
 from traitsui.api import TreeNode, ObjectTreeNode, MultiTreeNode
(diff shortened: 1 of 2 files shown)


corranwebster:
@martinRenou this should fix the issue you saw with the ValueEditor.

traitsui/qt4/tree_editor.py
```diff
@@ -23,7 +23,8 @@
 
 from pyface.qt import QtCore, QtGui
 
-from pyface.resource_manager import resource_manager
+from pyface.api import ImageResource
+from pyface.ui_traits import convert_image
```
martinRenou: I have an `ImportError: No module named ui_traits` here, with `pyface==5.1.0`, which is the latest version apparently.

corranwebster: You'll need Pyface master for current TraitsUI code to work.

martinRenou: Ok thanks

PR #388, opened 2017-09-12
Use trait instead of hardcoding checklist values

rahulporuri:
Personally, this was what I was looking for and I am assuming that
this is what people would want to do after coming across this example
i.e. to be able to dynamically update the checklist values

```
modified:   examples/demo/Standard_Editors/CheckListEditor_simple_demo.py
```

### DIFF
--- a/examples/demo/Standard_Editors/CheckListEditor_simple_demo.py
+++ b/examples/demo/Standard_Editors/CheckListEditor_simple_demo.py
@@ -28,7 +28,12 @@ class CheckListEditorDemo(HasTraits):
     values = List(Str, ['None'])
 
     # Specify the strings to be displayed in the checklist:
-    checklist = List(editor=CheckListEditor(
+    checklist_static = List(editor=CheckListEditor(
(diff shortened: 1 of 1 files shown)


corranwebster:
I'm 50-50 on this.  The case where things don't change dynamically is still important, common and basic, so it is useful to have an example of this use case in the demos.  I agree that showing using a dynamic example is also useful, but this example doesn't really show that it is dynamic (ie. changing the `values` trait changes the options presented).

A better option for the PR would be to add the dynamic example, rather than change the existing example.

corranwebster:
This is better, but to clarify part of the above: you should add some additional UI component that changes the list of allowed values in the trait referred to by `names`, eg. add a trait like `max_value = Enum(7, 8, 9)` and with a change handler that modifies the `values` trait:
```
def _max_value_changed(self, new):
    self.values = [str(i) for i in range(1, self.max_value+1)]
```
and add an item for this to the view.

kitchoi:
@rahulporuri are you still intending to update this PR? Or is this a feature request? If so, perhaps this should be closed and a new issue be open referencing this one?

PR #393, opened 2017-10-11
Fix action buttons in Qt

corranwebster:
#384 broke action dialog buttons on Qt.

This fixes signature of the `perform` method of `traitsui.qt4.ui_base.ButtonEditor` to match what the Qt button's `clicked` signal expects.

### DIFF
--- a/traitsui/qt4/ui_base.py
+++ b/traitsui/qt4/ui_base.py
@@ -153,7 +153,7 @@ class BasePanel(object):
     def is_button(self, action, name):
         """ Returns whether a specified action button is a system button.
         """
-        if isinstance(action, str):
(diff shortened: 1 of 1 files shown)


rkern:
Do we need a test? What problem is this fixing, compared to https://github.com/enthought/traitsui/commit/6c70bceb4127d910d1e0e64f3e37446d2ad05c48 which introduced this signature and added a test that I assumed exercised that signature.

corranwebster:
Probably makes sense to have a test: the focus #384 was on the handler and I had (perhaps foolishly) assumed that buttons were being tested in `test_actions`.  I'll see if I can whip up a test for this in a separate PR.

PR #397, opened 2017-10-12
Avoid PyQt 4.12-1 build

corranwebster:
Current build of pyqt appears to be bad on windows.  Temporary downgrade for CI system.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -102,7 +102,7 @@ dependencies = {
 }
 
 extra_dependencies = {
-    'pyside': {'pyside'},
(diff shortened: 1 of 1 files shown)


etstool.py
```diff
@@ -103,7 +103,7 @@
 
 extra_dependencies = {
     'pyside': {'pyside'},
-    'pyqt': {'pyqt'},
+    'pyqt': {'pyqt < 4.12'},  # FIXME: build of 4.12-1 appears to be bad
```
itziakos: no spaces around `<`

corranwebster: Thanks

Issue #310, opened 2016-10-12
Bug when using a SimpleColorEditor as a cell editor


cfarrow:
I've got a table (via `TableEditor`) that displays colors in one of the columns. When I use the "simple" `ColorEditor`, I get an exception after editing the color. The editor presents a dialog to the user for selecting a color. By the time the dialog is closed (actually before it ever opens), the `factory` attribute on the `SimpleColorEditor` is set to `None`, and this causes and exception in the following code. (https://github.com/enthought/traitsui/blob/master/traitsui/qt4/color_editor.py#L125)

The problem is in how the `TableEditor` cleans up the traitsui editors for its cells. When a cell editor is created, the traits editor is attached to the QT control (https://github.com/enthought/traitsui/blob/master/traitsui/qt4/table_editor.py#L8298). When the control is cleaned up, it also `dispose`s the traits editor (https://github.com/enthought/traitsui/blob/master/traitsui/qt4/table_editor.py#L1125). In the case of the `SimpleEditor`, the `_SimpleField` editor gets cleaned up right before the color dialog, which clears out the `factory` and other traits from the `SimpleColorEditor`. (The cleanup is triggered by the dialog stealing focus from the cell.)

The right way to handle this is not clear. Either QT controls should not clean up traits editors, or cell editors should not create popups. 

The editor cleanup was introduced (or fixed) in 864770bc76c0e15fc8e7f94eb4f2bf52e99b6f97 (a long time ago). @rkern, you made that change. What are your thoughts on this?

rkern:
Hmm, yeah, handling the sequence of events when the editor is a popup needs some thought. If you revert 864770b, does it affect the behavior with the `SimpleColorEditor`?

cfarrow:
Reverting 864770b does not affect the behavior.

cfarrow:
Edit: I forgot to install from source. It *does* correct the behavior.

corranwebster:
Possibly related to #308?

cfarrow:
It seems to be related. However, reverting 864770b does not resolve #308.

PR #399, opened 2017-10-12
Fix for color column

### LINKED ISSUE #302: IndexError: tuple index out of range with ColorColumn
I am getting the following error when running some sample code using `ColorColumn`:

```
IndexError                                Traceback (most recent call last)
/Users/jcorson/Library/Enthought/Canopy/edm/envs/User/lib/python2.7/site-packages/traitsui/qt4/table_model.py in data(self, mi, role)
     87 
     88         if role == QtCore.Qt.DisplayRole or role == QtCore.Qt.EditRole:
---> 89             text = column.get_value(obj)
     90             if text is not None:
     91                 return text

/Users/jcorson/Library/Enthought/Canopy/edm/envs/User/lib/python2.7/site-packages/traitsui/color_column.py in get_value(self, object)
     50         value = getattr( self.get_object( object ), self.name )
     51         if type( value ) is tuple:
---> 52             value = "(%3d, %3d, %3d)" % self._as_int_rgb_tuple( value[:-1] )
     53         elif type( value ) is not str:
     54             value = str( value )

/Users/jcorson/Library/Enthought/Canopy/edm/envs/User/lib/python2.7/site-packages/traitsui/color_column.py in _as_int_rgb_tuple(self, color_values)
     62         return ( int( 255 * color_values[0] ),
     63                  int( 255 * color_values[1] ),
---> 64                  int( 255 * color_values[2] ) )
     65 
```

This is using Traitsui 5.1.0.
The error seems to be caused by `value[:-1]` in color_column.py. Since `_as_int_rgb_tuple_()` returns specific indices in `color_values` is there a reason to not just call  `self._as_int_rgb_tuple(value)` in https://github.com/enthought/traitsui/blob/master/traitsui/color_column.py#L53 ?

Sample Code:

```
from traits.api import HasTraits, Str, Int, RGBColor, List
from traitsui.api import View, Group, Item, TableEditor, ObjectColumn
from traitsui.color_column import ColorColumn


class MyEntry(HasTraits):
    name= Str()
    value=Int(0)
    color=RGBColor()

    entry_view = View(
        Group(
            Item('name'),
            Item('value'),
            Item('color')
        )
    )

my_editor = TableEditor(
    columns=[
        ObjectColumn(name='name'),     
        ObjectColumn(name='value'),
        ColorColumn(name='color', style='readonly')
    ],
    orientation='vertical',
    show_toolbar=True,
    row_factory=MyEntry,
)

class MyData(HasTraits):
    data_list = List(MyEntry)

    view=View(
        Item(
            'data_list',
            editor=my_editor,
            show_label=False
        ), 

    )

if __name__ == '__main__':
    d1 = MyEntry(name='a', value=2, color=(1.0, 0.3, 0.1))
    d2 = MyEntry(name='b', value=3, color=(0.1, 0.0, 0.9))
    data = MyData(data_list=[d1, d2])

    data.configure_traits()
```

corranwebster:
This fixes #302.  Includes a simple smoke test for color column.

### DIFF
--- a/traitsui/color_column.py
+++ b/traitsui/color_column.py
@@ -32,13 +32,13 @@ from traitsui.table_column \
 class ColorColumn(ObjectColumn):
     """ Table column object for Color traits. """
 
-#-- ObjectColumn Overrides -----------------------------------------------
+    #-- ObjectColumn Overrides -----------------------------------------------
 
     def get_cell_color(self, object):
         """ Returns the cell background color for the column for a specified
             object.
         """
-        color_values = getattr(object, self.name + '_')
+        color_values = getattr(object, self.name + '_', (1.0, 1.0, 1.0))
         if isinstance(color_values, tuple):
             tk_color = self._as_int_rgb_tuple(color_values)
         else:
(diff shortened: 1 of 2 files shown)


traitsui/color_column.py
```diff
@@ -48,18 +48,19 @@ def get_cell_color(self, object):
...
     def _as_int_rgb_tuple(self, color_values):
         """ Returns object color as RGB integers. """
-        return (int(255 * color_values[0]),
-                int(255 * color_values[1]),
-                int(255 * color_values[2]))
+        return tuple(int(255 * v) for v in color_values)
+
+    def _int_rgb_tuple_to_str(self, color_values):
+        """ Returns object color as RGB integers. """
+        return '(' + ', '.join('%3d' % x for x in color_values) + ')'
```
itziakos: I would prefer to use `format` for such a task

traitsui/tests/test_color_column.py
```diff
@@ -0,0 +1,54 @@
...
+        Item(
+            'data_list',
+            editor=my_editor,
+            show_label=False
+        ),
+
+    )
+
+@skip_if_null
+def test_color_column():
```
itziakos: Can we include this test method inside a `unittest.TestCase`?

corranwebster:
Issues addressed and some improvements to fix.  Note that in this instance #310 is still an issue and limits the usefulness of this editor, but it should work for `readonly` and `custom` styles as it stands.

PR #400, opened 2017-10-12
Prevent loopback from slider when updating text field of range editor

### LINKED ISSUE #364: RangeEditor changes value entered manually in text box
Given the following example

```python
from traits.api import HasTraits, Float
from traitsui.api import Item, RangeEditor, View


class Slider(HasTraits):
    value = Float(5.)
    traits_view = View(Item('value', editor=RangeEditor(low=0., high=12.)))
    

if __name__ == "__main__":
    s = Slider()
    s.configure_traits()
```

which generates this UI:
screen shot 2017-06-16 at 13 57 52

If enter the value 4
screen shot 2017-06-16 at 13 57 56

and press Tab, I get the following:
screen shot 2017-06-16 at 13 57 44

I assume this has to do with the resolution of the slider, because if I change `high` to 10, and and the value 4 in the text box I get a 4.0. It makes for a strange experience.

corranwebster:
Fixes #364.  Has a test.

### DIFF
--- a/traitsui/tests/editors/test_range_editor_text.py
+++ b/traitsui/tests/editors/test_range_editor_text.py
@@ -22,7 +22,7 @@ A RangeEditor in mode 'text' for an Int allows values out of range.
 from __future__ import print_function
 
 from traits.has_traits import HasTraits
-from traits.trait_types import Int
+from traits.trait_types import Float, Int
 from traitsui.item import Item
 from traitsui.view import View
 from traitsui.editors.range_editor import RangeEditor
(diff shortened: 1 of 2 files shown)


traitsui/tests/editors/test_range_editor_text.py
```diff
@@ -66,6 +78,31 @@ def test_wx_spin_control_editing():
     assert num.number >= 3 and num.number <= 8
 
 
+@skip_if_not_qt4
+def test_avoid_slider_feedback():
```
itziakos: Can we please include these two functions in a unittest.TestCase?

corranwebster: Not when adding to an existing nose-based suite.  At some point it would be good to switch to unittest-based only, but it should be a stand-alone PR.

PR #401, opened 2017-10-12
Item label enable when support

### LINKED ISSUE #389: Item labels are not grayed out when the editor field is disabled.
When a trait editor field is disabled (e.g. using `enabled_when`) the associated label widget should be also `disabled` so that it is obvious to the user that the field is not editable.

*Please note that this feature request is related to #298 because of the way labels are created in traitsui*

corranwebster:
Fixes #389.  With tests.

### DIFF
--- a/traitsui/tests/test_labels.py
+++ b/traitsui/tests/test_labels.py
@@ -115,6 +115,26 @@ class NoLabelResizeTestDialog(HasTraits):
     )
 
 
+class EnableWhenDialog(HasTraits):
+    """ Test labels for enable when. """
+
+    bool_item = Bool(True)
+
+    labelled_item = Str('test')
+
+    unlabelled_item = Str('test')
+
+    traits_view = View(
+        VGroup(
+            Item('bool_item',),
+            Item('labelled_item', enabled_when='bool_item'),
+            Item('unlabelled_item', enabled_when='bool_item', show_label=False),
+        ),
+        resizable=True
+    )
+
+
+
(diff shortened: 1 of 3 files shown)


traitsui/tests/test_labels.py
```diff
@@ -200,6 +220,48 @@ def test_qt_no_labels_on_the_right_bug():
...
+
+    with store_exceptions_on_all_threads():
+        dialog = EnableWhenDialog()
+        ui = dialog.edit_traits()
+
+        labelled_editor  = ui.get_editors('labelled_item')[0]
+        unlabelled_editor  = ui.get_editors('unlabelled_item')[0]
+        nose.tools.assert_is_none(unlabelled_editor.label_control)
+
+        nose.tools.assert_true(labelled_editor.label_control.isEnabled())
```
itziakos: I prefer to have these two test functions into a single `unittest.TestCase`

corranwebster: Some comment as previous PR - the current test uses nose and I don't want to mix nose and unittest in suites.

rkern: You can also put them into a plain-old-class named `TestWhatever`. But I wouldn't worry about "mixing" `unittest` and `nose`. `TestCases` are perfectly good `nose` tests.

corranwebster:
@itziakos Have simplified tests down to just one test function with common code.  Will worry about unittest conversion of test suite in separate PR later.

PR #404, opened 2017-11-06
Check old value only non event/spring objects

VenturaFranklin:
I have a test script: 
```from traits.api import HasTraits, Button
class Parent(HasTraits):
    def __init__(self):
        self.add_trait('say_hello', Button(
            get=self._say_hello_fired,
            set=self._say_hello_fired
            )
                       )
    def extra(self, *args, **kwargs):
        pass
    def _say_hello_fired(self):
        print("hello")
        return self.extra
        return View(
                Item('say_hello',
                     show_label=False),
                )
parent = Parent()
# parent.configure_traits()
parent.say_hello()
```
I want to enable both scripting and GUI control with minimal source code changes. In the example above I want both the button in the GUI when configure_traits is uncommented and then separately the parent.say_hello() method call to work.
This works, however, I get an erroneous call on init to say_hello in the GUI side because in [Editor.__init__ editor.py](https://github.com/enthought/traitsui/blob/512d24196e43f374dd5f7de5f127585eed6e0a85/traitsui/editor.py#L145)
```
...
try:
    self.old_value = getattr( self.object, self.name )
except AttributeError:
    ctrait = self.object.trait(self.name)
    if ctrait.type == 'event' or self.name == 'spring':
        # Getting the attribute will fail for 'Event' traits:
        self.old_value = Undefined
    else:
        raise
...
```
The check if it is an event happens after it checks for old value.
Can this be changed with out screwing things up?:
```
try:
    ctrait = self.object.trait(self.name)
    if ctrait.type == 'event' or self.name == 'spring':
        # Getting the attribute will fail for 'Event' traits:
        self.old_value = Undefined
    else:
        self.old_value = getattr( self.object, self.name )
except AttributeError:
        raise
```

### DIFF
--- a/traitsui/editor.py
+++ b/traitsui/editor.py
@@ -143,14 +143,14 @@ class Editor(HasPrivateTraits):
         """
         HasPrivateTraits.__init__(self, **traits)
         try:
-            self.old_value = getattr(self.object, self.name)
-        except AttributeError:
-            ctrait = self.object.base_trait(self.name)
+            ctrait = self.object.trait(self.name)
             if ctrait.type == 'event' or self.name == 'spring':
                 # Getting the attribute will fail for 'Event' traits:
                 self.old_value = Undefined
             else:
-                raise
+                self.old_value = getattr( self.object, self.name )
+        except AttributeError:
+            raise
 
         # Synchronize the application invalid state status with the editor's:
         self.sync_value(self.factory.invalid, 'invalid', 'from')


corranwebster:
Hi, thanks for contributing.

I am a little confused about what you are trying to do in your example, in particular for the `get` method of the `Button`; the `Button` trait is a subclass of `Event` and so it is write-only.  Similarly, getters, setters and trait change handlers have different signatures and purposes, so trying to combine them together almost certainly won't work.  Also, since a `Button` doesn't hold a value, you can't call it.  If this code is almost working for you, it is I think more by accident.


In more idiomatic Traits, I think I would write your code something like the following:
```
from traits.api import Button, HasTraits

class Parent(HasTraits):

    say_hello_button = Button(label='Say Hello')

    def say_hello(self, *args, **kwargs):
        """ Do whatever was in your 'extra' method. """
        pass

    def _say_hello_fired(self):
        self.say_hello()

        return View(Item('say_hello_button', show_label=False))
```

In particular I think the logic of the code you are changing is fine---basically it is:

- try to get the trait's value
- if it raises an exception, is it because it is an event trait (which includes `Button`) or a spring?  Then the value of the trait is undefined.
- otherwise it was some other unexpected exception, so re-raise

This is just the "better to ask forgiveness than permission" approach (as opposed to "look before you leap").

Finally, for completeness, it may be worthwhile completely separating your UI (the button) from the "say hello" functionality:
```
from traits.api import HasTraits

class Parent(HasTraits):
    def say_hello(self, *args, **kwargs):
        """ Do whatever was in your 'extra' method. """
        pass

# possibly in a different module...
from traits.api import Button, Instance
from traitsui.api import ModelView, View, Item

class ParentModelView(ModelView):
    model = Instance(Parent)
    say_hello = Button()
    def _say_button_fired(self):
         if self.model is not None:
              self.model.say_hello()
    traits_view = View(Item('say_hello', show_label=False))

# non-gui code
parent = Parent()
parent.say_hello()

# gui code
model_view = ParentModelView(model=parent)
model_view.configure_traits()
```

Of course I may be missing the point of what you are trying to do, in which case I apologise.  If this is the case, can you provide some more information about what you are attempting and what you expect to happen.

VenturaFranklin:
I was heavily coupling my model and view. In my example I wanted the view to behave as the model. ie. the button would not need another function to call, and I would not need a whole model separate from the view.


I run into a lot of duplication and "pass through" functions. Is there a better way to do this?
```    
# This is done a lot...
def _say_button_fired(self):
        self.model.say_hello()

def _another_button_fired(self):
        self.model.another_hello()
# ....etc....
```

corranwebster:
If you have a lot of these on a particular class, I was going to suggest a method involving trait metadata, but it appears that there may be a Traits bug in `on_trait_change` when listening to events by metadata.

Because traits inspects listener signatures to decide how to call the listener functions or methods, you will likely have to have some sort of pass-through which adapts from what traits expects to what your functions expect (unless they are truly simple and take no arguments in general).  But you can automate the creation of these pass-through listener functions.  For example, you could do something like this:
```
from traits.api import Button, HasTraits
from traits.trait_base import not_none

class Parent(HasTraits):
    say_hello = Button(method='do_hello')
    
    say_goodbye = Button(method='do_goodbye')
    
    def do_hello(self, *args, **kwargs):
        print("Hello")
        
    def do_goodbye(self, *args, **kwargs):
        print("Goodbye")
        
    def __init__(self, **traits):
        super(Parent, self).__init__(**traits)
        for name, trait in self.traits(method=not_none).items():
            method_name = getattr(trait, 'method')
            self._connect_listener(name, method_name)

    def _connect_listener(self, trait_name, method_name):
        def handler():
            method = getattr(self, method_name)
            return method()
        self.on_trait_change(handler, trait_name)
```
But this is a bit magical and less clear than the boilerplate methods.

The other option that may fit your use-case is that if you have lots of buttons it may be that what you want is a toolbar-style interface for the buttons, in which case you can follow the discussion on toolbars in the traitsui docs.

PR #418, opened 2018-02-07
Start fixing some wxPython-phoenix issues.

prabhuramachandran:
These are mostly backwards compatible changes so should not break
with previous versions of wxPython (i.e. 3.x at least).  There are many
more changes needed but this is a small first cut.


corranwebster:
The appveyor failures look unrelated.  Other than that, LGTM.

prabhuramachandran:
Thanks, there are some parts similar to #405.

PR #440, opened 2018-03-28
Bug fix

mm89039:
Numpy.numarray no longer exists


corranwebster:
Thanks for the bugfix!  Looks good to me, so I will merge once CI finishes (although TravisCI OS X builds look like they are having some issues this morning).

PR #455, opened 2018-04-21
Add popup completions for EnumEditor on Qt

### LINKED ISSUE #454: Allow popup display of completions in EnumEditor
Via @chenqi0805 and @ERAnderson - the ability to have a popup showing potential completions in an EnumEditor as the user types is useful when there are potentially a large number of completions:

image

This can be achieved in Qt with a handler which sets the completer mode on the control; but it would be nice to have this exposed as an option in the `Editor`.

corranwebster:
Includes code clean-up, tests and docs.

See #454 but only for Qt.

Equivalent on Wx would need a custom widget.


corranwebster:
Tests failing on Qt, passing on pyside...

PR #460, opened 2018-06-05
Add demo for tabular editor with context menu

jenniportman:
Since I spent some time figuring this out, I thought I'd add a demo. This adapts the demo for the `TableEditor` with context menu to work with a `TabularEditor`

A lot of the code is duplicated between the two demos because I assume the intent is that each demo is standalone. If that is not the case, I can certainly import some of the classes from the table editor demo.

### DIFF
--- a/examples/demo/Advanced/Tabular_editor_with_context_menu_demo.py
+++ b/examples/demo/Advanced/Tabular_editor_with_context_menu_demo.py
@@ -1,0 +1,165 @@
+"""
+Defining column-specific context menu in a Tabular Editor. Shows how the
+example for the Table Editor can be adapted to work with a Tabular Editor.
+
+The demo is a simple baseball scoring system, which lists each player and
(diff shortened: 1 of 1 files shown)


examples/demo/Advanced/Tabular_editor_with_context_menu_demo.py
```diff
@@ -0,0 +1,165 @@
...
+class PlayerAdapter(TabularAdapter):
+
+    alignment = 'center'
+
+    width = 0.09
+
+    def get_menu(self, object, trait, row, column):
+        column_name = self.column_map[column]
+        if column_name not in ['name', 'average']:
+            menu = Menu(Action(name='Add',
```
corranwebster: From a style point of view, it may be better to have the menu be a module global or stored in a class trait, since there is no local context being used in the menu definitions.

PR #469, opened 2018-07-15
[WIP] IPyWidgets Backend for TraitsUI

corranwebster:
Highly experimental, to the point that it possible should live in its own project, but it can live here for now.  To get it to work currently:

- create a new EDM environment
- install the pre-release ipywidgets with `pip install ipywidgets --pre`
- install traitsui and dependencies with `edm install traitsui`
- remove the EDM installed traitsui with `pip uninstall traitsui`
- install this branch with `python setup.py install`

If you are working on this you will need to `python setup.py install` after each change you make.  This sucks, but currently no better option.

Things that still need to be done:

- [ ] build out a bunch of editors
- [ ] toolkit specific color and font traits
- [ ] support for embedding in ipywidgets UIs (_may_ work already, but untested)
- [ ] check various ways of configuring views
- [ ] clean up the `ui_panel.py` code

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -148,6 +148,10 @@ if __name__ == "__main__":
                 'wx = traitsui.wx:toolkit',
                 'qt = traitsui.qt4:toolkit',
                 'null = traitsui.null:toolkit',
+                'ipywidgets = traitsui.ipywidgets:toolkit',
+            ],
+            'pyface.toolkits': [
+                'ipywidgets = traitsui.ipywidgets:toolkit',
             ],
         },
         platforms=["Windows", "Linux", "Mac OS-X", "Unix", "Solaris"],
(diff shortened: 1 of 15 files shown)


corranwebster:
Looks like test failure in last commit was because Github was inaccessible - re-started the failed test to be sure.

prabhuramachandran:
@corranwebster -- thanks.  there are a couple of things that you might want to refactor:
- In the last commit there is a `_repr_html_` method that I inject to make it easy to display a UI, this really should go somewhere else, I was not sure where though.
- The button editor doesn't work because pyface doesn't implement
resource_manager:PyfaceResourceFactory for the ipywidgets backend. I see that you already have this in the list of issues above but thought I'd just mention it here.

I'll try to push a few more editors as time permits.

corranwebster:
It looks like IPyWidgets buttons are limited to showing icons from the Font Awesome icon set only, so we probably won't be able to integrate with that.

prabhuramachandran:
Yes, it might be a good idea to take some notes on what we would need better support for and send that to the ipywidgets devs.

Issue #472, opened 2018-07-24
Remove "Specifies" when rendering a tooltip


prabhuramachandran:
Traitsui automatically adds "Specifies" to the tooltip when a `desc` is added to a trait definition.  This is not configurable and can be painful sometimes. Could this be removed or at least made configurable?

rkern:
Safest and easiest would be to use another metadata keyword that, ahem, specifies the complete string that you want to use and soft-deprecate `desc`. When the new keyword is not `None`, then it will be used, and `desc` will be ignored; otherwise, `'Specifies ' + desc` will be used.

prabhuramachandran:
Sounds good to me, what should this be?  `tooltip`?

rkern:
Yes, I believe that would be appropriate.

prabhuramachandran:
Sent a PR for this in #473.

prabhuramachandran:
Closing as #473 is merged.

PR #473, opened 2018-07-25
Use tooltip metadata optionally instead of desc.

prabhuramachandran:
This implements the suggestion in #472.  If a `tooltip` metadata
attribute exists for a trait, it is used as is for the tooltip,
otherwise the original behavior of using the `'Specifies' + item.desc`
is used.

### DIFF
--- a/traitsui/qt4/ui_panel.py
+++ b/traitsui/qt4/ui_panel.py
@@ -1101,8 +1101,9 @@ class _GroupPanel(object):
         label text.
 
         We also set the help on the QLabel control (from item.help) and the
-        tooltip (it item.desc exists; we add "Specifies " at the start of the
-        item.desc string, if item.tooltip exists we just use it as is).
+        tooltip (if the ``tooltip`` metadata on the edited trait exists, then
+        it will be used as-is; otherwise, if the ``desc`` metadata exists, the
+        string "Specifies " will be prepended to the start of ``desc``).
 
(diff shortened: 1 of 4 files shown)


traitsui/qt4/ui_panel.py
```diff
@@ -1098,9 +1100,9 @@ def _create_label(self, item, ui, desc, suffix=':'):
         we append a suffix (by default a colon ':') at the end of the
         label text.
 
-        We also set the help on the QLabel control (from item.help) and
-        the tooltip (it item.desc exists; we add "Specifies " at the start
-        of the item.desc string).
+        We also set the help on the QLabel control (from item.help) and the
+        tooltip (it item.desc exists; we add "Specifies " at the start of the
+        item.desc string, if item.tooltip exists we just use it as is).
```
rkern: I would talk about `tooltip` first rather than second, in order to mirror the code's logic. There are also a couple of errors in the original text that should be fixed. This is how I would write the parenthetical:

```
(if the ``tooltip`` metadata on the edited trait exists, then it will be used as-is;
otherwise, if the ``desc`` metadata exists, the string "Specifies " will be
prepended to the start of ``desc``)
```

prabhuramachandran: Thanks! Done.

corranwebster:
LGTM

PR #479, opened 2018-10-06
Removing duplicate 'wx.' in reference to CB_READONLY

ransonr:
Removing duplicate 'wx.' in reference to CB_READONLY


corranwebster:
Thanks!

PR #482, opened 2018-11-09
BUG: Fix a Python 2-ism in the TreeEditor code.

rkern:
Adds `six` to the list of dependencies.

### DIFF
--- a/traitsui/qt4/tree_editor.py
+++ b/traitsui/qt4/tree_editor.py
@@ -17,7 +17,7 @@
 #  Imports:
 #-------------------------------------------------------------------------
 
-from __future__ import absolute_imports
+from __future__ import absolute_import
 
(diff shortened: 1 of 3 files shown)


traitsui/qt4/tree_editor.py
```diff
@@ -17,10 +17,14 @@
 #  Imports:
 #-------------------------------------------------------------------------
 
+from __future__ import absolute_imports
```
corranwebster:
```suggestion
from __future__ import absolute_import
```

rkern:
I _am_ super-keen on using `six` where possible. As long as it's not confusing `2to3` (which it doesn't seem to be), it's one less thing we'll have to do to migrate to a single-source, Python 3-native repo, which we will need to do.

PR #493, opened 2019-01-17
Change getattr to xgetattr in tree get children

### LINKED ISSUE #491: Tree node get children method should handle nested trait names
The [`get_children`](https://github.com/enthought/traitsui/blob/193898594c85ad612b8b7ab9815bc810ace466f8/traitsui/tree_node.py#L219-L222) method uses the standard `getattr` rather than `traits.trait_base.xgetattr`, and so nested trait names cannot be used as children.

notmatthancock:
Closes #491

### DIFF
--- a/traitsui/tree_node.py
+++ b/traitsui/tree_node.py
@@ -220,7 +220,7 @@ class TreeNode(HasPrivateTraits):
     def get_children(self, object):
         """ Gets the object's children.
         """
-        return xgetattr(object, self.children)
+        return xgetattr(object, self.children, default=None)
 
     #-------------------------------------------------------------------------
     #  Gets the object's children identifier:
--- a/traitsui/tree_node.py
+++ b/traitsui/tree_node.py
@@ -459,13 +459,13 @@ class TreeNode(HasPrivateTraits):
     def get_background(self, object):
         background = self.background
         if isinstance(background, six.string_types):
-            background = getattr(object, background, background)
+            background = xgetattr(object, background, default=background)
         return background
 
     def get_foreground(self, object):
         foreground = self.foreground
         if isinstance(foreground, six.string_types):
-            foreground = getattr(object, foreground, foreground)
+            foreground = xgetattr(object, foreground, default=foreground)
         return foreground
 
     #-------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


traitsui/tree_node.py
```diff
@@ -246,7 +247,8 @@ def append_child(self, object, child):
     def insert_child(self, object, index, child):
         """ Inserts a child into the object's children.
         """
-        getattr(object, self.children)[index: index] = [child]
+        children = self.get_children()
+        children[index: index] = [child]
```
rkern:
```suggestion
        children[index:index] = [child]
```

notmatthancock:
Updated the comment. Note sure why travis is failing

corranwebster:
Travis failure look to be related to PySide2, which is new to the test suite (as in its been there about 3 months), and it seems to have started dying before any tests actually run.  The issue is orthogonal to this PR.

corranwebster:
This is a semantic change, as the behaviour of classes that override `get_children` but not the other methods will change.  However, I think that any class that overrides `get_children` would have had to override the other methods as well, so I think that this shouldn't break anyone's code.

traitsui/tree_node.py
```diff
@@ -219,7 +220,7 @@ def has_children(self, object):
     def get_children(self, object):
         """ Gets the object's children.
         """
-        return getattr(object, self.children)
+        return xgetattr(object, self.children)
```
corranwebster:
```suggestion
        return xgetattr(object, self.children, None)
```
The default return value of `xgetattr` is `Undefined` (see https://github.com/enthought/traits/blob/master/traits/trait_base.py#L345).  For backwards compatibility you need the default to be `None`.

corranwebster:
If you are going to do this in these three cases, you should probably do elsewhere in the file as well, eg. https://github.com/enthought/traitsui/blob/ffeb70b99979da852292ad9b785b181e833e0eb1/traitsui/tree_node.py#L462 and https://github.com/enthought/traitsui/blob/ffeb70b99979da852292ad9b785b181e833e0eb1/traitsui/tree_node.py#L1737 and several other places.

notmatthancock:
Suggestions incorporated. Thanks @corranwebster

PR #499, opened 2019-02-18
Add a StyledTreeEditor for Qt

jwiggins:
This overrides some of the machinery in `TreeEditor` for Qt so that users can control the rendering of items with a `QStyledItemDelegate`.


corranwebster:
Is it possible to just fold these extra capabilities into the Qt TreeEditor?  I'd be fine with having a trait on the `TreeEditor` factory which holds a renderer that is only used by Qt and ignored by Wx.

traitsui/qt4/extra/styled_tree_editor.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+class StyledTreeNode(TreeNode):
+    """ A TreeNode to use with StyledTreeEditor
+    """
+    #: Set to True to use ItemDelegate for drawing
+    delegates_style = Bool(False)
+
+
+if __name__ == '__main__':
```
corranwebster: Could you put this into the `examples` directories (not sure where is best, since it is Qt-specific.

jwiggins:
I felt like it was too invasive to put this in the normal Qt TreeEditor.

traitsui/qt4/extra/styled_tree_editor.py
jwiggins: That shouldn't be a problem

jwiggins:
CI is clean. (thanks @rahulporuri!)

corranwebster:
I'm working on a refactor of this to integrate it into the main editor and generally make it more flexible (eg. be able to override renderers for additional columns if there are any).

jwiggins:
Sounds good

PR #500, opened 2019-02-20
For now, allow CI failures on pyside2 jobs

rahulporuri:
and create separate jobs for pyside2 on 2.7, 3.5 and 3.6, which can
be allowed to fail.

this is hopefully a temporary measure, reverted when a fix for the
pyside2 issues causing CI failure is discovered.


corranwebster:
👍

PR #502, opened 2019-02-21
Tree node renderer for Qt

corranwebster:
This allows complete overriding of the renderers for `TreeEditor`s in Qt.  It is architected in such a way that we can potentially support a Wx version should we re-write the WxPython `TreeEditor` using `DataView`s.

This is based on #499 and this PR should replace that one.

screen shot 2019-02-21 at 2 21 25 pm


examples/demo/Extras/Tree_editor_with_TreeNodeRenderer.py
```diff
@@ -0,0 +1,113 @@
...
+
+    data = Array
+
+    def _data_default(self):
+        return np.random.standard_normal((1000,)).cumsum()
+
+
+class MyData(HasTraits):
+    name = Unicode('Rooty McRootface')
+    items = List(Instance(MyDataItem))
```
itziakos: (suggestion) replace `items` with `elements`. To avoid naming confusion with `_items` postfix on traits notification expressions.

corranwebster:
Looks like I have uncovered a bug with the `TreeEditor` under Python 3 - the `word_wrap` option would have failed but that codepath wasn't ever run; by making the editor always use a delegate we now get the error.

I suspect #499 will fail under Python 3 as well (tests pass, but example fails).

traitsui/tree_node_renderer.py
```diff
@@ -0,0 +1,52 @@
...
+        ----------
+        node : ITreeNode instance
+            The tree node to render.
+        column : int
+            The column in the tree that should be rendererd.
+        object : object
+            The underlying object being edited.
+        paint_context : object
+            A toolkit-dependent context for performing paint operations.
+        """
```
jwiggins: You should probably mention the optional return value here

corranwebster: Yep, although I think I have to get rid of it and replace it with an explicit `size_hint` method to fix the Python 3 issue cleanly.

traitsui/qt4/tree_editor.py
```diff
@@ -1933,3 +1914,93 @@ def _get_action(self, event):
...
+
+    def paint(self, painter, option, index):
+        """ Render the contents of the item. """
+        item = self.editor._tree.itemFromIndex(index)
+        expanded, node, instance = self.editor._get_node_data(item)
+        column = index.column()
+
+        renderer = node.get_renderer(object, column=column)
+        if renderer is None and self.editor.factory.word_wrap:
+            renderer = WordWrapRenderer()
```
jwiggins: I don't think you want to do this here. You're in a pretty hot code path (I think)

traitsui/qt4/helper.py
```diff
@@ -220,3 +220,59 @@ def __init__(self, icon, slot):
 #-------------------------------------------------------------------------
 
 DockStyle = Enum('horizontal', 'vertical', 'tab', 'fixed')
+
+#-------------------------------------------------------------------------
+#  Text Rendering helpers
+#-------------------------------------------------------------------------
+
+def wrap_text_with_eilision(text, font, width, height):
```
jwiggins: Typo: elision

traitsui/tree_node_renderer.py
```diff
@@ -0,0 +1,79 @@
...
+            A toolkit-dependent context for performing sizing operations.
+
+        Returns
+        -------
+        size : tuple of (width, height) or None
+        """
+        raise NotImplementedError()
+
+    def get_label(self, node, object, column=0):
+        """ Get the label associated with an item and column. """
```
jwiggins: Why is this here and not somewhere else?

traitsui/qt4/tree_editor.py
```diff
@@ -1933,3 +1911,74 @@ def _get_action(self, event):
...
+                super(TreeItemDelegate, self).paint(painter, option, index)
+            paint_context = (painter, option, index)
+            size = renderer.paint(
+                self.editor, node, column, instance, paint_context
+            )
+            if size is not None:
+                do_later(self.sizeHintChanged.emit, index)
+
+    def updateSizeHint(self, index, size):
+        """ Updates the size stored in the size_map. """
```
jwiggins: Is this method still needed?

traitsui/tree_node_renderer.py
corranwebster: This is a convenience method that papers over the gorpy way that the tree node classes handle multiple columns.  I'd love to fix tree node, but not in this PR (and I'd need to be really careful about code breakage).

traitsui/qt4/helper.py
corranwebster: Yep, already caught it.

traitsui/qt4/tree_editor.py
corranwebster: No!

traitsui/tree_node_renderer.py
jwiggins: Fair enough. It seemed slightly out of place.

traitsui/qt4/tests/test_helper.py
```diff
@@ -0,0 +1,118 @@
...
+        average_char_width = font_metrics.averageCharWidth()
+        line_spacing = font_metrics.lineSpacing()
+
+        width = 20 * average_char_width
+        height = 20 * line_spacing
+
+        lines = wrap_text_with_elision(lorem_ipsum, font, width, height)
+
+        expected_lines = get_expected_lines(lorem_ipsum, 20)[:20]
+        expected_lines[-1] = u'fugiat nulla pariat\u2026'
```
jwiggins: Not necessary with `from __future__ import unicode_literals`

corranwebster: Belt _and_ suspenders!

corranwebster:
Found another bug in the `TreeEditor` but I am not going to fix it as part of this PR (I think it existed before this PR).  See #503.

corranwebster:
All comments addressed, and tests now running cleanly.  I'll self merge this in a little while.

PR #504, opened 2019-02-25
Get Pyside2 tests working again

corranwebster:
The basic approach is to pin the pyside2 version for now.


corranwebster:
Yep, will do.

PR #510, opened 2019-03-01
Fixes for Undo/Redo

corranwebster:
Improve undo/redo exception handling.  Don't try to undo readonly items.  Short-circuit some expensive string comparisons.

### DIFF
--- a/traitsui/undo.py
+++ b/traitsui/undo.py
@@ -125,8 +125,9 @@ class UndoItem(AbstractUndoItem):
         """
         try:
             setattr(self.object, self.name, self.old_value)
-        except:
-            pass
+        except Exception:
+            from traitsui.api import raise_to_debug
+            raise_to_debug()
 
     #-------------------------------------------------------------------------
     #  Re-does the change:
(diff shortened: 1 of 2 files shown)


traitsui/undo.py
```diff
@@ -197,7 +201,7 @@ def merge_undo(self, undo_item):
                                     return False
                                 self.new_value = v2
                                 return True
-                        except:
+                        except Exception:
                             pass
```
rkern: `raise_to_debug()` like the others? Or is this one legitimate?

corranwebster: This one's legitimate based on the comment earlier in the function: https://github.com/enthought/traitsui/pull/510/files#diff-fbedc1554be3859074ab2232abb6ca73R185

I'd be happier if I knew what particular exception `wxColour` raises, but mainly I just care that we're not doing a bare `except` here.

rkern: Weird: `isinstance(wx.Colour(), collections.Sequence) == False` for me, but `len(wx.Colour()) == 3`.

PR #515, opened 2019-03-11
BUG: Use an existing omnipresent trait for the spring dummy.

### LINKED ISSUE #514: Spring items break pickling with Traits 5.0.0
I encountered a strange behaviour while pickling traits with Traits 5.0.0 and Traitsui 6.0.0

If one trait contains an instance of another trait, whose view contains a Spring item, pickling results in an error.

In the following code, if one comments out the spring item in the view of the MyTrait the code works correctly, otherwise results in an error.

Most likely related to: [https://github.com/enthought/traits/issues/447](https://github.com/enthought/traits/issues/447)

```


import pickle

from traits.api import HasStrictTraits, Str, Instance
from traitsui.api import View, Group, Item, InstanceEditor, spring

class MyTrait(HasStrictTraits):
    MyString = Str
    
    traits_view = View(Group(Item('MyString'), spring))   # If spring is removed, this code runs OK


class MySecondTrait(HasStrictTraits):
    MyFirst = Instance(MyTrait)
    
    
    def save(self):
        with open('test.dump', 'wb') as fid:
            st = self.__getstate__()
            pickle.dump(st, fid, protocol=2)
        
    def load(self):
        with open('test.dump', 'rb') as fid:
            self.__setstate__(pickle.load(fid))
        
    
    traits_view = View(Group(Item('MyFirst', editor=InstanceEditor(editable=True), style='custom'), show_labels=False), buttons=['OK'])
    
    
m0 = MyTrait(MyString='test')
m1 = MySecondTrait(MyFirst=m0)

m1.configure_traits()
m1.save()


m2 = MySecondTrait()
m2.load()
print(m2.MyFirst.MyString)

```

Expected output:

```
test
```

Error 
```

  File "D:\Temp\testpickle.py", line 46, in <module>
    m1.save()
  File "D:\Temp\testpickle.py", line 31, in save
    pickle.dump(st, fid, protocol=2)
_pickle.PicklingError: Can't pickle <class 'traits.trait_types.Disallow'>: it's not the same object as traits.trait_types.Disallow
```

rkern:
Fixes #514

Regardless of any underlying issues that we might want to also fix in Traits itself, this is likely a good idea.

### DIFF
--- a/traitsui/tests/test_ui.py
+++ b/traitsui/tests/test_ui.py
@@ -20,10 +20,10 @@ Test cases for the UI object.
 from __future__ import absolute_import
 import nose
 
-from traits.has_traits import HasTraits
+from traits.has_traits import HasTraits, HasStrictTraits
 from traits.trait_types import Str, Int
 import traitsui
-from traitsui.item import Item
+from traitsui.item import Item, spring
 from traitsui.view import View
 
 from traitsui.tests._tools import *
@@ -39,6 +39,17 @@ class FooDialog(HasTraits):
         Item('my_int'),
         Item('my_str'),
         buttons=['OK']
+    )
+
+
+class DisallowNewTraits(HasStrictTraits):
+    """ Make sure no extra traits are added.
+    """
+    x = Int(10)
+
+    traits_view = View(
+        Item('x'),
+        spring,
     )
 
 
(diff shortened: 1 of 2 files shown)


traitsui/tests/test_ui.py
```diff
@@ -191,3 +202,11 @@ def test_destroy_after_ok_qt():
 
     nose.tools.assert_is_none(ui.control)
     nose.tools.assert_equal(control.deleteLater._n_calls, 1)
+
+
+def test_no_spring_trait():
```
corranwebster: Needs a `@skip_if_null` decorator, I think.

corranwebster:
LGTM - test failure is unrelated.

PR #516, opened 2019-03-12
Fixed crash with Python3 (unicode vs bytes)

FedeMiorelli:
Drag-n-drop on table rows leads to a crash with Python 3 and PyQt5.

Fixed based on same changes applied in https://github.com/enthought/traitsui/commit/3aedd8ce69a04c81ce82d3acd06158f693003906


corranwebster:
Looks good to me.  Thanks for the contribution!

PR #521, opened 2019-03-22
Fix crash when creating view of TabularEditor with no columns

### LINKED ISSUE #520: Creating a view of TabularEditor causes segfault when columns not defined
In the FORCE project (https://github.com/force-h2020/force-wfmanager), we're using TabularEditor to display thousands of rows with columns set by the results of a user-defined workflow, which are not known before the results pane is created. PR #521  aims to fix this issue.

The creation of the results pane currently segfaults (macOS Mojave, qt4, current HEAD of traitsui master) when qt4 tries to get the font of a non-existent column: 

```
  File "/Users/mevans/src/traitsui/traitsui/tabular_adapter.py", line 376, in get_font
    return self._result_for('get_font', object, trait, row, column)
  File "/Users/mevans/src/traitsui/traitsui/tabular_adapter.py", line 701, in _result_for
    self.column_id = column_id = self.column_map[column]
IndexError: list index out of range
```
Below is an MWE that replicates this behaviour.

```
from traits.api import HasTraits, Str, List, Instance, Tuple, Property, on_trait_change
from traitsui.api import View, UItem, TabularEditor, ListColumn
from traitsui.tabular_adapter import TabularAdapter
from pyface.timer.do_later import DoLaterTimer

class DataModel(HasTraits):
    fields = List(Str)
    steps = List(Tuple)

class Pane(HasTraits):
    model = Instance(DataModel)
    adapter = Instance(TabularAdapter, ())
    rows = Property(List(Tuple), depends_on='model.steps')
    columns = Property(List(Tuple), depends_on='model.fields')

    @on_trait_change('model.fields')
    def _update_adapter(self):
        self.adapter.columns = self.model.fields

    def _get_rows(self):
        return self.model.steps

    def _get_columns(self):
        return [ListColumn(label=name, index=index)
                for index, name
                in enumerate(self.model.fields)]

    def traits_view(self):
        return View(UItem('rows',
                          editor=TabularEditor(adapter=self.adapter)),
                    resizable=True)

def set_columns(model):
    model.fields = ['foo', 'bar']

def set_data(model):
    model.steps = [(1, 2), (3, 4)]

if __name__ == '__main__':
    model = DataModel()
    pane = Pane(model=model)
    DoLaterTimer(200, set_columns, (model,), {})
    DoLaterTimer(400, set_data, (model,), {})
    pane.configure_traits()
```

ml-evs:
This PR aims to fix #520 by returning None any time a `View` tries to get the properties of a non-existent column in `TabularAdapter`.

I can't spot any unintended consequences of this, but as it seems that `get_font` is the main culprit here, this change could be applied at that level instead if that makes more sense.

### DIFF
--- a/traitsui/qt4/tabular_editor.py
+++ b/traitsui/qt4/tabular_editor.py
@@ -634,7 +634,12 @@ class _TableView(QtGui.QTableView):
             # we make a reasonable guess based on the minimum size hint and the font
             # of the first row.
             size = vheader.minimumSectionSize()
-            font = editor.adapter.get_font(editor.object, editor.name, 0)
+
+            # Check if any columns have been added, and use their font, otherwise
+            # use the default font
+            font = None
+            if 0 in editor.adapter.column_map:
+                font = editor.adapter.get_font(editor.object, editor.name, 0)
             if font is not None:
                 size = max(
                     size, QtGui.QFontMetrics(
--- a/traitsui/tabular_adapter.py
+++ b/traitsui/tabular_adapter.py
@@ -372,7 +372,9 @@ class TabularAdapter(HasPrivateTraits):
         font object should be returned. Note that all columns for the specified
         table row will use the font value returned.
         """
-        return self._result_for('get_font', object, trait, row, column)
+        if column in self.column_map:
+            return self._result_for('get_font', object, trait, row, column)
+        return None
 
     def get_text_color(self, object, trait, row, column=0):
         """ Returns the text color to use for a specified row or cell.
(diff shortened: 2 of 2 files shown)


corranwebster:
Looks like we have legitimate test failures here, so we'll need to think things through a bit further.

rkern:
Assuming the Travis failure is intermittent, this is fine as far as it goes, but it probably would require a comment in the source why that method and not the others has that check. Since the answer to that question is toolkit-specific, I think a better place for it might be in the `qt4` `TabularModel.data()` method rather than the adapter. I would be curious to know what `column`'s data is being requested when there are no columns displayed.

rkern:
Ah: https://github.com/enthought/traitsui/blob/master/traitsui/qt4/tabular_editor.py#L637

That would be the place to fix this. I'm not seeing a segfault, just a normal Python traceback.

ml-evs:
Fair point, I've fixed it at that level instead. 

The segfault was intermittent and was probably due to some exception handling specific in FORCE, the MWE from my issue only raises an IndexError.

rkern:
LGTM, pending CI!

corranwebster:
LGTM

PR #524, opened 2019-04-10
Introduce DateRangeEditor (for Qt only)

kitchoi:
This pull request introduces a `DateRangeEditor` and this has a side-effect of implementing the `multi_select=True` functionality for the custom `DateEditor` in Qt.

### DIFF
--- a/examples/demo/Advanced/Date_range_editor_demo.py
+++ b/examples/demo/Advanced/Date_range_editor_demo.py
@@ -16,9 +16,7 @@ class DateRangeEditorDemo(HasTraits):
 
     view = View(
                 Group(Item('date_range',
-                           editor=DateRangeEditor(
-                               allow_no_range=True,
-                            ),
+                           editor=DateRangeEditor(),
                            style='custom',
                            label='Date range'),
                       label='Date range'),
--- a/examples/demo/Advanced/Date_range_editor_demo.py
+++ b/examples/demo/Advanced/Date_range_editor_demo.py
@@ -2,7 +2,7 @@
 #  License: BSD Style.
 
 """
-A Traits UI editor that wraps a WX calendar panel.
+A Traits UI editor that wraps a Qt calendar panel.
 """
 from __future__ import absolute_import, print_function
 
--- a/examples/demo/Advanced/Date_range_editor_demo.py
+++ b/examples/demo/Advanced/Date_range_editor_demo.py
@@ -1,0 +1,38 @@
+#  Copyright (c) 2007-2009, Enthought, Inc.
+#  License: BSD Style.
+
+"""
+A Traits UI editor that wraps a WX calendar panel.
+"""
+from __future__ import absolute_import, print_function
+
+from traits.api import HasTraits, Date, Tuple
+from traitsui.api import View, Item, DateRangeEditor, Group
+
+
+class DateRangeEditorDemo(HasTraits):
+    """ Demo class to show DateRangeEditor. """
+    date_range = Tuple(Date, Date)
+
+    view = View(
+                Group(Item('date_range',
+                           editor=DateRangeEditor(
+                               allow_no_range=True,
+                            ),
+                           style='custom',
+                           label='Date range'),
+                      label='Date range'),
+                resizable=True)
+
+    def _date_range_changed(self):
+        print(self.date_range)
+
+
+#-- Set Up The Demo ------------------------------------------------------
+
+demo = DateRangeEditorDemo()
+
+if __name__ == "__main__":
+    demo.configure_traits()
+
+#-- eof -----------------------------------------------------------------------
--- a/traitsui/editors/date_range_editor.py
+++ b/traitsui/editors/date_range_editor.py
@@ -6,6 +6,12 @@ from .date_editor import DateEditor
 
 
 class ToolkitEditorFactory(DateEditor):
+    """ Editor for a date range. The target value should be a tuple
+    containing two dates (start date, end date)
+    """
+
+    # This must be set to true for setting a date range.
+    multi_select = Bool(True)
 
     # Whether it is possible to unset the date range.
     allow_no_range = Bool(False)
--- a/traitsui/editors/date_range_editor.py
+++ b/traitsui/editors/date_range_editor.py
@@ -1,7 +1,7 @@
 
 from __future__ import absolute_import
 
-from traits.api import Bool
+from traits.api import Bool, Constant
 from .date_editor import DateEditor
 
 
@@ -11,7 +11,7 @@ class ToolkitEditorFactory(DateEditor):
     """
 
     # This must be set to true for setting a date range.
-    multi_select = Bool(True)
+    multi_select = Constant(True)
 
     # Whether it is possible to unset the date range.
     # If true, then the date range will be set to (None, None)
--- a/traitsui/editors/date_range_editor.py
+++ b/traitsui/editors/date_range_editor.py
@@ -16,7 +16,7 @@ class ToolkitEditorFactory(DateEditor):
     # Whether it is possible to unset the date range.
     # If true, then the date range will be set to (None, None)
     # when all the dates are unselected.
-    allow_no_range = Bool(False)
+    allow_no_selection = Bool(False)
 
 
 DateRangeEditor = ToolkitEditorFactory
--- a/traitsui/editors/date_range_editor.py
+++ b/traitsui/editors/date_range_editor.py
@@ -1,0 +1,13 @@
+
+from __future__ import absolute_import
+
+from traits.api import Bool
+from .date_editor import DateEditor
+
+
+class ToolkitEditorFactory(DateEditor):
+
+    allow_no_range = Bool(False)
+
+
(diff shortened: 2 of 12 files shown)


corranwebster:
This looks good in principle.  For added developer ease, it might make sense to create a `namedtuple` for the date range so that you can say something like `date_range.start_date` and `date_range.end_date`.

kitchoi:
Ready for review.

corranwebster:
The appearance of the widget looks fine and the date highlighting is good, but I am having trouble with interactions (on OS X, may be different on other OS):

- there are interactions that I would expect to work which don't (for example shift-clicking to extend the current selection)
- it is unclear to me when clicking on a date will select one data and when it will select a range (as best I can tell it seems like when you have a range selected, one click clears the range and selects a new date, the second click doesn't select a range just another date, and then the third click creates the new range)

It's possible I am doing things wrong.

I think the second fixable by having the line https://github.com/enthought/traitsui/pull/524/files#diff-08bd5851eb1d7784d7cc73c3264601beR49 be `(value, value)` instead of `(None, None)` and basically never allow the `(None, None)` setting except via code-based updates.

Things like shift-clicking are harder to implement, and would probably require writing a custom `QCalendarWidget` subclass in Python, or attaching an event filter, to catch shift-clicks.

examples/demo/Advanced/Date_range_editor_demo.py
```diff
@@ -0,0 +1,38 @@
+#  Copyright (c) 2007-2009, Enthought, Inc.
+#  License: BSD Style.
+
+"""
+A Traits UI editor that wraps a WX calendar panel.
```
corranwebster: Not WX!

traitsui/editors/date_range_editor.py
```diff
@@ -0,0 +1,22 @@
...
+from .date_editor import DateEditor
+
+
+class ToolkitEditorFactory(DateEditor):
+    """ Editor for a date range. The target value should be a tuple
+    containing two dates (start date, end date)
+    """
+
+    # This must be set to true for setting a date range.
+    multi_select = Bool(True)
```
corranwebster: Perhaps use a `Constant(True)`?

corranwebster:
One more question - given that the current `DateEditor` allows multi-selection, and so works with either single values or lists of values, do we need a separate `DateRangeEditor` or could the functionality be folded in as an extra option on the `DateEditor` class?

kitchoi:
Thank you for reviewing!

> there are interactions that I would expect to work which don't (for example shift-clicking to extend the current selection)

Do you mean handling the situation when `shift_to_select` is set to `True`?  I am not entirely sure I fully understand the behaviour defined by this flag. Could this be handled as a separate issue, that `shift_to_select` is not handled in Qt custom `DateEditor`?

> it is unclear to me when clicking on a date will select one data and when it will select a range (as best I can tell it seems like when you have a range selected, one click clears the range and selects a new date, the second click doesn't select a range just another date, and then the third click creates the new range)

After having selected a range, one click clears the range and the date on which that click occurs is not selected: the font is normal, rather than bold.  Selected dates are identified by the bold font, different background and foreground color.  Perhaps the bold font isn't obvious enough.  Is there a way to undo the blue background (possibly OS dependent theme) when the user clicks a date?

> I think the second fixable by having the line https://github.com/enthought/traitsui/pull/524/files#diff-08bd5851eb1d7784d7cc73c3264601beR49 be (value, value) instead of (None, None) and basically never allow the (None, None) setting except via code-based updates.

The ability to remove the range, i.e. equivalent to selecting no dates, is needed for the project which motivated this PR.  So I would prefer keeping it if possible.

> given that the current DateEditor allows multi-selection, and so works with either single values or lists of values, do we need a separate DateRangeEditor or could the functionality be folded in as an extra option on the DateEditor class?

Selecting a date range is indeed a special case of selecting multiple date.
I would like to be able to define a date range with just two dates instead of a list of all the dates in the range: (1) it is easier to use the traits (2) it is more efficient.  It is surely possible to fold the logic into `DateEditor` with a target trait being `Tuple(Date, Date)` and define yet another flag on the factory.  But with the `DateEditor` already supporting either a `Date` or `List(Date)`, I find the `update_object` and `update_editor` are already complicated enough.  The support for `allow_no_range` contributes to that complexity as well.  Hence I find it much simpler to use and to implement if we extract this special case to a separate editor.

traitsui/qt4/date_range_editor.py
```diff
@@ -0,0 +1,81 @@
...
+        else:
+            start_date = value
+
+        self.value = (start_date, end_date)
+        self._apply_style_to_range(start_date, end_date)
+
+    def _apply_style_to_range(self, start_date, end_date):
+        num_days = (end_date - start_date).days + 1
+
+        selected_dates = [
```
kitchoi: Efficiency: I think I can use an iterator here instead of a list.

corranwebster:
> Do you mean handling the situation when shift_to_select is set to True? I am not entirely sure I fully understand the behaviour defined by this flag. Could this be handled as a separate issue, that shift_to_select is not handled in Qt custom DateEditor?

Yes, something like that - didn't realize there was already a flag, and if so, yes it could be a separate PR.

> After having selected a range, one click clears the range and the date on which that click occurs is not selected: the font is normal, rather than bold. Selected dates are identified by the bold font, different background and foreground color. Perhaps the bold font isn't obvious enough. Is there a way to undo the blue background (possibly OS dependent theme) when the user clicks a date?

Yep - I didn't notice the bolding at all.  If there were some way to show no selected date in the UI that would be good.  I have this fear there may be no easy way to remove the selection or change the colour, though.

> The ability to remove the range, i.e. equivalent to selecting no dates, is needed for the project which motivated this PR. So I would prefer keeping it if possible.

Maybe turn this flag off in the example, then, and keep it as a thing that needs to be specially turned on.

> Hence I find it much simpler to use and to implement if we extract this special case to a separate editor.

The update logic can be pushed out into methods depending on the case, and I imagine that `allow_no_range` (or perhaps better `allow_no_selection`) is a case that could be useful on both the other versions of the editor.  But in any case it's not a huge issue, so the current PR is OK.

kitchoi:
> Maybe turn this flag off in the example, then, and keep it as a thing that needs to be specially turned on.

No problem!

> The update logic can be pushed out into methods depending on the case, and I imagine that allow_no_range (or perhaps better allow_no_selection) is a case that could be useful on both the other versions of the editor. But in any case it's not a huge issue, so the current PR is OK.

`allow_no_selection` is a better name indeed.
In fact with this PR the custom `DateEditor(multi_select=True)` already allows no selection: you should be able to get an empty list when you click a selected date again.  See `examples/demo/Advanced/Date_editor_demo.py`.  The inability to remove the selection in the UI applies here too.

If the current API is accepted, would you agree that we should rename `traitsui.qt4.date_editor.CustomEditor` to `BaseCustomDateEditor` so that it is more explicit the class is meant to be subclassed?  There is another [existing `StyledDateEditor`](https://github.com/enthought/traitsui/blob/master/traitsui/qt4/styled_date_editor.py) which inherits `traitsui.qt4.date_editor.CustomEditor`.  Of course we will also need to add `CustomEditor = BaseCustomEditor` in the `date_editor` module [for the import mechanism to work](https://github.com/enthought/traitsui/blob/707d2277541fd9a49897012195e3a58cfd00cd8a/traitsui/editor_factory.py#L221-L234).  Also wonder if this should be a separate PR...

corranwebster:
Yes, I'd shift clean-up to a separate PR - I think there is probably a place for a PR which gives a subclassed `QCalendarWidget` with better selection behaviour and that would be the place to clean-up the class heirarchy.

traitsui/qt4/date_editor.py
```diff
@@ -154,14 +164,60 @@ def update_editor(self):
...
-            print('Invalid date:', year, month, day)
-            raise
+        value = datetime.date(q_date.year(), q_date.month(), q_date.day())
+        if self.factory.multi_select:
+            if value in self.value:
+                self.unselect_date(value)
+            else:
+                self.select_date(value)
+
+            self.value = sorted(self._selected)
```
kitchoi: Not sure if sorting the date is weird or not.  We could maintain the order of the dates to be in the same order as how the user selects them.  It is also somewhat inefficient that we keep sorting the list of date. @corranwebster What do you think?

corranwebster: Sorting is the right behaviour for the date rage (there is a definite start and end, and a 'backwards' selection is likely to be more confusing for users even if it is theoretically more pure), and sorting is _probably_ the right behaviour for the multi-select (although it is a bit less obvious if the data coming in from the traits side is unsorted - I can see an argument that it maybe should be optional but on by default).

Timsort is fairly efficient for almost-sorted lists, if I recall, so the sorting overhead is likely not that much (particularly when the number of selected dates is likely to be small).

kitchoi: Thanks!  I will keep the sort here then.

kitchoi:
Ready to be reviewed again.

PR #528, opened 2019-04-25
Ensure the tree model gets the expected node object on updates

### LINKED ISSUE #494: TreeNode doesn't handle extented label trait updates in Qt
If you use an extended trait name for the label of a TreeNode, you get an exception in the Qt backend if the label changes, as an unexpected object gets passed as affected node.  The fix is to use a listener which is a closure around the object in question, something like:
```
    def when_label_changed(self, object, listener, remove):
        """ Sets up or removes a listener for the label being changed on a
        specified object.
        """
        label = self.label
        if label[:1] != '=':
            def wrapped_listener(target, name, label):
                return listener(object, name, label)
            object.on_trait_change(wrapped_listener, label, remove=remove,
                                   dispatch='ui')
```

### LINKED ISSUE #529: Tree editor xgetattr incompatible with fast UI dispatch
When attempting to write a test for #493 I get the following:
```
TraitError: Trait notification dispatch type 'fast_ui' is not compatible with handler signature and extended trait name notification style.
```

As a result, we will probably need to revert #493 in whole or in part.

corranwebster:
This fixes #494 and fixes #529 including similar issues affected updating of children and header labels.

### DIFF
--- a/traitsui/tests/editors/test_tree_editor.py
+++ b/traitsui/tests/editors/test_tree_editor.py
@@ -16,16 +16,32 @@
 from __future__ import absolute_import
 import unittest
 
-from traits.api import Bool, HasTraits, Instance, Int, List
+from traits.api import Bool, HasTraits, Instance, List, Str
 from traitsui.api import Item, TreeEditor, TreeNode, View
 
-from traitsui.tests._tools import *
+from traitsui.tests._tools import (
+    press_ok_button, skip_if_null, skip_if_not_qt4,
+    store_exceptions_on_all_threads
+)
+
+
+class BogusWrap(HasTraits):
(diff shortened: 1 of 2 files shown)


corranwebster:
Nope - this doesn't work as we don't have the same listener object when it comes to removing the listener.  Fix may need to go into tree editor rather than tree node, in which case it may be simpler just to kill this PR.

corranwebster:
I've resolved by adding a listener cache to the `TreeNode` instance; I think that the key should be sufficiently unique that it never gets duplicated. It is unique for type of thing being updated, trait listened to, object listened to, and listener which is a bound method and so includes the editor instance, so in particular it should be fine if there are multiple editor instances sharing the same `TreeNode`.

PR #547, opened 2019-05-03
Scroll to column implementation.

avivekan:
This change allows scrolling to the column index (nearly identical to scroll_to_row). Corran, I couldnt find any tests for tabular_editor, but if I did miss it please let me know and I will add one asap for scroll_to_column. 

In the near future, I'll send you another PR to populate a selected_column property similar to selected_row, but this one would be useful for our project now.

### DIFF
--- a/traitsui/qt4/tabular_editor.py
+++ b/traitsui/qt4/tabular_editor.py
@@ -443,7 +443,10 @@ class TabularEditor(Editor):
         """
         scroll_hint = self.scroll_to_row_hint_map.get(
             self.factory.scroll_to_row_hint, self.control.PositionAtCenter)
-        self.control.scrollTo(self.model.index(0, column), scroll_hint)
+        self.control.scrollTo(
+            self.model.index(self.activated_row, column),
+            scroll_hint
+        )
 
     #-- Table Control Event Handlers -----------------------------------------
 
--- a/traitsui/qt4/tabular_editor.py
+++ b/traitsui/qt4/tabular_editor.py
@@ -101,6 +101,9 @@ class TabularEditor(Editor):
     # The event triggering scrolling.
     scroll_to_row = Event(Int)
 
+    # The event triggering scrolling.
+    scroll_to_column = Event(Int)
+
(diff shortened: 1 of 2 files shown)


traitsui/qt4/tabular_editor.py
```diff
@@ -433,6 +438,14 @@ def _scroll_to_row_changed(self, row):
             self.factory.scroll_to_row_hint, self.control.PositionAtCenter)
         self.control.scrollTo(self.model.index(row, 0), scroll_hint)
 
+    def _scroll_to_column_changed(self, column):
+        """ Scroll to the given column.
+        """
+        import ipdb; ipdb.set_trace()
```
rahulporuri: debugging statement.

avivekan: Thanks, missed that.

traitsui/qt4/tabular_editor.py
```diff
@@ -433,6 +438,14 @@ def _scroll_to_row_changed(self, row):
...
+        scroll_hint = self.scroll_to_row_hint_map.get(
+            self.factory.scroll_to_row_hint, self.control.PositionAtCenter)
+        self.control.scrollTo(self.model.index(0, column), scroll_hint)
```
corranwebster: I think this will always scroll back up to the top (the row index is 0).  You might use `activated_row` instead?

avivekan: Thanks Corran, made the change. I noticed that scroll_to_row has this issue, and that's what I am poking around to see if I can store a 'selected_column' property as well.

corranwebster:
Looks good!  Thank you for the contribution!

traitsui/qt4/tabular_editor.py
corranwebster: Don't worry about this in this PR - you'd want an `activated_row` property rather than a `selected_row` because selections can potentially span multiple rows (and/or columns).

If you want to handle this, then you can make another PR, but I don't think it's urgent.

Issue #556, opened 2019-05-10
TreeNodeObject fails to initialize _listener_cache


corranwebster:
See https://github.com/enthought/mayavi/issues/779

This line https://github.com/enthought/traitsui/blob/master/traitsui/tree_node.py#L1898 requires a `_listener_cache` which is at least an empty dictionary, but we don't initialize anywhere.  Some thought may be needed if `TreeNodeObject` classes are expected to be stateless.

This breaks Mayavi and potentially a number of other projects.

PR #557, opened 2019-05-10
Fix TreeNodeObject listener cache

corranwebster:
Fix for #556 and some other fixes to copy-pasted code.

### DIFF
--- a/traitsui/tests/editors/test_tree_editor.py
+++ b/traitsui/tests/editors/test_tree_editor.py
@@ -89,11 +89,6 @@ class BogusTreeNodeObject(TreeNodeObject):
 
     bogus_list = List
 
-    wrapped_bogus = Instance(BogusWrap)
-
-    def _wrapped_bogus_default(self):
-        return BogusTreeNodeObject(bogus_list=[BogusTreeNodeObject()])
-
 
 class BogusTreeNodeObjectView(HasTraits):
     """ A traitsui view visualizing Bogus objects as trees. """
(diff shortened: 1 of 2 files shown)


traitsui/tree_node.py
```diff
@@ -1889,11 +1892,11 @@ def tno_when_label_changed(self, node, listener, remove):
         """
         label = node.label
         if label[:1] != '=':
-            memo = ('label', label, object, listener)
```
jwiggins: So it was grabbing the `object` builtin?

corranwebster: Yep.  This is a dumb error on my part:

- I assumed `TreeNodeObject` subclassed `TreeNode`
- I then copy-pasted and didn't pay attention to the context fully.

jwiggins: I guess the icing on the cake was Mayavi's use of `TreeNodeObject` as the base of it's `Base` class?

corranwebster: Yep, whereas I almost never use `TreeNodeObject`.

I've added a regression test.

traitsui/tests/editors/test_tree_editor.py
```diff
@@ -110,6 +153,34 @@ def _test_tree_editor_releases_listeners(self, hide_root, nodes=None,
             notifiers_list = bogus.trait(trait)._notifiers(False)
             self.assertEqual(0, len(notifiers_list))
 
+    def _test_tree_node_object_releases_listeners(self, hide_root, nodes=None,
+                                                  trait='bogus_list',
+                                                  expected_listeners=1):
```
jwiggins: None of the calling code uses the `expected_listeners` arg. You should probably remove it.

traitsui/tests/editors/test_tree_editor.py
```diff
@@ -110,6 +153,34 @@ def _test_tree_editor_releases_listeners(self, hide_root, nodes=None,
...
+                bogus=bogus,
+                hide_root=hide_root,
+                nodes=nodes,
+            )
+            ui = tree_editor_view.edit_traits()
+
+            # The TreeEditor sets a listener on the bogus object's
+            # children list
+            notifiers_list = bogus.trait(trait)._notifiers(False)
+            self.assertEqual(expected_listeners, len(notifiers_list))
```
jwiggins: And just write `self.assertEqual(1, len(notifiers_list)) here

PR #564, opened 2019-06-05
Update unittest import

rahulporuri:
Import it from the standard library instead of importing it from `traits`.


corranwebster:
LGTM

rahulporuri:
thanks for the review @corranwebster

PR #565, opened 2019-06-06
Fix the sizeHint() default for TreeItemDelegate

jwiggins:
The default implementation of `QStyledItemDelegate::sizeHint()` is exactly what Qt calls when no item delegate is set on a `QTreeView`.


jwiggins:
Thanks for the quick review!

PR #574, opened 2019-06-25
Fix modal view application

### LINKED ISSUE #573: Problem with QT eventloop for Modal dialog
When using the Qt backend, creating a modal dialog with no other UI present, clicking the OK or Cancel buttons or closing the window will not terminate the event loop after exiting the dialog. 

This unexpected behavior does not happen when using the Wx backend.

Below is a minimal example displaying this issue:

```
from traits.api import HasTraits, Int, Instance
from traitsui.api import Item, ModelView, View


class TestModel(HasTraits):
    cell = Int


class TestModelView(ModelView):
    #: model
    model = Instance(TestModel)

    #: index of object
    index = Int()


dummyview = View(
    Item('model.cell'),
    title="Enter IDs and conditions",
    buttons=['OK', 'Cancel']
)


def base_ui(model, view):
    mv = TestModelView(model=model)
    result = mv.configure_traits(kind='modal', view=view)
    print(result)


if __name__ == '__main__':
    tm = TestModel()
    base_ui(model=tm, view=dummyview)
```

Clicking OK or cancel results in the UI hanging. Manually quitting the UI will allow the event loop to terminate and execution to continue.

The line in question causing problems is https://github.com/enthought/traitsui/blob/master/traitsui/qt4/view_application.py#L139

Its possible this line should not be executed when the dialog kind is `modal` or `livemodal`.

corranwebster:
The event loop was being run twice under Qt when using `configure_traits` in conjunction with a modal `kind`.

Also adds fairly comprehensive tests for `view_application` which covers most use cases under Qt.

Fixes #573.

### DIFF
--- a/traitsui/tests/test_view_application.py
+++ b/traitsui/tests/test_view_application.py
@@ -20,6 +20,10 @@ from ._tools import is_current_backend_qt4
 GuiTestAssistant = toolkit_object('util.gui_test_assistant:GuiTestAssistant')
 no_gui_test_assistant = (GuiTestAssistant.__name__ == 'Unimplemented')
 
+if no_gui_test_assistant:
+    class GuiTestAssistant(object):
+        pass
+
 
 class SimpleModel(HasTraits):
     cell = Int
--- a/traitsui/tests/test_view_application.py
+++ b/traitsui/tests/test_view_application.py
@@ -1,0 +1,227 @@
+#  Copyright (c) 2019, Enthought, Inc.
+#  All rights reserved.
+#
+#  This software is provided without warranty under the terms of the BSD
+#  license included in enthought/LICENSE.txt and may be redistributed only
+#  under the conditions described in the aforementioned license.  The license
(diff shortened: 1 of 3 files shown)


traitsui/tests/test_view_application.py
```diff
@@ -0,0 +1,231 @@
...
+    def stop_event_loop(self):
+        self.gui.stop_event_loop()
+        self.event_loop_timeout = True
+
+    def test_modal_view_application_close(self):
+        self.view_application('modal')
+
+        self.assertTrue(self.closed)
+        self.assertFalse(self.event_loop_timeout)
+        self.assertFalse(self.result)
```
mgrady3: Just a question on what this is testing, since I'm less familiar with the traits/traitsui internals, not a suggestion to change anything:

Is self.result just grabbing the return code or something similar of each test when the GUI exits?

corranwebster: Yes - it is essentially the expected OK/Cancel status.

PR #588, opened 2019-07-17
Fix/scroll to row preserve column

### LINKED ISSUE #581: TabularEditor: race condition for scroll_to_row and selected_row (and others...)
I found a race condition when interacting with the TabularEditor: when clicking a cell I would expect the TabularEditorEvent "clicked" to be fired first (from `_on_click(...)`), but there's a race condition where it sometimes calls _scroll_to_row_changed before calling _click_changed, and sometimes it wont call _click_changed at all. This leads to the table sometimes jumping to column 0, then back to the selected column sometimes when I click a cell (sometimes = every 2-3 clicks).

I can work around this by removing the `scroll_to_row='selection_trait'` in the code below, but I'd still like to move focus onto a selected row when needed. I've also tried making a new trait altogether for scroll_to_row (separate from selected_row), but when I set both I see the sporadic jumping behavior.

Example code:
```
UItem(
    editor=TabularEditor(
        adapter=self.some_adapter,
        show_row_titles=True,
        refresh='refresh_table_trait',
        selected_row=(
            'selection_trait'
        ),
        scroll_to_row=(
            'selection_trait'
        ),
    ),
),
```

I'm working on a fix for this in branch `enthought/traitsui::scroll-to-row-preserve-column`. Filing this ticket for tracking.

avivekan:
Fixes #581 

A race condition occurs when externally defining both scroll_to_row and selected_row, where it either selects or selects and scrolls. After talking to Corran, we decided that after selecting a row, we can scroll to it immediately after without an external call. I implemented both a scroll to row and a scroll to column change here, and keep track of the clicked cell's row/column index to ensure the scrolling happens correctly (instead of jumping to column 0 at each click).

### DIFF
--- a/traitsui/qt4/tabular_editor.py
+++ b/traitsui/qt4/tabular_editor.py
@@ -76,6 +76,10 @@ class TabularEditor(Editor):
     selected_row = Int(-1)
     multi_selected_rows = List(Int)
 
+    # The optional extended name of the trait to synchronize the selection
+    # column with:
+    selected_column = Int(-1)
+
     # The most recently actived item and its index:
     activated = Any(comparison_mode=NO_COMPARE)
     activated_row = Int(comparison_mode=NO_COMPARE)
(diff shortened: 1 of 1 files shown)


corranwebster:
LGTM

PR #590, opened 2019-07-19
Fix tree node copy failure to copy

### LINKED ISSUE #578: Copy/pasting a tree node doesn't copy
I don't have a minimal example but see this with Mayavi's tree editor.  Basically when we copy and paste a node, it doesn't copy the object it but uses the same node.  I think in the changes to the tree editor code this was missed out.  Basically the copy just keeps a reference to the node but when the node is inserted make_copy=False and so it doesn't deep copy the item.

corranwebster:
Make the copy action actually copy the underlying data. Fixes #578.


corranwebster:
Ah - no this isn't right as double pastes will share the same object.

Issue #594, opened 2019-07-25
Qt TableEditor missing features


corranwebster:
The Qt `TableEditor` is missing some features compared to the Wx version.  These include:

- [ ] Lines still show even when `show_lines` is `False`
- [ ] Foreground colors ignored.
- [ ] Background colors ignored.
- [ ] Label colors ignored.
- [ ] Font ignored.
- [ ] Search UI in toolbar.

corranwebster:
Some of these are easy to fix with a bit of additional logic in the Qt table model. Selection colours and line colours require overriding the painting of the cell, which will have performance impacts.

srk:
QT5 Search UI in toolbar now seems to work for me (TraitsUI 7.2.1, Pyface 7.4.0), but I can't get the TableEditor default toolbar customization and column selection icons to appear.

PR #597, opened 2019-07-26
Fix table editor styling

corranwebster:
Improves situation for #594 although it is not a complete fix.  This:

- fixes most color traits on the Qt TableEditor so that they work as advertised and match behaviour of WxPython backend as closely as feasible.
- fixes the font traits on the Qt TableEditor so that they actually do something.
- fixes the row_height on the Qt TableEditor so that if gives the right default size.

It also includes some drive-by documentation fixes.

### DIFF
--- a/traitsui/wx/table_editor.py
+++ b/traitsui/wx/table_editor.py
@@ -24,12 +24,12 @@ from operator import itemgetter
 
 import wx
 
-from pyface.ui.wx.grid.api import Grid
 from pyface.dock.api import (
     DockWindow, DockSizer, DockSection, DockRegion, DockControl
 )
 from pyface.image_resource import ImageResource
 from pyface.timer.api import do_later
+from pyface.ui.wx.grid.api import Grid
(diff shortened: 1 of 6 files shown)


traitsui/wx/table_editor.py
```diff
@@ -18,65 +18,44 @@
...
 
 import wx
 
-from traits.api \
-    import Int, List, Instance, Str, Any, Button, Tuple, \
-    HasPrivateTraits, Bool, Event, Property
-
-from traitsui.api \
-    import View, Item, UI, InstanceEditor, EnumEditor, Handler, SetEditor, \
+from pyface.ui.wx.grid.api import Grid
```
rahulporuri: nit : it looks like the rest of the imports are ordered alphabetically but this one is out of order! (_i'm out of order? no you're out of order!_)

corranwebster:
@rahulporuri Import ordering fixed.  Any other review comments?

PR #601, opened 2019-08-07
Make View icon trait an Image trait

### LINKED ISSUE #600: View's icon trait should be an Image, not Any
The `icon` trait of a `View` is currently defined as `Any` but using code expects the value to hold an `ImageResource`.  The trait should probably be modified to be an `Image` trait so that it accepts both image names and image resources.

corranwebster:
Does the same for the `UI` class.  Also does some clean-up of old, commented out code.

Fixes #600

### DIFF
--- a/traitsui/ui.py
+++ b/traitsui/ui.py
@@ -28,6 +28,7 @@ from __future__ import absolute_import
 import shelve
 import os
 
+from pyface.ui_traits import Image
 from traits.api import (
     Any,
     Bool,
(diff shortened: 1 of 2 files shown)


corranwebster:
@rahulporuri Could you review this?

rahulporuri:
FTR, an `Image` trait is
> 
    """ Defines a trait whose value must be a ImageResource or a string
        that can be converted to one.
    """

PR #624, opened 2019-08-15
Add 'docs' command to etstool that builds HTML documentation

corranwebster:
This adds a "docs" subcommand to etstool.py so that you should now be able to correctly build documentation from an existing environment with the command:
```
python etstool.py docs ...
```
with the usual `--runtime`, `--version` and `--environment` options available.  So for example:
```
python etstool.py install
python etstool.py docs
```
will build documentation for the default runtime and toolkit (currently 3.6 and PyQt).

HTML documentation will be in `docs/build/html` as usual.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -123,6 +123,11 @@ runtime_dependencies = {
     '2.7': {'mock'},
 }
 
+doc_dependencies = {
+    "sphinx",
+    "enthought_sphinx_theme",
+}
+
 environment_vars = {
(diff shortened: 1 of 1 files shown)


corranwebster:
@sandhya-sago if you have time, could you review this?

It will also help with the problem you were having about not getting documenation which looked like what we have up on the website.

corranwebster:
Test failure on Travis was unrelated.  I have restarted the build.

sandhya-sago:
Thanks, I have tried this and the html looks updated.

PR #628, opened 2019-08-15
Fix call to bind

### LINKED ISSUE #626: Wx backend FileEditor and DirectoryEditor exceptions on close
From #623:

It looks like the file and directory editor bugs are because of incorrect calls to `Bind`. The line here https://github.com/enthought/traitsui/blob/master/traitsui/wx/file_editor.py#L125 does not match the signature of `Bind`: https://wxpython.org/Phoenix/docs/html/wx.EvtHandler.html#wx.EvtHandler.Bind

The correct call should be something like:
```
panel.Bind(wx.EVT_BUTTON, None, id=self._button.GetId())
```

I suspect the `DirectoryEditor` is the same.

sandhya-sago:
Fixes #626.
Fixed the incorrect call, now file editor and directory editor demos are working.

### DIFF
--- a/traitsui/wx/file_editor.py
+++ b/traitsui/wx/file_editor.py
@@ -122,7 +122,7 @@ class SimpleEditor(SimpleTextEditor):
         """ Disposes of the contents of an editor.
         """
         panel = self.control
-        panel.Bind(wx.EVT_BUTTON, None, self._button.GetId())
+        panel.Bind(wx.EVT_BUTTON, None, id=self._button.GetId())
         self._button = None
 
         if self.history is not None:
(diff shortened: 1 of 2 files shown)


traitsui/wx/file_editor.py
```diff
@@ -122,7 +122,7 @@ def dispose(self):
         """ Disposes of the contents of an editor.
         """
         panel = self.control
-        panel.Bind(wx.EVT_BUTTON, self._button.GetId(), None)
+        panel.Bind(wx.EVT_BUTTON, None, self._button.GetId())
```
corranwebster: Since this is using the id of the button rather than the instance, you need to pass the id in via keyword argument.
```suggestion
        panel.Bind(wx.EVT_BUTTON, None, id=self._button.GetId())
```

corranwebster:
Test failures are unrelated.

PR #633, opened 2019-08-16
Fix multi-select in a TabularEditor

### LINKED ISSUE #605: traitsui 6.1.2 breaks TabularEditor multi_select
Demo code:

```
from traits.api import HasTraits, List, Int, Unicode, on_trait_change
from traitsui.api import View, UItem, TabularEditor
from traitsui.tabular_adapter import TabularAdapter


class Demo(HasTraits):
    table = List(List(Unicode), [['A', 'B'], ['C', 'D'], ['E', 'F']])
    selected_rows = List(Int)

    traits_view = View(
        UItem(
            'table',
            editor=TabularEditor(
                adapter=TabularAdapter(
                    columns=['1', '2'],
                ),
                multi_select=True,
                selected_row='selected_rows',
            ),
        ),
        width=500,
        height=500,
    )

    @on_trait_change('selected_rows')
    def _handle_selected_rows(self):
        print(self.selected_rows)


Demo().configure_traits()
```

To reproduce, run the above script, click on the first row, and then hold shift and click on the second row. In 6.1.1, the first two rows will be highlighted. In 6.1.2, only the second row will be highlighted (after an instant where both rows are highlighted).

corranwebster:
Fixes #605 

The problem was a `clicked` handler which was assuming single select.

### DIFF
--- a/traitsui/qt4/tabular_editor.py
+++ b/traitsui/qt4/tabular_editor.py
@@ -358,13 +358,15 @@ class TabularEditor(Editor):
         """ When mouse is clicked on a specific cell, update the selected
             indices first
         """
-        self.selected_row = self.clicked.row
-        self.selected_column = self.clicked.column
+        if not self.factory.multi_select:
+            self.selected_row = self.clicked.row
+            self.selected_column = self.clicked.column
 
     def _column_clicked_changed(self):
         """ When column is clicked, update selected column first
         """
-        self.selected_column = self.clicked.column
+        if not self.factory.multi_select:
+            self.selected_column = self.clicked.column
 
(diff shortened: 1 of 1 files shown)


corranwebster:
@scott-maddox can you please verify that this fixes your issue.

scott-maddox:
That does, indeed, appear to fix the issue! Might be worth adding a regression test based on the repro code, if that's not too hard.

corranwebster:
Unfortunately, the tabular editor tests are not in a state add a regression test easily - something which I'm working on rectifying, but there's a lot of testing infrastructure needed first.

PR #635, opened 2019-08-19
LED demo for QT backend

### LINKED ISSUE #630: Add Qt version of LedEditor
Qt has a QLCDNumber widget that displays a value in a similar way to the WxPython LEDNumberCtrl.  It has some differences, however:

- can't set alignment (probably easiest just to warn if set to something other than right)
- fixed number of digits (but can be changed easily, so probably just want to set it to the length of the string being set).

This editor is a bit of a gimmick, but it would be a good first issue for someone wanting to get into writing editors.

sandhya-sago:
Implement LEDEditor for Qt toolkits

Fixes #630


examples/demo/Extras/LED_display_qt.py
```diff
@@ -0,0 +1,79 @@
...
+
+"""
+This demo illustrates use of the LEDEditor for displaying numeric values
+using a simulated LED display control.
+"""
+
+from threading import Thread
+from time import sleep
+
+from traits.api \
```
corranwebster: This is an old style of import line continuation that we no-longer prefer.  Current Enthought style prefers one of:

- All on one line:
  ```
  from traits.api import HasTraits, Instance, Int, Bool, Float
  ```
  if less than 79 chars; or

- using parenthesis to wrap
  ```
  from traits.api import (
      HasTraits, Instance, Int, Bool, Float
  )
  ```
  if 80 chars or more.

corranwebster:
This looks basically good.  A couple of comments:

When creating a PR like this, it's good to have at least a minimal amount of text describing what it is (eg. something like "Implement `LEDEditor` for Qt toolkits." is enough).  This makes it easier to create the change log when it is time for the next release.  Similarly, as I mentioned before, if you say `Fixes #630` in the description, GitHub will take care of closing the related issue automatically when this PR is merged.

One improvement that we should do is to merge the two demo examples.  Right now the wx and qt versions have hard-coded imports from the respective backends, so they can't be used with the other backend.  One way around this is to introspect the backend using ``ETSConfig``, something like:
```
from traits.etsconfig.api import ETSConfig

if ETSConfig.toolkit == 'wx':
    from traitsui.wx.extra.led_editor import LEDEditor
else:
    from traitsui.qt4.extra.led_editor import LEDEditor
```
Ideally we's like to fix things so we didn't need to do this and could do something like `from traitsui.editors.extra.led_editor import LEDEditor` but that can be in a different PR.

traitsui/qt4/extra/led_editor.py
```diff
@@ -0,0 +1,20 @@
+from pyface.qt import QtGui
```
corranwebster: We need a copyright header here.  At a minimum it should be:
```
# (C) Copyright 2019 Enthought, Inc., Austin, TX
# All rights reserved.
```
but we usually also include some text pointing to the license, something like:
```
# This software is provided without warranty under the terms of the BSD
# license included in LICENSE.txt and may be redistributed only
# under the conditions described in the aforementioned license.  The license
# is also available online at http://www.enthought.com/licenses/BSD.txt

File: traitsui/qt4/extra/led_editor.py [hunk 3]
```diff
@@ -0,0 +1,20 @@
...
+        self.control.setSegmentStyle(QtGui.QLCDNumber.Flat)
+        self.set_tooltip()
+
+    def update_editor(self):
+        self.control.display(self.str_value)
+
+
+class LEDEditor(BasicEditorFactory):
+
+    # The editor class to be created:
```
corranwebster:
```suggestion
    #: The editor class to be created:
```
The `#:` means that the comment gets picked up by Sphinx auto-documenters.

examples/demo/Extras/LED_display_qt.py
sandhya-sago: Updated the imports.

traitsui/qt4/extra/led_editor.py
sandhya-sago: Added copyright

traitsui/qt4/extra/led_editor.py
sandhya-sago: Not sure if this has to be part of the documentation, but changed.

traitsui/qt4/extra/led_editor.py
corranwebster: The copyright was only added to the demo file, not this one.

examples/demo/Extras/LED_display.py
```diff
@@ -1,26 +1,30 @@
-#  Copyright (c) 2007, Enthought, Inc.
-#  License: BSD Style.
+# (C) Copyright 2019 Enthought, Inc., Austin, TX
```
corranwebster:
```suggestion
# (C) Copyright 2007-19 Enthought, Inc., Austin, TX
```

Issue #632, opened 2019-08-16
RangeEditor needs to handle invalid model values better


corranwebster:
It is possible that the underlying value being displayed by a `RangeEditor` is legitimately holding a non-numeric value (eg. see the `CompoundEditor` demo).  Currently the logic around handling changes of the value to an invalid quantity don't work well.

This is probably just a matter of putting better exception handling on the various `update` methods on the editor.

Both backends are bad, just in different ways.

sandhya-sago:
Corran, I do not see any error for QT.

PR #637, opened 2019-08-20
Fix compound_editor

sandhya-sago:
#632 
Handle exception in range editor when using compound_editor.

### DIFF
--- a/traitsui/wx/range_editor.py
+++ b/traitsui/wx/range_editor.py
@@ -294,8 +294,11 @@ class SimpleSliderEditor(BaseRangeEditor):
         """ Returns the slider setting corresponding to the user-supplied value.
         """
         if self.high > self.low:
-            ivalue = int((float(value - self.low) /
+            try:
+                ivalue = int((float(value - self.low) /
                           (self.high - self.low)) * 10000.0)
+            except TypeError:
+                ivalue = self.low
         else:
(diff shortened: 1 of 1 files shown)


traitsui/wx/range_editor.py
```diff
@@ -244,6 +244,9 @@ def update_object_on_enter(self, event):
                 # Pretend it didn't happen (i.e. do not change self.value).
                 value = self.value
                 self.control.text.SetValue(str(value))
+                # for compound editor, value may be non-numeric
+                if not isinstance(value, (int, float)):
```
corranwebster: This fails with long ints on Python 2.7 (which we still have to support at least until the end of the year) as well as likely other numeric types which might be reasonable (eg. NumPy scalar types, standard lib `Decimal` and `Rational`).

Additionally, we want to run as much of the subsequent code as we can, since we want to reset colors and error handling if the value is valid for the underlying trait.  I think a better approach is to wrap the `self.control.slider.SetValue` in a `try...except...` ("it's better to ask forgiveness" in this case, as opposed to "look before you leap")

sandhya-sago: I have made the changes for wx.
For QT, the code looks for
`if not isinstance(value, (int, float)): ...
`
Maybe this should be changed as well?

Issue #511, opened 2019-03-04
Add an updated event trait to ListStrEditor


corranwebster:
Like the `TabularEditor`, there may be times when you want to listen for a trait that indicates that the state of the editor has been changed so much that it should be completely refreshed.  Having the ability to specify a trait name to listen to for this would be useful, the implementation would be essentially the same as the `TabularEditor`.

PR #647, opened 2019-08-27
Add update trait to list editor

skailasa:
Addressing #511, is this all that was meant by this issue?


corranwebster:
It's part of it, but the real work is in the `traitsui/qt4/list_str_editor.py` and `traitsui/wx/list_str_editor.py` which would need to look for this trait on the editor factory and then hook up an `on_trait_change` (or perhaps `sync_value`) that calls the appropriate update method.

PR #650, opened 2019-08-28
Fix dataframe editor errors

corranwebster:
Catches errors when setting values in a DataFrame editor and logs at debug level rather than reporting to output.  Also cleans up setting logic to remove warnings.  And adds tests for setting data.


corranwebster:
Arg - bad branching. Closing PR.

PR #652, opened 2019-08-29
Fix tabular editor column widths on Qt

### LINKED ISSUE #638: Qt Tabular Editor doesn't implement fractional or negative widths
According to the TabularAdapter `get_width` docs:
```
        If the value is <= 0, the column will have a *default* width, which is
        the same as specifying a width of 0.1.
       ...
        If the value is a float such that 0.0 < value <= 1.0, it is treated as
        the *unnormalized fraction of the available space* that is to be
        assigned to the column. What this means requires a little explanation.
```
These cases are ignored by the Qt backend, and moreover, positive integers are treated as a hint rather than a fixed value (ie. the expected behaviour of the negative case).

corranwebster:
This PR does the following:

- adds a (tested!) algorithm for computing column widths for Table and Tabular editors similar to the way it specified in the documentation
- uses the algorithm in the Qt TabularEditor implementation
- automatically updates on resize of table or user modification of columns, preserving user choices.

The end result is _much_ nicer behaviour when re-sizing the columns.

Fixes #638.

### DIFF
--- a/traitsui/helper.py
+++ b/traitsui/helper.py
@@ -1,6 +1,6 @@
 # ------------------------------------------------------------------------------
 #
-#  Copyright (c) 2005, Enthought, Inc.
+#  Copyright (c) 2005-19, Enthought, Inc.
 #  All rights reserved.
 #
 #  This software is provided without warranty under the terms of the BSD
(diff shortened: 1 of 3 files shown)


corranwebster:
@sandhya-sago If you have time, a review of this would be useful.

corranwebster:
@sandhya-sago I am not seeing the issues which you were reporting - I think you may have been looking at the TableEditor examples, not the TabularEditor.

corranwebster:
We need this for a client project, so I'm going to self-merge.  Any issues can be cleaned up in a later PR.

PR #664, opened 2019-09-12
Add a TraitsUITestAssistant class

corranwebster:
Second attempt.

This adds a subclass of `GUITestAssistant` with convenience methods for testing TraitsUI GUIs.

The API is currently experimental.

### DIFF
--- a/traitsui/qt4/testing/traitsui_test_assistant.py
+++ b/traitsui/qt4/testing/traitsui_test_assistant.py
@@ -1,0 +1,142 @@
+#  Copyright (c) 2019, Enthought, Inc.
+#  All rights reserved.
+#
+#  This software is provided without warranty under the terms of the BSD
+#  license included in LICENSE.txt and may be redistributed only
+#  under the conditions described in the aforementioned license.  The license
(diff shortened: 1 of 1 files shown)


traitsui/qt4/testing/traitsui_test_assistant.py
```diff
@@ -0,0 +1,142 @@
...
+        loop until the underlying toolkit object is destroyed.
+
+        Parameters
+        ----------
+        object : HasTraits instance or None
+            The object to call edit_traits on.  If None, then the method
+            will attempt to use self.object, if it exists.
+        **kwargs : keyword arguments
+            Additional keyword arguments will be passed through to
+            edit_traits.
```
kitchoi: Drive-by comment: As a user of this method, I would not mind having to pass the entire dictionary of keyword arguments as one argument, e.g.  `create_ui(object=some_instance, edit_kwargs=dict(view=my_view))`  and then `object.edit_traits(**edit_kwargs)`. Then there will be room for future extension to the signature of `create_ui`.

PR #689, opened 2019-12-03
Move HasDynamicViews from traits

corranwebster:
It didn't really belong in Traits, moved into `traitsui.extras` and updated demo.


corranwebster:
Yep, straight copy other than fixing imports.

Issue #698, opened 2019-12-19
Unify Docstring processing between demo.py and tutor.py


corranwebster:
The demo.py code simply extracts the raw docstring and embeds it into an HTML widget:

- extraction is here: https://github.com/enthought/traitsui/blob/master/traitsui/extras/demo.py#L105
- the docstring ends up stored here: https://github.com/enthought/traitsui/blob/master/traitsui/extras/demo.py#L423

This means no markup is possible, in particular Sphinx-style docstrings get badly mangled.

On the other hand tutor.py uses `docutils` to produce nice HTML from Sphinx-style formatting: https://github.com/enthought/traitsui/blob/maint/620_move_tutor/traitsui/extras/tutor.py#L1274
and https://github.com/enthought/traitsui/blob/maint/620_move_tutor/traitsui/extras/tutor.py#L1444

It would be great if some of the code parsing and html generation could be made common to the two libraries as a first step to getting things unified.

PR #702, opened 2019-12-20
Use a common utility to generate HTML from RST in demo.py and tutor.py

shoeb-github:
contributes to #698. 
This PR unifies the document text generation between **demo.py** and **tutor.py** by using utility functions which are thin wrappers around `docutils` to generate html from reStructuredText.. A future PR would unify the text parsing between **demo.py** and **tutor.py**.

### DIFF
--- a/traitsui/extras/demo.py
+++ b/traitsui/extras/demo.py
@@ -861,25 +861,30 @@ class Demo(HasPrivateTraits):
 
 
 # -------------------------------------------------------------------------
-#  Function to run the demo:
+#  Utilities to convert rst strings/files to html.
 # -------------------------------------------------------------------------
 
 
-def demo(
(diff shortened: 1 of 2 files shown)


shoeb-github:
Some of the docstrings in the demos use html markup. This works in our case because the `docutils` api we use (with the default restructuredtext parser) simply embeds the string in html body. May be these should be re-written as restructuredtext.

traitsui/extras/demo.py
```diff
@@ -539,7 +541,7 @@ def _get_description(self):
         if self._description is None:
             self._get_init()
 
-        return self._description
+        return publish_html_str(self._description)
```
corranwebster: This is an expensive operation, so we should use a `@cached_property` rather than regenerating every time.

shoeb-github: Fixed in 36adae8

PR #705, opened 2019-12-24
Add docutils as an optional dependency of traitsui

### LINKED ISSUE #704: Make docutils an optional dependency of traitsui
`traitsui/extras/demo.py` and `tutor.py` require `docutils` to convert restructured text into html. It is not clear if `docutils` should be made a dependency of `traitsui` because its not being used anywhere else. After offline discussion, we decided to make it an `extras` dependency in `setup.py`. In future if HTML editor in `traitsui` requires `docutils` it could be made a core dependency.

shoeb-github:
closes #704

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -107,6 +107,7 @@ dependencies = {
     "nose",
     "coverage",
     "configobj",
+    "docutils"
 }
 
 extra_dependencies = {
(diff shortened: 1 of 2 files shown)


etstool.py
```diff
@@ -107,6 +107,7 @@
     "nose",
     "coverage",
     "configobj",
+    "docutils"
```
midhun-pm: should it be here or under extra_dependencies?

shoeb-github: `extra_dependencies` is being used to get dependencies for a particular toolkit. Also, other optional dependencies are listed under `dependencies`.

PR #706, opened 2019-12-27
Add scrollable trait to the ListEditor

mayankmanj:
The `scrollable` trait of the `ListEditor` allows us to hide the vertical scrollbar, and display all the items contained in the list.

Drive-by fixes -- Removes some duplicate code.

### DIFF
--- a/docs/source/traitsui_user_manual/factories_basic.rst
+++ b/docs/source/traitsui_user_manual/factories_basic.rst
@@ -905,7 +905,7 @@ ListEditor()
 :Default for:
     List [18]_
 :Optional parameters:
-    *editor*, *rows*, *style*, *trait_handler*, *use_notebook*
+    *editor*, *rows*, *style*, *scrollable*, *trait_handler*, *use_notebook*
 
     The following parameters are used only if *use_notebook* is True:
     *deletable*, *dock_style*, *export*, *page_name*, *select*, *view*
@@ -923,11 +923,13 @@ reordering items within the list.
 The simple style displays a single item at a time, with small arrows on the
 right side to scroll the display. The custom style shows multiple items. The
 number of items displayed is controlled by the *rows* parameter; if the number
-of items in the list exceeds this value, then the list display scrolls. The
-editor used for each item in the list is determined by the *editor* and *style*
-parameters. The text style of list editor is identical to the custom style,
-except that the editors for the items are text editors. The read-only style
-displays the contents of the list as static text.
+of items in the list exceeds this value, then the list display scrolls. If the
+*scrollable* parameter is False, the editor displays all objects in the list and
+does not render the vertical scrollbar. The editor used for each item in the list
+is determined by the *editor* and *style* parameters. The text style of list
+editor is identical to the custom style, except that the editors for the items
+are text editors. The read-only style displays the contents of the list as
+static text.
 
 By default, the items use the trait handler appropriate to the type of items in
 the list. You can specify a different handler to use for the items using the
--- a/traitsui/editors/list_editor.py
+++ b/traitsui/editors/list_editor.py
@@ -86,7 +86,7 @@ class ToolkitEditorFactory(EditorFactory):
     mutable = Bool(True)
 
     #: Should the scrollbars be displayed if the list is too long.
-    scrollable = Bool(True)
+    scrollable = Bool(True, sync_value=True)
 
     #: The style of editor to use for each item:
     style = style_trait
--- a/traitsui/editors/list_editor.py
+++ b/traitsui/editors/list_editor.py
@@ -85,6 +85,9 @@ class ToolkitEditorFactory(EditorFactory):
     #: Can the list be reorganized, or have items added and deleted.
     mutable = Bool(True)
 
+    #: Should the scrollbars be displayed if the list is too long.
+    scrollable = Bool(True)
+
     #: The style of editor to use for each item:
     style = style_trait
 
--- a/traitsui/qt4/list_editor.py
+++ b/traitsui/qt4/list_editor.py
@@ -48,6 +48,9 @@ class SimpleEditor(Editor):
     #: Is the list of items being edited mutable?
     mutable = Bool(True)
 
+    #: Is the editor scrollable?
+    scrollable = Bool(True)
+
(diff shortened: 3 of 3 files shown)


traitsui/qt4/list_editor.py
```diff
@@ -382,6 +382,21 @@ def _mutable_default(self):
         """
         return self.factory.mutable
 
+    def _scrollable_default(self):
+        """ Trait handler to set the scrollable from the factory.
+        """
+        return self.factory.scrollable
+
+    def __list_pane_default(self):
```
corranwebster: I'd just create this directly in the init; or if you want to encapsulate, have this as a private method which creates the `list_pane` attribute.

traitsui/qt4/list_editor.py
```diff
@@ -742,21 +750,3 @@ def _menu_action(self, event, name=""):
         else:
             # TODO: Fix tab order based on the context_object's list
             self.control.addTab(self._pagewidgets[name], name)
-
-    def _context_menu_requested(self, event):
```
corranwebster: Why is this removed?  This seems unrelated to the main change?

traitsui/qt4/list_editor.py
```diff
@@ -382,6 +382,21 @@ def _mutable_default(self):
         """
         return self.factory.mutable
 
+    def _scrollable_default(self):
```
corranwebster: You don't have a `scrollable` trait declared on the list editor (as opposed to the list editor factory).  So this _probably_ isn't working the way that you expect.

traitsui/qt4/list_editor.py
```diff
@@ -221,10 +222,9 @@ def update_editor_item(self, event):
 
         # Otherwise, find the proxy for this index and update it with the
         # changed value:
-        for control in self.control.widget().children():
-            if isinstance(control, QtGui.QLayout):
-                continue
-
+        children = (self._layout.itemAt(ind).widget()
+                    for ind in range(self._layout.count()))
+        for control in children:
```
corranwebster: A better way to do this would just be:
```
for control in self._list_pane.children():
```
I think.  Then you don't need to track the `_layout` specifically as a trait.

traitsui/qt4/list_editor.py
```diff
@@ -96,20 +98,19 @@ def init(self, parent):
...
         # Create a mapper to identify which icon button requested a contextmenu
         self.mapper = QtCore.QSignalMapper(self.control)
 
         # Create a widget with a grid layout as the container.
-        self._list_pane = QtGui.QWidget()
-        self._list_pane.setSizePolicy(
-            QtGui.QSizePolicy.Expanding, QtGui.QSizePolicy.Expanding
-        )
-        layout = QtGui.QGridLayout(self._list_pane)
+        self._layout = layout = QtGui.QGridLayout(self._list_pane)
```
corranwebster: `_layout` isn't declared as a trait on the class (but I don't think you actually need it - see previous comment).

traitsui/editors/list_editor.py
```diff
@@ -85,6 +85,9 @@ class ToolkitEditorFactory(EditorFactory):
     #: Can the list be reorganized, or have items added and deleted.
     mutable = Bool(True)
 
+    #: Should the scrollbars be displayed if the list is too long.
+    scrollable = Bool(True)
```
corranwebster: If you declare as:
```
scrollable = Bool(True, sync_value=True)
```
then the value will be copied from the factory to the actual editor automatically.

This is a new-ish feature.

mayankmanj: This is not working for me -- the value is not being copied from the factory to the editor. I have defined a `scrollable` trait in the editor (set to `True` by default), and removed the `_scrollable_default` method that was copying the value earlier.

traitsui/qt4/list_editor.py
mayankmanj: The `scrollable` was present in the parent class `UIEditor` -- but I realize now that it is set to `False` by default. Setting default to `True` here.

traitsui/qt4/list_editor.py
mayankmanj: This is duplicate code from lines [728-744](https://github.com/enthought/traitsui/blob/35f0510ca3090574f121786221548b09ac1882a8/traitsui/qt4/list_editor.py#L728-L744) above.

Would you like me to address this in a separate PR instead?

docs/source/traitsui_user_manual/factories_basic.rst
```diff
@@ -923,11 +923,13 @@ reordering items within the list.
 The simple style displays a single item at a time, with small arrows on the
 right side to scroll the display. The custom style shows multiple items. The
 number of items displayed is controlled by the *rows* parameter; if the number
```
mayankmanj: I think the `rows` parameter is used only in WX, not in QT. Should this be mentioned here?

mayankmanj: Maybe a better solution for controlling the display of scrollbar is simply to respect the `rows` parameter in `QT`.

traitsui/qt4/list_editor.py
corranwebster: Ah.  That makes sense - it wasn't obvious from the diff.  It's fine to do drive-by fixes like this, just mention it in the PR description at the top that you are also doing this change.

docs/source/traitsui_user_manual/factories_basic.rst
corranwebster: Rows is meant to be a vertical size hint.  I can see the temptation to try and use it, but we don't support changing it dynamically at the moment, which is what would be needed for removing scrollability in the general case.  Something like the current PR is a better approach.

It wouldn't be a bad idea to be able to support the rows parameter in the Qt backend, but that should probably be a different PR (and obviously would be ignored if Scrollable is False.

traitsui/editors/list_editor.py
corranwebster: Hmmm... possibly a bug somewhere, then, but this line should copy things from the factory to the editor: https://github.com/enthought/traitsui/blob/master/traitsui/editor.py#L584

mayankmanj:
Thank you! Merging.

Issue #696, opened 2019-12-17
Unify demo and tutor codebase


corranwebster:
These two example running tools are very similar.  It would be good to unify them in some way.

Issue #24, opened 2011-09-28
TabularEditor: broken behavior of left and right cursor keys 


WarrenWeckesser:
In OS X, wx backend:

In a Tabular editor, the left and right cursor keys are supposed to move the currently selected row up and down, respectively.  This appears to be broken.  Run either of these demos and hit the left and right cursor keys several times:
    traitsui/examples/demo/Advanced/Tabular_editor_demo.py
    traitsui/examples/demo/Advanced/Auto_update_TabularEditor_demo.py
The editor appears to lose track of what row is currently selected.

PR #710, opened 2019-12-31
[WIP]: Use demo.py to implement tutorials

shoeb-github:
xref: #696 

This is an implementation of `tutor.py` using  `demo.py`.

The approach is to extend the `DemoPath` class to also parse file extensions other than `.py` like `.rst`, `.htm`. However, this does not look clean. It is probably better to do this unification as outlined in issue  [#24](https://github.com/enthought/ets-examples/issues/24) of ETS examples.

Things that don't work:
- no support for `.desc`, `.swf` files.

To do:
- [ ] Add `next`, `previous` buttons in the UI to navigate the tutorial directory structure.


corranwebster:
Rather than subclassing, how much of this functionality can be merged into the `demo.py` classes? In other words, can we enhance the demos with tutor functionality without breaking them?

Or alternatively, how much functionality can we drop from the tutor.py code?

Also, I don't think we have _any_ actual tutorials using videos or similar, so we can drop support for a lot of the file formats without losing anything.

traitsui/extras/tutor.py
```diff
@@ -873,729 +191,285 @@ def run_code(self):
...
 
-    def _add_htm_item(self, path):
-        """ Creates a description item for an HTML file.
+    def _parse_html_file(self, path, name):
+        """ Parse a html file
         """
-        # Check if there is a corresponding .rst (restructured text) file:
-        dir, base_name = os.path.split(path)
-        rst = os.path.join(dir, os.path.splitext(base_name)[0] + '.rst')
+        return self._parse_htm_file(path, name)
```
corranwebster: This just does infinite recursion? (it looks like the original version did as well)

I think we can just remove this.

traitsui/extras/tutor.py
```diff
@@ -873,729 +191,285 @@ def run_code(self):
...
-    # Should the Python code be automatically executed on start-up?
-    auto_run = Bool(False)
+        return dirs + files
 
-    #-- Event Handlers -------------------------------------------------------
+    #-- Factory Methods for Creating Tutorial Files Based on File Type --------
 
-    def _path_changed(self, path):
-        """ Creates the appropriate section based on the value of the path.
+    def _parse_avi_file(self, path, name):
```
corranwebster: This can be safely removed.

traitsui/extras/tutor.py
```diff
@@ -873,729 +191,285 @@ def run_code(self):
...
-    def _add_swf_item(self, path):
-        """ Creates a description item for a Flash file.
-        """
-        if is_windows and wx_available:
-            self.descriptions.append(FlashItem(path=path))
+        return self._parse_jpg_file(path, name)
 
-    def _add_mov_item(self, path):
-        """ Creates a description item for a QuickTime movie file.
+    def _parse_mov_file(self, path, name):
```
corranwebster: This can be removed

traitsui/extras/tutor.py
```diff
@@ -873,729 +191,285 @@ def run_code(self):
...
-    def _add_avi_item(self, path):
-        """ Creates a description item for an AVI movie file.
-        """
-        self._add_wmv_item(path)
+        html = QTMovieTemplate % (path2, path2)
+        return TutorialHTMLFile(description=html, name=name, parent=self)
 
-    def _add_jpg_item(self, path):
-        """ Creates a description item for a JPEG image file.
+    def _parse_mp3_file(self, path, name):
```
corranwebster: This can be removed.

traitsui/extras/tutor.py
```diff
@@ -873,729 +191,285 @@ def run_code(self):
...
+        path = self.path
+        for name in os.listdir(path):
+            cur_path = os.path.join(path, name)
+            if os.path.isdir(cur_path):
+                dirs.append(TutorialPath(name=name, parent=self))
+            elif self.use_files:
+                name, ext = os.path.splitext(name)
+
+                # If we have a handler for the file type, invoke it:
+                method = getattr(
```
corranwebster: If you wanted to do a bit better of a refactor, replacing "magic method names" with a dictionary mapping file extensions to factory functions would likely make the code much cleaner, and might allow something like having a tutorial be the same base class as a demo but with a slightly different set of registered file handlers, assuming you can't get the codebases completely aligned.

corranwebster: The factory functions could be class methods on the appropriate classes being created.

traitsui/extras/tutor.py
```diff
@@ -873,729 +191,285 @@ def run_code(self):
...
-
-        # Add the resulting item to the descriptions list:
-        self.descriptions.append(item)
+        text = _read_file(path)
+        return TutorialTxtFile(description=text, name=name, parent=self)
 
-    def _create_html_item(self, **traits):
-        """ Creates a platform specific html item and adds it to the list of
-            descriptions.
+    def _parse_wmv_file(self, path, name):
```
corranwebster: This can safely be removed.

shoeb-github:
> Rather than subclassing, how much of this functionality can be merged into the `demo.py` classes? In other words, can we enhance the demos with tutor functionality without breaking them?

Or alternatively, how much functionality can we drop from the tutor.py code?

Also, I don't think we have _any_ actual tutorials using videos or similar, so we can drop support for a lot of the file formats without losing anything. 



I think it is possible to add this functionality to the demo without altering the demo functionality.

shoeb-github:
closing in favor of #712`

PR #712, opened 2020-01-08
Delete tutor.py and add its functionality to demo.py

shoeb-github:
Contributes to #696 

Modified `demo.py`:
- to recognize file extensions other than `.py`.
- added navigation functionality to walk the directory tree (from `tutor.py`).

This still does not support `.desc` files from old tutorials. These are config files and it is probably better to rewrite these files in a format support by demo's `configobj`.


shoeb-github:
Yes, I will replace them in separate PR.

PR #733, opened 2020-02-04
Fix incorrect editor Undo action

### LINKED ISSUE #672: Undo reverts editors incorrectly
When undoing edits across multiple editors, the undo mechanism sometimes re-does an edit that has been undone previously.

To reproduce:
```

from traits.api import Float, HasStrictTraits, Unicode
from traitsui.api import Item, Menu, MenuBar, RedoAction, UndoAction, View


class StockItem(HasStrictTraits):
    """ An viewmodel for an item stored in a laboratory.

    This consists of a stock identifier together with a quantity.
    """

    #: The stock item name.
    stock = Unicode()

    #: The quantity of the stock in the storeroom.
    quantity = Float(1.0)

    traits_view = View(
        Item('stock'),
        Item('quantity'),
        buttons=['Cancel', 'OK'],
        title="Stock Item",
        menubar=MenuBar(
            Menu(
                UndoAction,
                RedoAction,
                name='Edit',
            )
        )
    )


if __name__ == '__main__':
    stock_item = StockItem()
    stock_item.configure_traits()
```

Then:

- type a single letter into the 'stock' editor
- change the 'quantity' editor value to 10 by inserting a 0
- undo via menu (works as expected, quantity goes back to 1)
- undo again (stock reverts as expected, but quantity becomes 10)

midhun-pm:
Fixes #672 

The method that logs changes made to an editor: `traitsui.editor.Editor.log_change` was being called when the user performs some new action on the editor and also when an undo/redo action is performed. 

This was resulting in a corrupted `history` in `traitsui/ui.py:103`

This PR checks adds a check by looking at the `updating` flag which indicates the source of the change whether on the editor itself or externally and only calls `log_change` if the change was made on the editor.  

I feel that only changes made by the user need to be logged and not external changes, can't think of a case where the user should be allowed to undo an external change.

Looking for suggestions how to test this thoroughly.

### DIFF
--- a/traitsui/editor.py
+++ b/traitsui/editor.py
@@ -282,9 +282,7 @@ class Editor(HasPrivateTraits):
         *undo_args
             Any arguments to pass to the undo factory.
         """
-        # Indicate that the contents of the user interface have been changed:
         ui = self.ui
-        ui.modified = True
 
         # Create an undo history entry if we are maintaining a history:
         undoable = ui._undoable
@@ -513,6 +511,9 @@ class Editor(HasPrivateTraits):
             self.item.style != "readonly"
             and object.base_trait(name).type != "event"
         ):
+            # Indicate that the contents of the UI have been changed:
+            self.ui.modified = True
+
             if self.updating:
                     self.log_change(
                     self.get_undo_item, object, name, old_value, new_value
--- a/traitsui/editor.py
+++ b/traitsui/editor.py
@@ -513,9 +513,10 @@ class Editor(HasPrivateTraits):
             self.item.style != "readonly"
             and object.base_trait(name).type != "event"
         ):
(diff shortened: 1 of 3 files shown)


corranwebster:
Have a look at https://github.com/enthought/traitsui/blob/master/traitsui/tests/test_editor.py for a fairly comprehensive set of tests involving the `Editor` infrastructure.  You may find that things fit naturally into that test suite, or you may find that you need more and a separate set of tests makes more sense, but the ideas in there are useful.

As far as _what_ to test, you are really wanting to test the sanity of the undo history, so that's what you should be testing: drive some changes; observe what happens to the history; undo the changes, make sure the history looks right; redo and check that things are still sane.

Issue #645, opened 2019-08-27
Use importlib_metadata and importlib_resources


corranwebster:
These replace common use cases of pkg_resources

corranwebster:
I think this is done.

PR #745, opened 2020-02-24
Qt VideoEditor

corranwebster:
This is a viewer-only Qt `VideoEditor`.  This only presents the video (provided as a URL), not any associated controls like play buttons, position sliders, etc. but you can bind traits to player state and then build a custom interface around them - the example gives a good idea of how to do this.

There are some glitches on MacOS retina displays which appear to be a problem with the underlying QtMultimedia implementation.

There is no Wx implementation. There are no tests yet.


JackDra:
The video streaming seems to work fine both locally and with the demo video link on my mac. 

The only issue I had was there is a weird background image displayed behind the video, its a black image half a cut off text field at the top of the video pane
Screen Shot 2020-03-02 at 9 11 35 AM

JackDra:
> The video streaming seems to work fine both locally and with the demo video link on my mac.
> 
> The only issue I had was there is a weird background image displayed behind the video, its a black image half a cut off text field at the top of the video pane

Sorry this problem was fixed when I built with the latest version of pyqt5 instead of pyside2.

jvkersch:
Some quick drive-by comments: on Linux (Ubuntu 18.04), the player refused to play video with EDM-installed PyQt5 and PySide2, with the error message 'no service found for - "org.qt-project.qt.mediaplayer"'. Installing pyqt5 from PyPI, as Jack suggested, made the video playable, but (a) it is rotated over 90 degrees, and (b) playback speed is very slow. Likely not an issue with TraitsUI, but suggestions for debugging (if any) are welcome.

jvkersch:
Some updates: @mdsmarte and I looked into running the video editor on Windows and Linux, and here are some findings. On Linux, the EDM-installed PyQt5 does not have the multimedia plugin and fails to play video. Using PyQt5 from PyPI instead works. As reported above, playback is very choppy both with local and remote videos. Commenting out the following line makes playback smooth again (comparable to playing the video in a native media player): https://github.com/enthought/traitsui/blob/2b8620cf2f672873cac21d2dbec69e45b04e0a35/traitsui/qt4/video_editor.py#L343.

In Windows 10, I see the same kind of choppiness, with the same workaround. Playback works out of the box with both the EDM and the PyPI version of PyQt5. I did have to install the necessary codecs.

traitsui/qt4/video_editor.py
```diff
@@ -0,0 +1,360 @@
...
+            self.media_content = QMediaContent(None)
+            self.control.updateGeometry()
+
+    # Signal handlers -------------------------------------------------------
+
+    def _state_changed_emitted(self, state):
+        self.state = reversed_state_map[state]
+
+    def _position_changed_emitted(self, position):
+        self.position = position / 1000.0
```
jwiggins, line 294: Regarding @jvkersch's comment about avoiding a call to `self.media_player.setPosition` in `_position_changed`, perhaps this should be `self.trait_setq(position=position / 1000)`?

The point is to keep the trait in sync, not set the position of the media player, right?

jwiggins:
@aaronayres35 Weren't you seeing intermittent failures around the `HTMLEditor` tests? Is there an issue?
```
FAIL: test_open_external_link (traitsui.tests.editors.test_html_editor.TestHTMLEditor)
  File "D:\a\traitsui\traitsui\.edm\envs\traitsui-test-3.6-pyside2\lib\site-packages\traitsui\tests\editors\test_html_editor.py", line 310, in test_open_external_link
    html_view.inspect(HTMLContent()),
AssertionError: 'External Link' not found in ''
```

traitsui/qt4/video_editor.py
```diff
@@ -0,0 +1,362 @@
...
+    muted = Bool(False)
+
+    #: The playback volume on a logarithmic scale (perceptually linear).
+    volume = Range(0.0, 100.0)
+
+    #: The playback rate.  Negative values rewind the video.
+    playback_rate = Float(1.0)
+
+    def _get_frame(self):
+        return self.control._np_image
```
jwiggins, line 181: This isn't needed, right? It assumes that `self.control` is an `ImageWidget` instance.

traitsui/qt4/video_editor.py
```diff
@@ -0,0 +1,362 @@
...
+        self._muted_changed()
+        self._volume_changed()
+        self._playback_rate_changed()
+
+        self._connect_signals()
+
+        self.set_tooltip()
+
+    def dispose(self):
+        self._disconnect_signals()
```
jwiggins, line 242: This should call `super().dispose()`, no?

traitsui/qt4/video_editor.py
```diff
@@ -0,0 +1,362 @@
...
+            self.media_content = QMediaContent(qurl)
+            self.control.updateGeometry()
+        else:
+            self.media_content = QMediaContent(None)
+            self.control.updateGeometry()
+
+    # Signal handlers -------------------------------------------------------
+
+    def _state_changed_emitted(self, state):
+        self.state = reversed_state_map[state]
```
jwiggins, line 291: Should this be wrapped by a `self.updating_value()` context?

traitsui/examples/demo/Standard_Editors/VideoEditor_demo.py
```diff
@@ -0,0 +1,158 @@
...
+                state='state',
+                position='position',
+                duration='duration',
+                video_error='error',
+                media_status='status',
+                buffer='buffer',
+                muted='muted',
+                volume='volume',
+                playback_rate='playback_rate',
+                image_fun='image_fun'
```
jwiggins, line 120: It would be nice not to require users to provide traits for all of this editor state. How do the trait declarations in `VideoEditor` need to change?

traitsui/examples/demo/Standard_Editors/VideoEditor_demo.py
```diff
@@ -0,0 +1,158 @@
+#  Copyright (c) 2020, Enthought, Inc.
+#  All rights reserved.
```
jwiggins, line 2: Note to self: This needs to be updated.

jwiggins (commented):
@corranwebster I'm happy to finish this PR but I have several questions first.

corranwebster:
Yes, feel free to run with this as needed (time permitting), making whatever changes you think make sense - it may make sense to strip out the image transform stuff, particularly if it affects performance.

traitsui/qt4/video_editor.py
corranwebster: Not needed, it's a convenience.  Could be a function instead.

traitsui/qt4/video_editor.py
```diff
@@ -0,0 +1,362 @@
...
+    def _get_frame(self):
+        return self.control._np_image
+
+    def update_to_regular(self):
+        self.control = QVideoWidget()
+        self.control.setSizePolicy(
+            QSizePolicy.Expanding, QSizePolicy.Expanding
+        )
+        self.control.setBackgroundRole(QPalette.Window)
+        self.media_player.setVideoOutput(self.control)
```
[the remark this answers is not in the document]
corranwebster: Not sure - it was a year ago that I wrote this.  There was some hackery needed to get display working properly on high-dpi Mac displays, so this might be part of that.

traitsui/qt4/video_editor.py
corranwebster: Yes, almost certainly.

traitsui/qt4/video_editor.py
corranwebster: Possibly - again, don't have the context loaded into my head, so it you think it is needed, do it!

traitsui/examples/demo/Standard_Editors/VideoEditor_demo.py
corranwebster: This editor was designed to be low-level and makes heavy use of the `sync_name` metadata, behaviour is explained here:
https://github.com/enthought/traitsui/blob/7f3dbaf6c30ed221e1aa680e0c9bb2a34c552b39/traitsui/editor.py#L594-L617

In particular users of the code are responsible for setting up any play/pause buttons, muting etc.  If they are left empty, then the default values from the Editor will be used, I think, so we should make sure that those are sensible.

If we want to allow constant values (eg. setting `muted` to `True` and not synchronising), then we might need to have `muted` and `muted_name` traits.

traitsui/qt4/video_editor.py
jwiggins: I'm leaving this alone for now. It's not as straightforward as `position` was.

traitsui/examples/demo/Standard_Editors/VideoEditor_demo.py
jwiggins: > If they are left empty, then the default values from the Editor will be used

This is true, but since the default values are empty strings it causes a `TraitError` if nothing is passed.

I don't understand the mechanisms in TraitsUI well enough to change this [even after some experimentation], so I'm going to leave it alone in this PR.

jwiggins:
This is ready for review if someone would like to take a look

jwiggins:
Hmm... Yeah. Speculative fixes can be risky.

JackDra:
> Yes, feel free to run with this as needed (time permitting), making whatever changes you think make sense - it may make sense to strip out the image transform stuff, particularly if it affects performance.

If there is more time available (which sounds like no), I had thought about having both video player and image transformed video player coded up, and if the user does not pass an image transformation function, it will fall back to the regular video player implementation.

jwiggins:
> if the user does not pass an image transformation function, it will fall back to the regular video player implementation.

I'm pretty sure that's already implemented.

rahulporuri:
@jwiggins can you please update the PR description - which probably is out-of-date.

jwiggins:
I'm not sure what should be changed in the description

traitsui/editors/video_editor.py
```diff
@@ -0,0 +1,86 @@
...
+    #: the video.  The referenced trait should be a Bool.
+    muted = Str(sync_value='from', sync_name='muted')
+
+    #: The name of a trait to synchronise with the audio volume of
+    #: the video player.  The referenced trait should be a Range(0.0, 100.0).
+    #: Values are on a logarithmic scale.
+    volume = Str(sync_value='from', sync_name='volume')
+
+    #: The name of a trait to synchronise with the audio muted state of
+    #: the video.  The referenced trait should be a Float.
```
rahulporuri, line 52:
```suggestion
    #: The name of a trait to synchronise with the playback rate of
    #: the video.  The referenced trait should be a Float.
```

traitsui/examples/demo/Standard_Editors/VideoEditor_demo.py
```diff
@@ -0,0 +1,166 @@
...
+    channels = qimage.pixelFormat().channelCount()
+    return np.array(
+        qimage.bits().asarray(width * height * channels)
+    ).reshape(height, width, channels).astype('u1')
+
+
+def qimage_function(antialiasing=True):
+    def antialis_fun(image_fun):
+        """
+        Turns a image funciton into a QImage function,
```
rahulporuri, line 48:
```suggestion
        Turns a image function into a QImage function,
```

traitsui/examples/demo/Standard_Editors/VideoEditor_demo.py
```diff
@@ -0,0 +1,166 @@
...
+    return image.transpose(1, 0, 2)
+
+
+class VideoEditorDemo(HasTraits):
+    """Defines the main VideoEditor demo class."""
+
+    #: The URL that holds the video data.
+    video_url = Str()
+
+    #: A button that
```
rahulporuri, line 76: Half written trait comment. It would be awesome if all traits in this class are documented.

traitsui/examples/demo/Standard_Editors/VideoEditor_demo.py
```diff
@@ -0,0 +1,166 @@
...
+
+    muted = Bool(True)
+
+    volume = Range(0.0, 100.0, value=100.0)
+
+    playback_rate = Float(1.0)
+
+    image_fun = Callable()
+
+    def _state_changed(self, new):
```
rahulporuri, line 101: we'd prefer not introducing new `_*_changed` methods if that's possible. can this and `_play_pause_button_changed` be rewritten with `observe`?

traitsui/examples/demo/Standard_Editors/VideoEditor_demo.py
```diff
@@ -0,0 +1,166 @@
...
+        buttons=['OK'],
+        width=800,
+        height=600,
+        resizable=True,
+    )
+
+
+# Create the demo:
+demo = VideoEditorDemo(image_fun=test_image_fun)
+# demo = VideoEditorDemo()
```
rahulporuri, line 161:
```suggestion
```

traitsui/examples/demo/Standard_Editors/VideoEditor_demo.py
```diff
@@ -0,0 +1,166 @@
...
+demo.video_url = "https://commondatastorage.googleapis.com/gtv-videos-bucket/sample/ForBiggerEscapes.mp4"  # noqa: E501
```
rahulporuri, line 162: nit - not sure why we don't just set `video_url` when instantiating the demo object.

traitsui/qt4/video_editor.py
```diff
@@ -0,0 +1,365 @@
...
+    This editor uses the Qt QMultimedia machinery to display video streams
+    to the screen.  Rather than being self-contained, the editor only concerns
+    itself with displaying the video, and provides traits that control
+    behaviour and provide internal state of the control during playback.
+    """
+
+    #: does the drawing onto the image plane
+    control = Instance(QVideoWidget)
+
+    #: handels the image pulling so the frames can be processed.
```
rahulporuri, line 132:
```suggestion
    #: handles the image pulling so the frames can be processed.
```

traitsui/qt4/video_editor.py
```diff
@@ -0,0 +1,365 @@
...
+
+    #: The current playback position of the player, synchronized to the trait
+    #: named by factory.position.
+    position = Float()
+
+    #: The total duration of the current video, synchronized to the trait
+    #: named by factory.duration.
+    duration = Float()
+
+    #: The media player playback status (loading, buffering, etc.).
```
rahulporuri, line 156: can we update this - and other traits (`buffer`, `video_error`, `muted`, `volume`, `playback_rate` and `image_fun`) - to point out that these traits too are synced to the respective factory.X traits?

traitsui/qt4/video_editor.py
```diff
@@ -0,0 +1,365 @@
...
+    muted = Bool(False)
+
+    #: The playback volume on a logarithmic scale (perceptually linear).
+    volume = Range(0.0, 100.0)
+
+    #: The playback rate.  Negative values rewind the video.
+    playback_rate = Float(1.0)
+
+    #: Function to apply to the image. Takes ref to new frame and a size tuple
+    image_fun = Callable()
```
rahulporuri, line 175: this is nit-picky and bikesheddy but can we use `image_func` instead of `image_fun`? I've seen `*func` a lot but i don't think ive ever seen `*fun`.

traitsui/qt4/video_editor.py
```diff
@@ -0,0 +1,365 @@
...
+            QSizePolicy.Expanding, QSizePolicy.Expanding
+        )
+        self.control.setBackgroundRole(QPalette.Window)
+
+        self.surface = VideoSurface(widget=self.control)
+        self.surface.frameAvailable.connect(self.control.setImage)
+
+        self.media_player.setVideoOutput(self.surface)
+
+    def _image_fun_changed(self):
```
rahulporuri, line 201: same comment as before regd using `observe` instead of `_*_changed` methods

traitsui/qt4/video_editor.py
```diff
@@ -0,0 +1,365 @@
...
+        if self.image_fun is None:
+            self.update_to_regular()
+        else:
+            self.update_to_functional()
+
+    # ------------------------------------------------------------------------
+    # Editor interface
+    # ------------------------------------------------------------------------
+
+    def init(self, parent):
```
rahulporuri, line 211: We're passing in the `parent` but we are never using it. Not sure if we need to be though.

traitsui/qt4/video_editor.py
```diff
@@ -0,0 +1,365 @@
...
+
+        self.control = QVideoWidget()
+        self.control.setSizePolicy(
+            QSizePolicy.Expanding, QSizePolicy.Expanding
+        )
+        self.control.setBackgroundRole(QPalette.Window)
+        self.media_player.setVideoOutput(self.control)
+
+    def update_to_functional(self):
+        self.control = ImageWidget(image_fun=self.image_fun)
```
rahulporuri, line 190: should we be setting `widget` here like we are with `VideoSurface`?

traitsui/qt4/video_editor.py
rahulporuri: looks like there are more `_*_changed` methods in this class - for the other synced traits. can we please rewrite those event handlers as well?

traitsui/tests/editors/test_video_editor.py
```diff
@@ -0,0 +1,68 @@
...
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+import unittest
+
+import pkg_resources
```
rahulporuri, line 12: Can we use `importlib_resources` instead of `pkg_resources` here? We are trying to remove uses of `pkg_resources` from the codebases.

rahulporuri (approved):
Mostly LGTM with a bunch of comments.

I guess the video editor isn't ready for prime time yet so we will keep it out of `traitsui.api` for now. When we do decide to include it in the `api` module though, we will probably want to also include the video editor-specific traits defined in the factory module.

rahulporuri:
I LGTM'ed too early. I ran the example on windows and it isnt working as expected.

<details><summary>win + cp36 + pyqt5 5.14.2-4 and qt 5.12.6-10</summary>

```
# Packages in environment 'traitsui-test-3.6-pyqt5'
#  prefix: 'C:\Users\rporuri\.edm\envs\traitsui-test-3.6-pyqt5'
appdirs              1.4.3-1        enthought/free
apptools             5.1.0-1        enthought/free
bzip2_runtime        1.0.8-1        enthought/free
chaco                4.8.0-7        enthought/free
configobj            5.0.6.post9-1  enthought/free
coverage             4.3.4-1        enthought/free
distribute_remove    1.0.0-4        enthought/free
enable               5.1.0-1        enthought/free
flake8               3.8.4-3        enthought/free
flake8_ets           1.0.0-1        enthought/free
fonttools            3.29.1-1       enthought/free
freetype             2.10.0-7       enthought/free
h5py                 2.10.0-3       enthought/free
harfbuzz             2.6.4-4        enthought/free
icu4c                50.2-2         enthought/free
importlib_metadata   2.0.0-1        enthought/free
importlib_resources  3.3.0-1        enthought/free
libblosc             1.20.1-2       enthought/free
liblcms2             2.11-1         enthought/free
liblz4               1.9.2-1        enthought/free
libopenjpeg          2.4.0-2        enthought/free
libpng               1.6.37-1       enthought/free
libtiff              4.0.10-9       enthought/free
libzstd              1.4.8-1        enthought/free
mccabe               0.6.1-1        enthought/free
mkl                  2018.0.3-2     enthought/free
numexpr              2.6.9-3        enthought/free
numpy                1.17.4-1       enthought/free
packaging            16.8-4         enthought/free
pandas               0.25.3-3       enthought/free
pillow               7.2.0-2        enthought/free
pip                  20.0.2-3       enthought/free
pycodestyle          2.6.0-1        enthought/free
pyface               7.3.0-1        enthought/free
pyflakes             2.2.0-1        enthought/free
pygments             2.2.0-1        enthought/free
pyparsing            2.2.0-1        enthought/free
pyqt5                5.14.2-4       enthought/free
pyqt5_sip            12.7.2-1       enthought/free
pytables             3.5.1-7        enthought/free
python_dateutil      2.8.0-4        enthought/free
pytz                 2020.1-1       enthought/free
qt                   5.12.6-10      enthought/free
setuptools           41.6.0-4       enthought/free
six                  1.15.0-1       enthought/free
traits               6.2.0-1        enthought/free
traitsui             7.1.1-4        enthought/free
zipp                 3.4.0-2        enthought/free
zlib_runtime         1.2.11-1       enthought/free
```

</details>

video-editor-sigh

Note that I also see the following on the command line when i run the example

```
>python VideoEditor_demo.py
DirectShowPlayerService::doRender: Unresolved error code 0x80040218 (IDispatch error #24)
DirectShowPlayerService::doRender: Unresolved error code 0x80040218 (IDispatch error #24)
```

jwiggins:
@rahulporuri I believe you need to install codecs on Windows in order for Qt Multimedia to work properly.

traitsui/examples/demo/Standard_Editors/VideoEditor_demo.py
jwiggins: Why is this?

traitsui/qt4/video_editor.py
jwiggins: I'm not sure what you're asking. `VideoSurface` takes a `widget` argument, which is equal to `self.control` in this method.

traitsui/qt4/video_editor.py
jwiggins: This doesn't look like a simple transformation since the handlers are being used to initialized the Qt objects in `init()`. Unlikely to happen in this PR.

traitsui/qt4/video_editor.py
jwiggins: Seems to be a common pattern in TraitsUI Qt code...

traitsui/tests/editors/test_video_editor.py
jwiggins: It doesn't appear to be used anywhere else in TraitsUI. Perhaps this should be its own issue?

traitsui/examples/demo/Standard_Editors/VideoEditor_demo.py
rahulporuri: just because we're trying to not use `on_trait_change` and IIUC that's how the `_*_changed` methods are hooked up internally.

(Looks like `_*_changed` methods use `HasTraits._on_trait_change` which gets used by `HasTraits.on_trait_change`)

traitsui/tests/editors/test_video_editor.py
rahulporuri: we already have https://github.com/enthought/traitsui/issues/645. resolving here.

JackDra:
One last comment that is not necessary: 
It may be good to know on what platform, what movie formats are supported (with what extension tag). 

We had an issue in the past where our vp9 formatted files were not playing. An example file:
https://file-examples-com.github.io/uploads/2020/03/file_example_WEBM_480_900KB.webm

It might have been that our file did not actually have an extension `.webm` and QTMediaPlayer determines the format through the extension.

jwiggins:
Good suggestion @JackDra. I'll do that here for now and we can move the info to the appropriate place later. Most of this info was collected by @mdsmarte, but was attached to a private issue.

[Qt Wiki: Multimedia Backends](https://wiki.qt.io/Qt_5.13_Multimedia_Backends)

**Windows:**
Depending on how Qt was built [and possibly environment variables], videos are handled by either [WMF](https://docs.microsoft.com/en-us/windows/win32/medfound/supported-media-formats-in-media-foundation) or [DirectShow](https://docs.microsoft.com/en-us/windows/win32/directshow/supported-formats-in-directshow). If using DirectShow, a user can increase their codec and container-file capabilities by installing the [K-Lite Codec Pack](https://codecguide.com/about_kl.htm). WMF is [currently] the future of Windows media handling, but the Qt5 supplied by EDM only has DirectShow support.

**MacOS:**
If it's [supported](https://developer.apple.com/documentation/coremedia/1564239-video_codec_constants?language=objc) by [AVFoundation](https://developer.apple.com/av-foundation/), you can open it.

**Linux:**
Make sure [GStreamer](https://gstreamer.freedesktop.org/) is installed.

jwiggins:
@rahulporuri If you have no further objections, I'll merge this today.

rahulporuri:
> @rahulporuri If you have no further objections, I'll merge this today.

rahulporuri:
good news : after installing the k-lite basic codec pack, the video shows up on my machine (windows + pyqt5 from edm + py36)

bad news : the video layout is screwed up

well-that-is-unfornatute

rahulporuri:
> We had an issue in the past where our vp9 formatted files were not playing. An example file:
https://file-examples-com.github.io/uploads/2020/03/file_example_WEBM_480_900KB.webm

I'm seeing the same behavior with this file as well - the video is rotated 90 deg anti clockwise and left-right are flipped.

jwiggins:
The rotation is the expected behavior. It's showing that users can manipulate the stream via the `image_func` trait.

rahulporuri:
thanks a lot @jwiggins for pushing this across the finish line. please go ahead and merge.

jwiggins:
Thanks everyone, this is a nice addition to the library.

rahulporuri:
> are you hoping for a new traitsui release soon?

@jwiggins ^ . We are planning on making a release early Feb. We can make one sooner if that's useful.

jwiggins:
Talk to @mdsmarte and/or @ffishburn. Their customer project is an intended user of this feature.

PR #748, opened 2020-02-26
Add make target for Dash/Zeal docset

achabotl:
This PR makes it simple to create a docset from the TraitsUI html documentation, as used by the [Dash](http://kapelli.com/dash) and [Zeal](http://zealdocs.org/) api browsers.

It adds [doc2dash](https://pypi.python.org/pypi/doc2dash) as a documentation dependency to create the docset. All output is kept under `$(BUILDDIR)`.

It also builds the docset automatically when doing `etstools.py docs`.

This is an adaptation of https://github.com/enthought/traits/pull/180 for TraitsUI.


achabotl:
Hasn't seen action in 5 years. Closing.

PR #753, opened 2020-03-02
FEAT: QT VideoEditor pulling frames

JackDra:
This is currently not working, but the same implementation using pyqt5 instead of pyside2 does work.

It may be related to [this bug report](https://bugreports.qt.io/browse/PYSIDE-794?attachmentOrder=desc)

This is the basic idea for inserting a QAbstractVideoSurface between the input from the data source and the output QVideoWidget. This allows a user defined image transformation function to be inserted in before painting to the QVideoWidget.

I'm sure the actual implementation of this would need to be discussed and changed.

### DIFF
--- a/traitsui/qt4/video_editor.py
+++ b/traitsui/qt4/video_editor.py
@@ -133,7 +133,7 @@ class VideoEditor(Editor):
     frame = Property(Array(shape=(None, None, 3), dtype='u1'))
 
     #: does the drawing onto the image plane
-    control = Instance(ImageWidget)
+    control = Instance(QVideoWidget)
 
     #: handels the image pulling so the frames can be processed.
     surface = Instance(QAbstractVideoSurface)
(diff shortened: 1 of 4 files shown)


traitsui/qt4/video_editor.py
```diff
@@ -102,6 +190,21 @@ class VideoEditor(Editor):
...
+        else:
+            return [[[0, 0, 0]]]
+
+    def _get_image_fun(self):
+        return self.control.image_fun
+
+    def _set_image_fun(self, new_fun):
+        self.control.image_fun = new_fun
+
+    def _control_default(self):
```
corranwebster: TraitsUI needs to be very careful about when Qt stuff is created - if something happened to reference the control trait before there were a QApplication then this could cause an error.  Low probability, but don't rely on defaults like this.

traitsui/qt4/video_editor.py
```diff
@@ -114,16 +217,17 @@ def init(self, parent):
         parent : QWidget or None
             The parent widget for this widget.
         """
-
-        self.control = QVideoWidget()
```
corranwebster: I think that the default use case should probably just use `QVideoWidget` particularly if the alternative fails in some Qt APIs.  Something like: if we have an image manipulation function on the factory then instantiate an `ImageWidget` which uses that manipulation function; otherwise create a standard `VideoWidget` as the control.

Possibly need to do something about the surface as well.

traitsui/qt4/video_editor.py
JackDra: Removed default constructor

traitsui/qt4/video_editor.py
JackDra: Should be implemented now.

JackDra:
The current version should be working now. I tried to pull all the image manipulation functions into the demo. 

The decorator `qimage_function` provides a way to change an image manipulation function into a QImage manipulation function (that also bounds the resulting image in the current viewed frame)

There is a weird problem with the frame slider stuttering, but the video plays smooth.

JackDra:
> The current version should be working now. I tried to pull all the image manipulation functions into the demo.

I completely forgot (IMPORTANT), this is now working if I install using toolkit pyqt5. If I do so, i need to use the latest version off pip, since the edm has the same issue as in #745

> > The video streaming seems to work fine both locally and with the demo video link on my mac.
> > The only issue I had was there is a weird background image displayed behind the video, its a black image half a cut off text field at the top of the video pane
> 
> Sorry this problem was fixed when I built with the latest version of pyqt5 instead of pyside2.
> 
> pyqt5 had to be installed from pip tho...... there were some QAudio attribute errors when using the edm version of pyqt5. going off [this](https://stackoverflow.com/questions/52646129/qt-error-convertvolume-is-not-a-member-of-qaudio), there was a comment suggesting
> 
> > make sure you are building with a Qt version >= 5.8

corranwebster:
I'm going to merge this in to the main PR and go from there.  Thanks for the contribution.

PR #776, opened 2020-04-01
Refactored duplicated code

pbrod:
Moved duplicated code in ui_live.py and ui_modal.py to BaseDialog.display_ui method in ui_base.py for the wx backend.

This is similar to what is done for the qt4 backend.

Added check to ensure the wx.TextCtrl is single line before calling "SetHint" method in
traitsui/wx/text_editor.py to avoid "wx._core.wxAssertionError", because hints can only
be set for single line text controls.


corranwebster:
Again, thanks for this.  These are much appreciated.

PR #777, opened 2020-04-02
Updated HDF5_tree_demo.py

pbrod:
Updated HDF5_tree_demo.py by:
 * Refactored code in if __name___ == "__main__" block in HDF5_tree_demo.py into a main function.
 * Added the demo = XXXtrait() so it can be discovered and run by the traitsui.extra.demo.demo function.
 * Replaced open-close statement block with a "with open ..." block.
 * Refactored the _get_sub_arrays function to return both Tables and Arrays if any.
 * Refactored the _get_sub_groups function and made it simpler.
 * Added make_test_datasets function
 * Added 5 small example HDF5 data sets so the demo can be run right out of the box:
    * test_h5pydata.h5
    * test_fixed.h5
    * test_fixed_compressed.h5
    * test_table_dc.h5
    * test_table_no_dc.h5
 * Added HDF5_tree_demo2.py which shows how to use H5PY to instead of PyTables to display contents of several HDF5 files at once similar to HDF5_tree_demo.py.

Other updates
 * Insulated the demo.configure_traits() call inside if __name__ == '__main__' block in demo.py


corranwebster:
Thanks once again.  LGTM.

PR #779, opened 2020-04-03
Promote LEDEditor to a first-class editor

corranwebster:
It now has both Qt and Wx backends, so no reason not to.

### DIFF
--- a/docs/source/api/traitsui.editors.led_editor.rst
+++ b/docs/source/api/traitsui.editors.led_editor.rst
@@ -1,0 +1,7 @@
+traitsui\.editors\.led\_editor module
+=====================================
+
+.. automodule:: traitsui.editors.led_editor
+    :members:
+    :undoc-members:
+    :show-inheritance:
--- a/examples/demo/Extras/LED_display.py
+++ b/examples/demo/Extras/LED_display.py
@@ -11,24 +11,16 @@ This demo illustrates use of the LEDEditor for displaying numeric values
 using a simulated LED display control.
 """
 
-from threading  import Thread
-
+from threading import Thread
 from time import sleep
 
 from traits.api import HasTraits, Instance, Int, Bool, Float
 
-from traitsui.api import View, Item, HGroup, Handler, UIInfo, spring
-
-from traits.etsconfig.api import ETSConfig
-if ETSConfig.toolkit == 'wx':
-    from traitsui.wx.extra.led_editor import LEDEditor
-else:
-    from traitsui.qt4.extra.led_editor import LEDEditor
-
-# Handler class for the LEDDemo class view:
+from traitsui.api import View, Item, HGroup, Handler, LEDEditor, UIInfo, spring
 
 
 class LEDDemoHandler(Handler):
+    """ Handler class for the LEDDemo class view. """
 
     # The UIInfo object associated with the UI:
     info = Instance(UIInfo)
@@ -55,10 +47,9 @@ class LEDDemoHandler(Handler):
             sleep(.01)
         self.alive = False
 
-# The main demo class:
-
 
 class LEDDemo(HasTraits):
+    """ The main class for the LED Demo. """
 
     # A counter to display:
     counter1 = Int()
@@ -118,6 +109,7 @@ class LEDDemo(HasTraits):
         buttons=['OK'],
         handler=LEDDemoHandler
     )
+
 
 # Create the demo:
 demo = LEDDemo()
--- a/traitsui/wx/led_editor.py
+++ b/traitsui/wx/led_editor.py
@@ -1,6 +1,4 @@
-# -------------------------------------------------------------------------
-#
-#  Copyright (c) 2007, Enthought, Inc.
+#  Copyright (c) 2007-2020, Enthought, Inc.
 #  All rights reserved.
 #
 #  This software is provided without warranty under the terms of the BSD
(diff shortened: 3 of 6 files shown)


mdickinson:
I notice that #778 removes the documentation for `LEDEditor`. Is that intentional, and should it be being readded here?

mdickinson:
> I notice that #778 removes the documentation for `LEDEditor`

Rats. No it doesn't. It just moves it. Sorry for the noise.

mdickinson:
In the demo, the fields are marked "left aligned", "right aligned", etc., but I don't see any difference in the alignments. Should I?

Screenshot 2020-04-03 at 11 44 13

mdickinson:
Ah, I ran the demo before looking at the source, which says that alignment is wx-only.

traitsui/editors/led_editor.py
```diff
@@ -0,0 +1,20 @@
+# (C) Copyright 2020 Enthought, Inc., Austin, TX
+# All rights reserved.
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only
+# under the conditions described in the aforementioned license.  The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+# Thanks for using Enthought open source!
+
+""" A Traits UI editor that wraps a LED-style integer display.
```
mdickinson, line 9: typo: "a" -> "an"

traitsui/wx/led_editor.py
```diff
@@ -59,18 +53,12 @@ def update_editor(self):
...
-
-    #: The editor class to be created:
-    klass = _LEDEditor
 
-    #: The alignment of the numeric text within the control:
-    alignment = Enum("right", "left", "center")
+# editor names for factory to find
+ReadonlyEditor = LEDEditor
+SimpleEditor = LEDEditor
+CustomEditor = LEDEditor
```
kitchoi, line 64: Do we actually need to populate all the three styles?  If the custom/readonly editors are not implemented, then in the future we will be certain that they are never used because attempts to set `style='custom'` or `style='readonly'` will be met with an error (I think). If we populate them, it will add constraint to future extensions when we do want the custom / readonly style to behave completely differently.

e.g. The TableEditor does not have a custom editor. ButtonEditor does not have a readonly editor. They don't seem to be missed.

examples/demo/Extras/LED_display.py
```diff
@@ -11,24 +11,16 @@
...
 
 from traits.api import HasTraits, Instance, Int, Bool, Float
 
-from traitsui.api import View, Item, HGroup, Handler, UIInfo, spring
-
-from traits.etsconfig.api import ETSConfig
-if ETSConfig.toolkit == 'wx':
-    from traitsui.wx.extra.led_editor import LEDEditor
-else:
-    from traitsui.qt4.extra.led_editor import LEDEditor
```
kitchoi, line 26: Shall we keep `traitsui.qt4.extra.led_editor` and `traitsui.wx.extra.led_editor` as aliases? It costs almost nothing to maintain aliases. Otherwise this would be a breaking change.

docs/source/api/traitsui.editors.led_editor.rst
```diff
@@ -0,0 +1,7 @@
+traitsui\.editors\.led\_editor module
+=====================================
+
+.. automodule:: traitsui.editors.led_editor
+    :members:
+    :undoc-members:
+    :show-inheritance:
```
kitchoi, line 7: The API documentation is now being generated automatically. So this file and `traitsui.editors.rst` are not needed any more.

traitsui/wx/led_editor.py
```diff
@@ -59,18 +53,12 @@ def update_editor(self):
         """
         self.control.SetValue(self.str_value)
 
+    def _alignment_changed(self):
+        if self.control is not None:
+            self.control.SetAlignment(LEDStyles[self.alignment])
```
kitchoi, line 58: Could we make such a change in a separate PR such that this PR is more focused about exposing the existing editor as first-class / built-in / whatever editor?

Furthermore I expect there being very limited use case to mutate the alignment after the editor is created - we can wait for this feature request to arise. The fewer mutables there are, the simpler and more maintainable the code is.

traitsui/wx/led_editor.py
```diff
@@ -41,16 +32,19 @@
...
+    #: The alignment of the numeric text within the control.
+    alignment = Enum("right", "center", "left")
+
     def init(self, parent):
         """ Finishes initializing the editor by creating the underlying toolkit
             widget.
         """
         self.control = LEDNumberCtrl(parent, -1)
-        self.control.SetAlignment(LEDStyles[self.factory.alignment])
+        self.control.SetAlignment(LEDStyles[self.alignment])
```
kitchoi, line 47: Same here, I think we can defer making `alignment` mutable in a separate PR / issue and restore the original code which uses the factory value.

(I bet one probably needs to `sync_value` method like this https://github.com/enthought/traitsui/blob/7841ca7bf2742f7087f1fc9150ace89c57db63dd/traitsui/qt4/list_str_editor.py#L114 so that the changes are handled on the event loop and so that the value on the editor factory is actually transferred to the editor.)

kitchoi: > (I bet one probably needs to sync_value method like this...

Please ignore this half of the comment. The `sync_value` is used for synchronizing a trait on the editor with a trait on the object being edited. But the first half of the comment about defer making `alignment` mutable still stands.

PR #780, opened 2020-04-03
Python 3.8 API change: cgi.escape -> html.escape

christianbrodbeck:
Python 3.8 API change: https://docs.python.org/3/whatsnew/3.8.html#api-and-feature-removals

Since the documentation says the old location is deprecated in Python 3.2 I did hard replace rather than try/except.

### DIFF
--- a/traitsui/wx/ui_panel.py
+++ b/traitsui/wx/ui_panel.py
@@ -23,7 +23,7 @@ import wx
 import wx.html as wh
 import re
 
-from cgi import escape
+from html import escape
 
 from traits.api import Instance, Undefined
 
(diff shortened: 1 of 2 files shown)


traitsui/wx/ui_panel.py
```diff
@@ -57,7 +57,6 @@
 from .constants import screen_dx, screen_dy, WindowColor
 
 from .ui_base import BaseDialog
-from .constants import is_mac
```
christianbrodbeck, line 60: This was unused – if it was left there intentionally let me know and I will revert.

corranwebster:
LGTM.  Thank you!

PR #783, opened 2020-04-16
TraitsUI 7.0 Release

corranwebster:
Release TraitsUI 7.0.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -4,7 +4,7 @@ Traits UI Changelog
 Release 7.0.0
 -------------
 
-Traits 7.0.0 is a major release which focuses on modernization of the
+TraitsUI 7.0.0 is a major release which focuses on modernization of the
 codebase.  This release includes updating the WxPython backend to support
 WxPython 4, an effort led by Rob McMullen's first cut at support, with
 help from a number of other people, particularly Per Brodtkorb.  It also
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,20 +1,64 @@
 Traits UI Changelog
 ===================
 
-Release 6.2.0
+Release 7.0.0
 -------------
 
+Traits 7.0.0 is a major release which focuses on modernization of the
+codebase.  This release includes updating the WxPython backend to support
+WxPython 4, an effort led by Rob McMullen's first cut at support, with
+help from a number of other people, particularly Per Brodtkorb.  It also
+updates the code to use new features of Traits 6, and just like Traits 6,
+this release drops support for Python 2.
 
+In addition to the modernization, this code fixes many bugs and adds some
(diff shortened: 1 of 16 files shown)


kitchoi:
Changes look good to me.
I wonder if we want to merge the changes for `Union` -> `Either` back to master.
We can cherry-pick commits (e.g. changelog and setup.py) to be merged back to master in a separate PR and we can release 7.0 from the release branch.

setup.py
```diff
@@ -8,8 +8,8 @@
 from setuptools import setup, find_packages
 from io import open
 
-MAJOR = 6
-MINOR = 2
+MAJOR = 7
+MINOR = 0
 MICRO = 0
 
 IS_RELEASED = False
```
kitchoi, line 15: I don't know the release mechanism of traitsui all that well. Is this flag supposed to be changed to True later?

kitchoi: Hmmm this branch is targeting master so it makes sense that flag is not changed in the diff... Resolving.

corranwebster: We'll toggle this on the actual release commit, then bump the version to 7.1.0 unreleased as a second commit, then merge.

kitchoi: Yep. Make sense. Thanks!

CHANGES.txt
```diff
@@ -1,20 +1,64 @@
 Traits UI Changelog
 ===================
 
-Release 6.2.0
+Release 7.0.0
 -------------
 
+Traits 7.0.0 is a major release which focuses on modernization of the
```
kitchoi, line 7: Oops, how did I overlook this in the first pass. 🤦 
This should be "TraitsUI 7.0.0", not "Traits 7.0.0"

corranwebster: Arg.  This is annoying, but fixable.

ci-src-requirements.txt
```diff
@@ -1,2 +1 @@
-git+http://github.com/enthought/traits.git#egg=traits
-git+http://github.com/enthought/pyface.git#egg=pyface
+git+http://github.com/enthought/pyface.git@release/7.0#egg=pyface
```
kitchoi, line 1: pyface's master has the `Union` removed, so this can be changed back to master if needed. Maybe in a separate PR so the CI for this release PR is targeting pyface 7.0?

corranwebster: Yes - I am planning a separate PR after this is merged.  There is already one for Pyface.

CHANGES.txt
corranwebster: Right, I have fixed this in the documentation PR and the Github release.  The only place where it will remain is in the sdist uploaded to PyPI.  I think we can live with that.

PR #784, opened 2020-04-16
Convert Union back to Either as Union which is not released in traits 6.0

kitchoi:
**Targeting release branch for 7.0**

This PR reverts the `Union` back to `Either` as `Union` was not released with Traits 6.

Not sure if we wanted this on master too. We can cherry pick in either direction.

### DIFF
--- a/traitsui/context_value.py
+++ b/traitsui/context_value.py
@@ -62,7 +62,7 @@ instances of :class:`ContextValue` (abbreviated as ``CV``)::
 """
 
 
-from traits.api import HasStrictTraits, Instance, Str, Int, Float, Union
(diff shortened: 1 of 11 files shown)


kitchoi:
Hmmm not all of them were `Either`, some were `Traits`. And the difference is important for some.
Fixing...

kitchoi:
The upstream branch has CI configuration changed so that we are testing against traits 6.0. Which is great. Now the CI is seeing the same problem from pyface.

corranwebster:
Thanks.  I will merge this into 7.0 release branch.  Test failures are expected :(

PR #794, opened 2020-04-27
Rename _trait_values to avoid overshadowing traits machinery

kitchoi:
This PR does a simple renaming of a private attribute to avoid overshadowing an internal method on traits.
cf https://github.com/enthought/traits/issues/705


kitchoi:
Thanks for the review. Since this PR has been open for a while, I will merge master to this to get another CI build before merging...

kitchoi:
Also confirmed that there has not been new references nor usages of `_trait_values` introduced since opening this PR.
CI is green, merging...

PR #796, opened 2020-04-27
[WIP] Deprecate trait_view_for in favor of a new method

kitchoi:
The name of `traitsui.handlers.Handler.trait_view_for` contains "trait" as a prefix and risk overshadowing methods introduced by traits. cf. https://github.com/enthought/traits/issues/705 and the related issues.

This PR proposes one way round this issue: Deprecate the method in favor of a new method using a different name.

I have searched for "trait_view_for" in pyface, envisage, chaco, mayavi and found nothing. It is likely used internally by traitsui only. But to be on the safe side, this PR proposes deprecating it with a method named `get_view_for` that does the same thing. 

`trait_view_for` also searches for method names starting with "trait_view_for_" with a set of magic naming pattern. I can't find evidence of this being used, neither do I think this is a good idea, so it might be safe to deprecate that too.

This PR would need CI tests for wx InstanceEditor in order to go in, but I'd like to get feedback on this first before going forward with this.

### DIFF
--- a/traitsui/handler.py
+++ b/traitsui/handler.py
@@ -25,6 +25,7 @@
 
 # avoid deprecation warning
 from inspect import getfullargspec
+import warnings
 
 from traits.api import HasPrivateTraits, HasTraits, Instance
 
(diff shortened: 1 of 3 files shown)


corranwebster:
If this change were to take place, I think I would prefer `traitsui_view_for` rather than `get_view_for`.  In general, the idea would be that where TraitsUI uses "trait(s)_foo" we migrate towards "traitsui_foo" for the same reasons as in Traits.

kitchoi:
I started with `traitsui_view_for` and then `traits` reminds me that it also has a prefix of `"trait"` 😢 
Maybe `traits` needs to complain about `trait_`, `traits_`, not just just `trait`.

kitchoi:
Closing for now as being stale. This PR is now tracked in GH so it can be revived when needed.

PR #797, opened 2020-04-27
[WIP] Deprecate trait_handler in favor of handler_trait

kitchoi:
Similar to the motivation of #796, this PR proposes deprecating `trait_handler` in favor of an attribute with a different name that does not start with "trait".

The `trait_handler` is a trait that can be defined when a `ListEditor` is defined.  I have searched envisage, mayavi and chaco and I can't find evidence of its being redefined.

The name of `trait_handler` is so generic that I can imagine it being a method defined by traits, and traitsui would be overshadowing the method. Such a method does not exist now, but it might be introduced in the future.

This PR is opened for discussion. I haven't assessed how well tested the affected code is, hence the WIP status.

### DIFF
--- a/traitsui/editors/list_editor.py
+++ b/traitsui/editors/list_editor.py
@@ -17,7 +17,7 @@
 """ Defines the list editor factory for the traits user interface toolkits..
 """
 
-
+import warnings
 
 from traits.api import (
     HasTraits,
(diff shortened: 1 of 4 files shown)


kitchoi:
It seems that the list editor is not covered in CI, at all.
While this PR shows the full deprecation, the less risky approach for the short-term may be to simply add the filterwarning, as is also done here.

kitchoi:
Closing as there are currently no such discussions being taken place. At least this PR is now tracked GH and can be revived when needed.

PR #815, opened 2020-05-11
TST: Remove PySide < 2 and PyQt4 from CI

kitchoi:
Following https://github.com/enthought/pyface/pull/512, traitsui can no longer test again PySide < 2 and PyQt4. New builds on CI currently fail.

This PR removes PySide <2 and PyQt4 from the CI configuration in order to be consistent with Pyface. They are also removed from extra_requires as `pyface.qt` no longer supports these backends.


corranwebster:
LGTM

PR #836, opened 2020-05-22
Add more tests for Enum Editor

ievacerny:
Adds more tests for EnumEditor.

### DIFF
--- a/traitsui/tests/editors/test_enum_editor.py
+++ b/traitsui/tests/editors/test_enum_editor.py
@@ -97,7 +97,10 @@ def set_combobox_index(combobox, idx):
 def finish_combobox_text_entry(combobox):
     """ Finish text entry given combobox. """
     if is_current_backend_wx():
-        raise unittest.SkipTest("Test not implemented for wx")
+        import wx
+
+        event = wx.CommandEvent(wx.EVT_TEXT_ENTER.typeId, combobox.GetId())
+        wx.PostEvent(combobox, event)
 
     elif is_current_backend_qt4():
         combobox.lineEdit().editingFinished.emit()
@@ -106,7 +109,12 @@ def finish_combobox_text_entry(combobox):
 def get_all_button_status(widget):
     """ Gets status of all radio buttons in the layout of a given widget."""
     if is_current_backend_wx():
-        raise unittest.SkipTest("Test not implemented for wx")
+        button_status = []
+        for item in widget.GetSizer().GetChildren():
+            button = item.GetWindow()
+            if button.value != "":  # There are empty invisible buttons
+                button_status.append(button.GetValue())
+        return button_status
 
     elif is_current_backend_qt4():
         layout = widget.layout()
@@ -119,7 +127,13 @@ def get_all_button_status(widget):
 def click_radio_button(widget, button_idx):
     """ Simulate a radio button click given widget and button number."""
     if is_current_backend_wx():
-        raise unittest.SkipTest("Test not implemented for wx")
+        import wx
+
+        sizer_items = widget.GetSizer().GetChildren()
+        button = sizer_items[button_idx].GetWindow()
+        event = wx.CommandEvent(wx.EVT_RADIOBUTTON.typeId, button.GetId())
+        event.SetEventObject(button)
+        wx.PostEvent(widget, event)
 
     elif is_current_backend_qt4():
         widget.layout().itemAt(button_idx).widget().click()
@@ -128,7 +142,8 @@ def click_radio_button(widget, button_idx):
 def get_list_widget_text(list_widget):
     """ Return the text of currently selected item in given list widget. """
     if is_current_backend_wx():
-        raise unittest.SkipTest("Test not implemented for wx")
+        selected_item_idx = list_widget.GetSelection()
+        return list_widget.GetString(selected_item_idx)
 
     elif is_current_backend_qt4():
         return list_widget.currentItem().text()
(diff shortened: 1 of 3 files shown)


traitsui/tests/editors/test_enum_editor.py
```diff
@@ -101,18 +94,237 @@ def set_combobox_index(combobox, idx):
...
+
+        list_widget.SetSelection(idx)
+        event = wx.CommandEvent(wx.EVT_LISTBOX.typeId, list_widget.GetId())
+        wx.PostEvent(list_widget, event)
+
+    elif is_current_backend_qt4():
+        list_widget.setCurrentRow(idx)
+
+
+class TestEnumEditorMapping(unittest.TestCase):
```
kitchoi, line 165: Since all the tests are decorated with `skip_if_not_null`, we could also decorate the whole test case instead of decorating individual test methods.

traitsui/tests/editors/test_enum_editor.py
```diff
@@ -101,18 +94,237 @@ def set_combobox_index(combobox, idx):
...
+                "value",
+                editor=enum_editor_factory,
+                style=style,
+            )
+        )
+
+        with store_exceptions_on_all_threads():
+            editor = self.setup_ui(IntEnumModel(), formatted_view)
+
+            with self.assertRaises(AssertionError):  # FIXME issue #782
```
kitchoi, line 192: Nit: `<organization>/<repository>#<issue_number>` is preferred in case the code is moved/copied to another repository (it is rare, but has happened). Even if the repository is renamed, I think GitHub provides redirections.

traitsui/tests/editors/test_enum_editor.py
```diff
@@ -180,46 +374,199 @@ def check_enum_text_bad_update(self, view):
...
+    def setup_gui(self, model, view):
+        gui = GUI()
+        ui = model.edit_traits(view=view)
+        self.addCleanup(ui.dispose)
+
+        gui.process_events()
+        editor = ui.get_editors("value")[0]
+        widget = editor.control
+
+        return gui, widget
```
kitchoi, line 460: Just an observation: The `create_ui` context manager being introduced in https://github.com/enthought/traitsui/pull/664/files could be useful here.  I will try to review that PR soon.

ievacerny: I'm not sure it would be that useful here because GUI for wx tests would still have to be set up and cleaned up manually.

traitsui/tests/editors/test_enum_editor.py
```diff
@@ -101,18 +94,237 @@ def set_combobox_index(combobox, idx):
...
+    """ Finish text entry given combobox. """
+    if is_current_backend_wx():
+        import wx
+
+        event = wx.CommandEvent(wx.EVT_TEXT_ENTER.typeId, combobox.GetId())
+        wx.PostEvent(combobox, event)
+
+    elif is_current_backend_qt4():
+        combobox.lineEdit().editingFinished.emit()
+
```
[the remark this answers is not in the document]
ievacerny: Will add an `else` branch 👍  Although I don't see why tests for existing toolkits would start failing even if `else` branch isn't added -- am I missing something?

As for the new pattern, to me adding an `elif` branch to a helper method seems easier than going through this framework -- there is only one docstring and one signature to maintain and adding 
```
elif new_toolkit:
    implementation
```
seems easier than
```
@function.register("new_toolkit")
def function(signature):
    implementation
```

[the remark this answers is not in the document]
kitchoi: > Although I don't see why tests for existing toolkits would start failing even if else branch isn't added -- am I missing something?

Sorry for my weird way of putting things. 

If the else branch isn't added, and if we have a new toolkit, when we run the test suite with this new toolkit, the tests here will not be skipped (because the toolkit isn't null). Since `is_current_backend_wx` and `is_current_backend_qt4` will return false, the tests will likely fail because the helper functions are no-op in that new toolkit.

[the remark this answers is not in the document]
kitchoi: > there is only one docstring and one signature to maintain and adding

I would probably only maintain the docstring once only like this:

```
@toolkit_dispatch
def get_something(widget):
    """ Docstring here """

@get_something("wx")
def _(widget):
    return widget.GetName()
```
This is similar to the `functools.singledispatch`: It is a functional way to achieve polymorphism.

traitsui/tests/editors/test_enum_editor.py
kitchoi: Yes, we are missing the GUITestAssistant for Wx too: https://github.com/enthought/pyface/pull/266

In any case, what I meant to say is that this is a pattern repeated so many times that it needs to be refactored. I was also meaning to reference an implementation where this use case is solved using a context manager.

ievacerny: Yes, I agree about refactoring but I don't think there's a need to rush the review on https://github.com/enthought/traitsui/pull/664 because it wouldn't help here until the same exists for wx. And I don't think that's going to happen soon enough to make it into this PR.

traitsui/tests/editors/test_enum_editor.py
kitchoi: Ah just realized my comment wrongly stated `skip_if_not_null`. `skip_if_null` is the correct one. Thanks!

traitsui/wx/ui_panel.py
```diff
@@ -1011,7 +1011,7 @@ def add_items(self, content, panel, sizer):
             item_sizer.Add(
                 control,
                 growable,
-                flags | layout_style | wx.ALIGN_CENTER_VERTICAL,
```
kitchoi, line 1014: I think this is now fixed on master. Could you merge master to this branch please?

PR #838, opened 2020-05-22
Add set editor tests

ievacerny:
Adds tests for set editor.

### DIFF
--- a/traitsui/tests/editors/test_set_editor.py
+++ b/traitsui/tests/editors/test_set_editor.py
@@ -5,6 +5,8 @@ from pyface.gui import GUI
 from traits.api import HasTraits, List
 from traitsui.api import SetEditor, UItem, View
 from traitsui.tests._tools import (
+    click_button,
+    is_control_enabled,
     is_current_backend_qt4,
     is_current_backend_wx,
     skip_if_null,
@@ -32,7 +34,7 @@ def get_view(can_move_all=True, ordered=False):
 
 
 def get_list_items(list_widget):
-    """ Return a list of strings. """
+    """ Return a list of strings from the list widget. """
     items = []
 
     if is_current_backend_wx():
@@ -43,11 +45,20 @@ def get_list_items(list_widget):
         for i in range(list_widget.count()):
             items.append(list_widget.item(i).text())
 
+    else:
+        raise unittest.SkipTest("Test not implemented for this toolkit")
+
     return items
 
 
-def click_on_item(editor, item_no, in_used=False):
-    """ Simulate a click on an item in a specified list. """
+def click_on_item(editor, item_idx, in_used=False):
+    """ Simulate a click on an item in a specified list.
+
+    The function deselects all items in both used and unused lists, then
+    selects an item at index item_idx either in the used list (if
+    in_used=True) or in the unused list. Finally the function simulates a
+    click on the selected item.
+    """
     unused_list = editor._unused
     used_list = editor._used
 
@@ -61,7 +72,7 @@ def click_on_item(editor, item_no, in_used=False):
             used_list.Deselect(i)
         # Select the item in the correct list
         list_with_selection = used_list if in_used else unused_list
-        list_with_selection.SetSelection(item_no)
+        list_with_selection.SetSelection(item_idx)
 
         event = wx.CommandEvent(
             wx.EVT_LISTBOX.typeId, list_with_selection.GetId()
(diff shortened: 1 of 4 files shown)


traitsui/tests/editors/test_set_editor.py
```diff
@@ -0,0 +1,592 @@
...
+    """ Return if the button is enabled or not."""
+    if is_current_backend_wx():
+        return button.IsEnabled()
+
+    elif is_current_backend_qt4():
+        return button.isEnabled()
+
+
+def click_button(button):
+    """ Click the provided button. """
```
kitchoi, line 133: This function is general enough that it will be useful for other tests too.
I think there are at least two similar instances in the code base already. Could this be contributed to somewhere more reusable e.g. `traitsui.tests._tools`?

traitsui/tests/editors/test_set_editor.py
```diff
@@ -0,0 +1,592 @@
...
+            used_list.item(i).setSelected(status)
+
+        if in_used:
+            used_list.itemDoubleClicked.emit(used_list.item(item_no))
+        else:
+            unused_list.itemDoubleClicked.emit(unused_list.item(item_no))
+
+
+def get_button_enabled(button):
+    """ Return if the button is enabled or not."""
```
kitchoi, line 124: This is the same as `is_enabled` from `traitsui.tests.test_labels`, which has the missing `else` block. What do you think about moving the one from `test_labels` to somewhere more reusable and both test modules can import from there?

traitsui/tests/editors/test_set_editor.py
```diff
@@ -0,0 +1,592 @@
...
+            if is_current_backend_wx():  # FIXME issue #840
+                with self.assertRaises(AssertionError):
+                    self.assertEqual(get_list_items(editor._used), ["two"])
+                self.assertEqual(get_list_items(editor._used), ["one", "two"])
+            else:
+                self.assertEqual(get_list_items(editor._used), ["two"])
+            self.assertEqual(list_edit.value, ["two"])
+
+    @skip_if_null
+    def test_simple_set_editor_use_ordered_selected(self):
```
kitchoi, line 549: This test looks almost identical to `test_simple_set_editor_use_button`.
It seems that the only difference is the last assertion: where the moved item is selected again if `ordered` is true. Did I understand correctly?

ievacerny: The setup is a also bit different because of `ordered` flag. But yes, apart from that the tests are identical, although I don't think they should necessarily be. If the editor has `ordered=False`, `update_editor` sorts the values alphabetically, but it isn't called on button clicks right now.

traitsui/tests/editors/test_set_editor.py
```diff
@@ -0,0 +1,592 @@
...
+
+    if is_current_backend_wx():
+        for i in range(list_widget.GetCount()):
+            items.append(list_widget.GetString(i))
+
+    elif is_current_backend_qt4():
+        for i in range(list_widget.count()):
+            items.append(list_widget.item(i).text())
+
+    return items
```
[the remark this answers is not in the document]
ievacerny: In case of another backend, the test should be skipped. I separated out items list because that was common code between both toolkits.

traitsui/tests/editors/test_set_editor.py
kitchoi: Hmmm maybe I am still confused what `ordered` is designed to do.

It looks like the most important effect of the `ordered=True` flag is:
- Show the list of values using the order as is given on the trait
- Show the move-up and move-down arrows for manually reordering values on the trait.

This is tied to whether the values being edited (in this case: `ListModel.value`) is ordered or not. Currently `ListModel.value` is `["one", "two"]`, which happens to be alphabetically ordered.

If the `ListModel.value` is changed to `["two", "one"]` with `ordered=True`, `update_editor` will be called but it won't try to alphabetically sort the values, then we expect the view to show `["two", "one"]`, as opposed to `["one", "two"]` (alphabetically ordered).

Could we add this check or does it exist here already? It could be done by instantiating `ListModel(value=["two", "one"])` instead of `ListModel()` and check the used list is `["two", "one"]`.

(The fact that the selection is clear when `ordered=False` looks to me a minor UX optimization, but I am only guessing on this point.)

ievacerny: I expanded both tests to show the sorting behaviour

kitchoi: Thank you!

PR #845, opened 2020-05-27
Add image enum editor tests

ievacerny:
Adds tests for image enum editor.

### DIFF
--- a/traitsui/tests/editors/test_image_enum_editor.py
+++ b/traitsui/tests/editors/test_image_enum_editor.py
@@ -288,9 +288,6 @@ class TestSimpleImageEnumEditor(unittest.TestCase):
         editor = ui.get_editors("value")[0]
         control = editor.control
 
-        if is_current_backend_qt4():
-            self.addCleanup(control.clear)
-
         return gui, control
 
     def test_simple_editor_more_cols(self):
@@ -356,10 +353,10 @@ class TestSimpleImageEnumEditor(unittest.TestCase):
 
             # Smoke test for ImageEnumItemDelegate painting
             combobox.showPopup()
+            combobox.hidePopup()
             gui.process_events()
 
             combobox.setCurrentIndex(1)
-            combobox.hidePopup()
             gui.process_events()
 
             self.assertEqual(enum_edit.value, 'top right')
--- a/traitsui/tests/editors/test_image_enum_editor.py
+++ b/traitsui/tests/editors/test_image_enum_editor.py
@@ -351,7 +351,12 @@ class TestSimpleImageEnumEditor(unittest.TestCase):
 
             self.assertEqual(enum_edit.value, 'top left')
 
+            # Smoke test for ImageEnumItemDelegate painting
+            combobox.showPopup()
+            gui.process_events()
+
             combobox.setCurrentIndex(1)
+            combobox.hidePopup()
             gui.process_events()
 
             self.assertEqual(enum_edit.value, 'top right')
--- a/traitsui/tests/editors/test_image_enum_editor.py
+++ b/traitsui/tests/editors/test_image_enum_editor.py
@@ -318,62 +318,48 @@ class TestSimpleImageEnumEditor(unittest.TestCase):
         enum_edit = EnumModel()
 
         with store_exceptions_on_all_threads():
-            with patch(cache_to_patch, wraps=image_cache) as patched_cache:
-                gui, control = self.setup_gui(enum_edit, get_view("simple"))
+            gui, control = self.setup_gui(enum_edit, get_view("simple"))
 
-                self.assertEqual(enum_edit.value, 'top left')
-                self.assertEqual(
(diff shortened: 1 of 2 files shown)


traitsui/tests/editors/test_image_enum_editor.py
```diff
@@ -0,0 +1,503 @@
...
+                    path=os.getcwd(),
+                    cols=4,
+                ),
+                style="custom",
+            ),
+            resizable=True,
+        )
+
+        with store_exceptions_on_all_threads():
+            gui, combobox = self.setup_gui(enum_edit, view)
```
kitchoi, line 420: It is good to have this test.
🤖 ❄️ :  `gui`  and `combobox` are not used.

traitsui/tests/editors/test_image_enum_editor.py
```diff
@@ -0,0 +1,501 @@
...
+    def test_simple_editor_combobox(self):
+        enum_edit = EnumModel()
+
+        with store_exceptions_on_all_threads():
+            with patch(cache_to_patch, wraps=image_cache) as patched_cache:
+                gui, combobox = self.setup_gui(enum_edit, get_view("simple"))
+
+                self.assertEqual(enum_edit.value, 'top left')
+                self.assertEqual(
+                    patched_cache.call_args[0][0], "@icons:top left_origin"
```
kitchoi, line 366: Yesterday I ran these tests with a Linux machine of mine with PyQt5 and a few of these tests checking the last call argument on the mock were consistently failing ("@icon:bottom right_origin" != "@icons: top left_origin". On close inspection, all the images were fetched from the cache, but the painted one is not necessarily the first one, the GUI itself was behaving fine. I haven't seen the same failures on Travis/Appveyor or my own OSX machine.

I understand this is an indirect test on what image ended up being used, and there is otherwise no way to check for this. Unfortunately this assertion might just be a bit too indirect and inconclusive, and is depending too strongly on implementation details that may vary from platforms to platforms, from one Qt build to another, causing the test to be unreliable. I think we shall accept that there is no reasonable way to inspect the image being used as of now and remove these checks.  The assertion on `enum_edit.value` is good and can be kept.

ievacerny: I removed the checks from `TestSimpleImageEnumEditor` but not from `TestReadOnlyImageEnumEditor`. For readonly editors that is the only reasonable way to test them, and even though those assertions rely on implementation details, they shouldn't fail if `update_editor` is called as expected.

traitsui/tests/editors/test_image_enum_editor.py
```diff
@@ -0,0 +1,488 @@
...
+            click_on_image(get_button_control(image_grid_control, 1))
+            gui.process_events()
+
+            self.assertEqual(enum_edit.value, 'top right')
+
+            # Check that dialog window is closed
+            self.assertEqual(list(control.GetChildren()), [])
+
+    @skip_if_not_qt4
+    @unittest.skipIf(is_linux, "Issue enthought/traitsui#854")
```
kitchoi, line 350: Maybe mention "Possible test interaction on Linux"?  This would indicate that the test should pass while run on its own. One is more likely to try running this test if they see the reason from their console without having to dive into the issue.

traitsui/tests/editors/test_image_enum_editor.py
```diff
@@ -0,0 +1,488 @@
...
+        ui = model.edit_traits(view=view)
+        self.addCleanup(ui.dispose)
+
+        gui.process_events()
+        editor = ui.get_editors("value")[0]
+        control = editor.control
+
+        return gui, control
+
+    @unittest.skipIf(is_linux, "Issue enthought/traitsui#854")
```
kitchoi, line 296: #854 only applies to Qt, or did I misunderstand? So this should `skipIf(is_linux and is_backend_qt4)`?

traitsui/tests/editors/test_image_enum_editor.py
```diff
@@ -0,0 +1,488 @@
...
+            )
+
+            click_on_image(get_button_control(control, 1))
+            gui.process_events()
+
+            self.assertEqual(enum_edit.value, 'top right')
+            self.assertEqual(
+                get_all_button_selected_status(control),
+                [False, True, False, False]
+            )
```
kitchoi, line 430: The assertion on the button status is actually testing Qt/Wx, rather than testing Traitsui editor, because the change is driven by clicking the widget.
The assertion would be meaningful if the change is caused by, e.g. an assignment on `enum_edit.value`, like the test below. Could you remove the button status assertion in this test please?

traitsui/tests/editors/test_image_enum_editor.py
kitchoi: Given the readonly editor only shows one image: the selected one, that makes sense. Thank you.

PR #829, opened 2020-05-19
Remove align centre which was causing problems

corranwebster:
Fixes #828.

It is possible we may need to do some further tweaking of Wx layout with this being removed, but this fixes the immediately reported problem.

PR #846, opened 2020-05-27
TST: Add tests for layout and labels using any non-null toolkits

kitchoi:
This PR adds a few simple tests to ensure that a UI can be created with various layout settings. In particular, this aims at increasing test coverage on `traitsui.wx.ui_panel` (changed from 46% to 50%...)

- Add `create_ui` context manager to `traitsui.tests._tools`.  We have seen this pattern again and again in the test suites. It is probably time to refactor this. Note that this context manager has deliberately left out the `EventLoopHelper.delete_widget` context manager. Otherwise we ~could~ would have to rely on toolkit-specific implementations of `EventLoopHelper`, which has not been implemented yet for wx.
- Drive-by changes to ensure existing tests call `ui.dispose` at the end of the test.

This was motivated by #829, e.g. the `TestOrientation.test_horizontal_layout` would fail before the fix provided by #829

### DIFF
--- a/traitsui/tests/_tools.py
+++ b/traitsui/tests/_tools.py
@@ -131,6 +131,30 @@ def filter_tests(test_suite, exclusion_pattern):
     return filtered_test_suite
 
 
+@contextmanager
+def create_ui(object, ui_kwargs=None):
+    """ Context manager for creating a UI and then dispose it when exiting
+    the context.
+
+    Parameters
+    ----------
+    object : HasTraits
+        An object from which ``edit_traits`` can be called to create a UI
+    ui_kwargs : dict or None
+        Keyword arguments to be provided to ``edit_traits``.
+
+    Yields
+    ------
+    ui: UI
+    """
+    ui_kwargs = {} if ui_kwargs is None else ui_kwargs
+    ui = object.edit_traits(**ui_kwargs)
+    try:
+        yield ui
+    finally:
+        ui.dispose()
+
+
 # ######### Utility tools to test on both qt4 and wx
 
 
--- a/traitsui/tests/test_labels.py
+++ b/traitsui/tests/test_labels.py
@@ -25,6 +25,7 @@ from traitsui.item import Item
 from traitsui.group import VGroup, HGroup
 
 from traitsui.tests._tools import (
+    create_ui,
     is_current_backend_qt4,
     is_current_backend_wx,
     skip_if_not_qt4,
@@ -166,9 +167,8 @@ class TestLabels(unittest.TestCase):
 
         from pyface import qt
 
-        with store_exceptions_on_all_threads():
-            dialog = dialog_class()
-            ui = dialog.edit_traits()
+        with store_exceptions_on_all_threads(), \
+                create_ui(dialog_class()) as ui:
 
             # all labels
             labels = ui.control.findChildren(qt.QtGui.QLabel)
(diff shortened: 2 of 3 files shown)


traitsui/tests/test_layout.py
```diff
@@ -71,10 +80,8 @@ def test_qt_resizable_in_vgroup(self):
 
         from pyface import qt
 
-        with store_exceptions_on_all_threads():
-            dialog = VResizeDialog()
-            ui = dialog.edit_traits()
-
+        with store_exceptions_on_all_threads(), \
+                create_ui(VResizeDialog()) as ui:
             text = ui.control.findChild(qt.QtGui.QLineEdit)
```
corranwebster, line 85: Since you are trying to make this more toolkit-independent, a better way to find the text control is one of either:
- `ui.ui_info.txt.control` (which works in simple cases like this); or
- `ui.get_editors('txt')[0].control`

That would make the only toolkit-dependent part actually getting the width from the widget.

traitsui/tests/_tools.py
```diff
@@ -131,6 +131,30 @@ def filter_tests(test_suite, exclusion_pattern):
...
+        An object from which ``edit_traits`` can be called to create a UI
+    ui_kwargs : dict or None
+        Keyword arguments to be provided to ``edit_traits``.
+
+    Yields
+    ------
+    ui: UI
+    """
+    ui_kwargs = {} if ui_kwargs is None else ui_kwargs
+    ui = object.edit_traits(**ui_kwargs)
```
ievacerny, line 151: There are already a few context manager that do the same + yield an editor (in `test_date_editor.py`, `test_date_range_editor.py`, `test_datetime_editor.py`). I would say that working with an editor is a common enough use case to yield it here together with `ui`, potentially with an option to specify editor by "name" (e.g. "txt" as in tests below). And then the other context managers can be replaced with this one.

traitsui/tests/test_labels.py
```diff
@@ -157,9 +158,8 @@ def _test_qt_labels_right_resizing(self, dialog_class):
 
         from pyface import qt
 
-        with store_exceptions_on_all_threads():
-            dialog = dialog_class()
-            ui = dialog.edit_traits()
+        with store_exceptions_on_all_threads(), \
+                create_ui(dialog_class()) as ui:
```
ievacerny, line 162: There's one more test above (`test_qt_show_labels_right_without_colon`) that creates ui and doesn't dispose of it.

kitchoi:
> Since it's a lot of tests, should that be a separate PR?

Yep. I could not resist and modified a few of the tests in the same test modules modified in this PR. Let's make sure all the GUI in the tests separately.

traitsui/tests/test_labels.py
kitchoi: Well spotted. I will do that here too.

traitsui/tests/_tools.py
kitchoi: Agreed that it is a common pattern to try to retrieve an editor with a name.

On the other hand, getting an editor may fail, mixing it with launching the GUI makes the traceback harder to understand. Sometimes one may also just launch the GUI and close it. The signature of the context manager will then have to allow `None` for the maybe-not-needed editor... `editor, = ui.get_editors("name")` is easy enough to me.  
See also https://github.com/enthought/traitsui/pull/861 for an even easier API for the most common use cases, where you don't need to call `get_editors` at all. :)

Issue #842, opened 2020-05-25
wxAssertionError "GetEventHandler() == this" in individual tests


ievacerny:
The error is raised when trying to destroy children of `TraitsUIPanel` in custom `EnumEditor.rebuild_editor` and `CheckListEditor.rebuild_editor`. No error is raised if `DestroyChildren` is called when there are no children. Example traceback:
```
Traceback (most recent call last):
  File "/Users/icernyte/.edm/envs/traitsui-test-3.6-wx/lib/python3.6/site-packages/traits/trait_notifiers.py", line 578, in _dispatch_change_event
    self.dispatch(handler, *args)
  File "/Users/icernyte/.edm/envs/traitsui-test-3.6-wx/lib/python3.6/site-packages/traits/trait_notifiers.py", line 682, in dispatch
    handler(*args)
  File "/Users/icernyte/Documents/repos/traitsui/traitsui/wx/enum_editor.py", line 130, in _values_changed
    self.rebuild_editor()
  File "/Users/icernyte/Documents/repos/traitsui/traitsui/wx/enum_editor.py", line 368, in rebuild_editor
    panel.DestroyChildren()
wx._core.wxAssertionError: C++ assertion "GetEventHandler() == this" failed at /Users/robind/projects/bb2/dist-osx-py36/build/ext/wxWidgets/src/common/wincmn.cpp(475) in ~wxWindowBase(): any pushed event handlers must have b
een removed
```
It might be related that same error is raised in `DataFrameEditor` tests when attempting to destroy the `TraitsUIPanel` itself:
```
Traceback (most recent call last):
  File "/Users/icernyte/Documents/repos/traitsui/traitsui/wx/toolkit.py", line 434, in destroy_control
    control.Destroy()
  File "/Users/icernyte/Documents/repos/traitsui/traitsui/wx/helper.py", line 358, in Destroy
    super().Destroy()
wx._core.wxAssertionError: C++ assertion "GetEventHandler() == this" failed at /Users/robind/projects/bb2/dist-osx-py36/build/ext/wxWidgets/src/common/wincmn.cpp(475) in ~wxWindowBase(): any pushed event handlers must have b
een removed
```
There might be similar issues with other editors that use `TraitsUIPanel` but most of them don't have any tests or avoid the scenarios that cause issues (e.g. `SetEditor` doesn't call `DestroyChildren`)

ievacerny:
While working on another issue I ended up tracking down why this error is being raised on destruction.

[...]

PR #848, opened 2020-05-28
Move value mapping from factory to individual editors

### LINKED ISSUE #782: format_func not used when EnumEditor is initialized with static values
To reproduce:
```

class Foo(HasTraits):

    true_or_false = Int()

        return View(
            Item(
                "true_or_false",
                editor=EnumEditor(
                    values=[0, 1],
                    format_func=lambda v: str(bool(v)).upper(),
                )
            )
        )

foo = Foo()
foo.configure_traits()
```

One expects to see the "TRUE" and "FALSE" in the dropdown. However one gets "0" and "1" in the view, which is unexpected.
Screenshot 2020-04-14 at 17 10 41

Printing the mapping:
```

ui = foo.edit_traits()
editor, = ui.get_editors("true_or_false")
print(editor.mapping)
ui.dispose()
```
We get:
```
{'0': 0, '1': 1}
```

If I create the editor first and then mutates the `values`, then I get the expected behaviour:
```
        editor = EnumEditor(
            format_func=lambda v: str(bool(v)).upper(),
        )
        # trigger _values_changed to fire
        editor.values = [0, 1] 
        return View(
            Item(
                "true_or_false",
                editor=editor,
            )
        )
```
Screenshot 2020-04-14 at 17 14 13

### LINKED ISSUE #835: wx enum_editor doesn't use format_func when mapping values
When `name` is specified on the editor factory, the mapping is done by the editor, not the factory. That is taken care of in `values_changed` method of the editor in qt4 but not wx.

wx:
https://github.com/enthought/traitsui/blob/e9f6b0cc03bbad20feb1f7e4f86ce2e174685fbb/traitsui/wx/enum_editor.py#L119-L124

qt4:
https://github.com/enthought/traitsui/blob/e9f6b0cc03bbad20feb1f7e4f86ce2e174685fbb/traitsui/qt4/enum_editor.py#L61-L66

Also, wx has its own version of `enum_value_changed` that is almost identical to `traitsui.helper.enum_values_changed` except for the lack of `strfunc` that is used to format values.

ievacerny:
Closes #782 
Closes #835 

Moves value mapping from factory to individual editors. Also fixes wx Enum and Set editor format_func issue by switching to general `enum_values_changed` version.

### DIFF
--- a/traitsui/qt4/enum_editor.py
+++ b/traitsui/qt4/enum_editor.py
@@ -58,17 +58,12 @@ class BaseEditor(Editor):
     #  BaseEditor Interface
     # -------------------------------------------------------------------------
 
-    def values_changed(self, object):
+    def values_changed(self):
         """ Recomputes the cached data based on the underlying enumeration model
             or the values of the factory.
         """
-        if object is self._object:
-            values = self._value()
-        else:
-            values = self.factory.values
-
         self._names, self._mapping, self._inverse_mapping = enum_values_changed(
-            values, self.string_value
+            self._value(), self.string_value
         )
 
     def rebuild_editor(self):
@@ -92,12 +87,13 @@ class BaseEditor(Editor):
             self._object, self._name, self._value = self.parse_extended_name(
                 factory.name
             )
-            self.values_changed(self._object)
+            self.values_changed()
             self._object.on_trait_change(
                 self._values_changed, " " + self._name, dispatch="ui"
             )
         else:
-            self.values_changed(factory)
+            self._value = lambda: self.factory.values
+            self.values_changed()
             factory.on_trait_change(
                 self._values_changed, "values", dispatch="ui"
             )
@@ -139,11 +135,11 @@ class BaseEditor(Editor):
 
     # Trait change handlers --------------------------------------------------
 
-    def _values_changed(self, object, name, new):
+    def _values_changed(self):
         """ Handles the underlying object model's enumeration set or factory's
             values being changed.
         """
-        self.values_changed(object)
+        self.values_changed()
         self.rebuild_editor()
 
 
--- a/traitsui/qt4/enum_editor.py
+++ b/traitsui/qt4/enum_editor.py
@@ -58,11 +58,17 @@ class BaseEditor(Editor):
     #  BaseEditor Interface
     # -------------------------------------------------------------------------
 
-    def values_changed(self):
-        """ Recomputes the cached data based on the underlying enumeration model.
+    def values_changed(self, object):
+        """ Recomputes the cached data based on the underlying enumeration model
+            or the values of the factory.
         """
+        if object is self._object:
+            values = self._value()
+        else:
+            values = self.factory.values
+
         self._names, self._mapping, self._inverse_mapping = enum_values_changed(
-            self._value(), self.string_value
+            values, self.string_value
         )
 
     def rebuild_editor(self):
@@ -86,13 +92,14 @@ class BaseEditor(Editor):
             self._object, self._name, self._value = self.parse_extended_name(
                 factory.name
             )
-            self.values_changed()
+            self.values_changed(self._object)
             self._object.on_trait_change(
                 self._values_changed, " " + self._name, dispatch="ui"
             )
         else:
+            self.values_changed(factory)
             factory.on_trait_change(
-                self.rebuild_editor, "values_modified", dispatch="ui"
+                self._values_changed, "values", dispatch="ui"
             )
 
     def dispose(self):
@@ -104,7 +111,7 @@ class BaseEditor(Editor):
             )
         else:
             self.factory.on_trait_change(
-                self.rebuild_editor, "values_modified", remove=True
+                self._values_changed, "values", remove=True
             )
 
         super(BaseEditor, self).dispose()
@@ -118,33 +125,25 @@ class BaseEditor(Editor):
     def _get_names(self):
         """ Gets the current set of enumeration names.
         """
-        if self._object is None:
-            return self.factory._names
-
         return self._names
 
     def _get_mapping(self):
         """ Gets the current mapping.
         """
-        if self._object is None:
-            return self.factory._mapping
-
         return self._mapping
 
     def _get_inverse_mapping(self):
         """ Gets the current inverse mapping.
         """
-        if self._object is None:
-            return self.factory._inverse_mapping
-
         return self._inverse_mapping
 
     # Trait change handlers --------------------------------------------------
 
-    def _values_changed(self):
-        """ Handles the underlying object model's enumeration set being changed.
+    def _values_changed(self, object, name, new):
+        """ Handles the underlying object model's enumeration set or factory's
+            values being changed.
         """
-        self.values_changed()
+        self.values_changed(object)
         self.rebuild_editor()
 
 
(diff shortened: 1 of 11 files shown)


traitsui/qt4/enum_editor.py
```diff
@@ -86,13 +92,14 @@ def init(self, parent):
             self._object, self._name, self._value = self.parse_extended_name(
                 factory.name
             )
-            self.values_changed()
+            self.values_changed(self._object)
             self._object.on_trait_change(
                 self._values_changed, " " + self._name, dispatch="ui"
             )
         else:
+            self.values_changed(factory)
```
kitchoi, line 100: An alternative would be to define `self._value = lambda : self.factory.values` here, then the original `values_changed`  and `_values_changed` would continue to work. The added branches makes me think harder to ensure that the various combination of code paths are okay. In particular, `self._object` never raises `AttributeError` because this editor is a subclass of `HasPrivateTraits`, one gets `None` for any attribute access with the preceding "underscore" (another misfeature IMO).

traitsui/wx/image_enum_editor.py
```diff
@@ -113,13 +117,24 @@ def init(self, parent):
...
         """ Populates a specified window with a grid of image buttons.
         """
-        # Create the panel to hold the ImageControl buttons:
-        self.control = panel = TraitsUIPanel(parent, -1)
+        # Clear any existing content:
+        panel = self.control
+        panel.SetSizer(None)
+        for child in panel.GetChildren():
+            _popEventHandlers(child)
+        panel.DestroyChildren()
```
kitchoi, line 137: Why is this needed? This is worrying because this could be blindly removing components depended on by other change handlers. This editor should know exactly what content are in the control.

traitsui/wx/image_enum_editor.py
```diff
@@ -113,13 +117,24 @@ def init(self, parent):
         """ Finishes initializing the editor by creating the underlying toolkit
             widget.
         """
-        self._create_image_grid(parent)
+        super(CustomEditor, self).init(parent)
```
kitchoi, line 120: The `parent` is not actually used by `BaseEnumEditor.init`. The base class's `init` does not do what the Editor Interface says it should do: "Finishes initializing the editor by creating the underlying widget".  In fact, all the subclasses rely on the base class's `init` not to create any widget at all.

This `BaseEnumEditor` can be a Mixin that does not subclass `Editor`, but it can still expect subclasses to be instances of `Editor`. The mixin can provide "setup" and "teardown" for subclasses to call in their `init` and `dispose` method. Then `BaseEnumEditor` can be completely free of the GUI toolkit and be refactored out.

This kind of refactoring is probably out-of-scope for this PR. Unless you feel very strongly against this idea or you'd like do the refactoring here, I will open a separate issue for this.

traitsui/wx/image_enum_editor.py
ievacerny: This is needed to remove previous buttons, otherwise they start overlapping. In this case the only children are the buttons that are re-added just below.  The same approach is used in `Enum` and `CheckList` custom editors.

kitchoi: I see. I think I read this with `init` and was surprised.  Could you confirm that this block is only required by `rebuild_editor` only?  If so, it would make more sense if this block is moved to `rebuild_editor` instead.

I could not find the same use of `_popEventHandlers` in the other editors. `_popEventHandlers` is private to `traitsui.wx.toolkit` and it only recognizes a subtype of event handler that is created there. I am still not convinced the connected components are disconnected. Looks like `ImageControl` needs to provide a method for `Unbind`-ing the events so that this container can call it to explicitly disconnect event handlers.

ievacerny: It's not necessary in the `init` - there are no children at that time and the operation essentially does nothing.

The other uses suffer from https://github.com/enthought/traitsui/issues/835 and calling `_popEventHandlers` seems to fix it. I assume `wx` code complains unless all handlers are removed, and since `_popEventHandlers` removes the error, I assume that it removes all attached handlers.

Re privacy of `_popEventHandlers`, it's not necessary to call it directly, there is a toolkit method `destroy_children` that does what is needed here. But to get it I would either have to instantiate `traitsui.wx.toolkit.GUIToolkit` or go through `traitsui.toolkit` to get the already instantiated one. Using `_popEventHandlers` directly seemed most appropriate for the simple use case.

kitchoi: Thank you. Let's move this block to `rebuild_editor` and let's call `GUIToolkit().destroy_children()`. Public interface is always preferred to private ones. If the logic of `destroy_children` needs to be changed, there is only one place to make such changes. We can worry about optimization later.

I still have reservation with this `destroy_children` in these `rebuild_editor`, and probably you do too. 

This is what I arrived at by reading the code, but I haven't tested this:

The `destroy_children` (or `_popEventHandlers`) probably does remove some lingering event handlers that are "pushed" recursively when the the UI is first initialized:
https://github.com/enthought/traitsui/blob/64858b8c9ce9189a8c9cfde60610e4bd12f30962/traitsui/ui.py#L406-L409
This may explain why #842 can be resolved by calling `_popEventHandlers`.

For these editors with `rebuild_editors, if `values` are modified many times, the first call to `_popEventHandlers` remove handlers that are not put back by `_create_image_grid `. In other words, the N-th (N>1) call to the `_popEventHandlers`  will find no event handlers to be removed.

`ImageControl` still binds event handlers and provides no way to unbind them.  However the error in #842 suggests that it only cares about the event handlers added by "pushing", so the error may not be relevant when the `values` are modified a second time.  I guess we can test that quite easily?

(Though we should still make sure what put up is taken down even if it does not cause errors now... that is perhaps another issue for `ImageControl` to provide a dispose method.)

traitsui/wx/image_enum_editor.py
ievacerny: Sounds good to me and I agree that it's best done separately

traitsui/wx/image_enum_editor.py
ievacerny: Moved `delete_children` to `rebuild_editor`. I don't think it's a problem that `_popEventHandlers` won't find any handlers to remove after the first time, it will be a simple no-op.

I opened #856 to add cleanup to `ImageControl`

traitsui/qt4/enum_editor.py
ievacerny: You're right, that is a much nicer approach. Thanks for the suggestion!

traitsui/wx/image_enum_editor.py
kitchoi: Opened https://github.com/enthought/traitsui/issues/857

traitsui/wx/image_enum_editor.py
```diff
@@ -113,13 +117,25 @@ def init(self, parent):
...
 
-    def _create_image_grid(self, parent):
+        # Create the panel to hold the ImageControl buttons:
+        self.control = TraitsUIPanel(parent, -1)
+        self._create_image_grid()
+
+    def rebuild_editor(self):
+        # Clear any existing content:
+        self.control.SetSizer(None)
+        GUIToolkit("traitsui", "wx", "traitsui.wx").destroy_children(
```
kitchoi, line 129: This is already instantiated in `traitsui.wx` and can be imported from `from traitsui.wx import toolkit`.  I see other modules doing this. I am sorry I did not realize this earlier.

Issue #623, opened 2019-08-15
Issues in demos


sandhya-sago:
- [ ] examples/demo/Standard_Editors/CompoundEditor_demo.py
Gives traceback error when alphabets are clicked (from second click onwards).
Traceback (most recent call last):
File "/Users/sgovindaraju/.edm/envs/traitsui_tutorial/lib/python2.7/site-packages/traitsui/wx/range_editor.py", line 251, in update_object_on_enter self._convert_to_slider(self.value))
File "/Users/sgovindaraju/.edm/envs/traitsui_tutorial/lib/python2.7/site-packages/traitsui/wx/range_editor.py", line 297, in _convert_to_slider
   ivalue = int((float(value - self.low) /
TypeError: unsupported operand type(s) for -: 'str' and 'int'

- [ ] examples/demo/Standard_Editors/DirectoryEditor_demo.py 
Traceback when closing or when OK is clicked and does not close.
 Traceback (most recent call last):
   File "/Users/sgovindaraju/.edm/envs/traitsui_tutorial/lib/python2.7/site-packages/traitsui/wx/ui_live.py", line 344, in _on_close_page
     self._on_cancel(event)
   File "/Users/sgovindaraju/.edm/envs/traitsui_tutorial/lib/python2.7/site-packages/traitsui/wx/ui_live.py", line 398, in _on_cancel
     self.close(wx.ID_CANCEL)
   File "/Users/sgovindaraju/.edm/envs/traitsui_tutorial/lib/python2.7/site-packages/traitsui/wx/ui_live.py", line 333, in close
     ui.finish()
   File "/Users/sgovindaraju/.edm/envs/traitsui_tutorial/lib/python2.7/site-packages/traitsui/ui.py", line 283, in finish
     self.reset(destroy=False)
   File "/Users/sgovindaraju/.edm/envs/traitsui_tutorial/lib/python2.7/site-packages/traitsui/ui.py", line 325, in reset
     editor.dispose()
   File "/Users/sgovindaraju/.edm/envs/traitsui_tutorial/lib/python2.7/site-packages/traitsui/wx/file_editor.py", line 125, in dispose
     panel.Bind(wx.EVT_BUTTON, self._button.GetId(), None)
   File "/Users/sgovindaraju/.edm/envs/traitsui_tutorial/lib/python2.7/site-packages/wx/_core.py", line 4224, in Bind
     assert handler is None or callable(handler)
 AssertionError

[...]

PR #853, opened 2020-05-29
TST: Test demo examples can be run

kitchoi:
This PR adds a test case to ensure demo examples can be run to the extent that a GUI can be opened and then closed. Unfortunately some of them have to be skipped, see inlined reasons.

Since `examples` is not a package (though included in source distribution), the tests are added to `integrationtests` folder, which isn't distributed.

A few dependencies are added for testing some of the examples. If that's too much, then we can only skip those examples in the tests (or evaluate if those examples should be removed?).

One downside is that these tests are rather noisy on wx due to DeprecationWarnings. These warnings should be fixed, see https://github.com/enthought/traitsui/issues/579

See #623 for behavioural issues on the demo examples.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -222,7 +222,7 @@ def test(runtime, toolkit, environment):
         environ["EXCLUDE_TESTS"] = "(wx|qt)"
 
     commands = [
-        #"edm run -e {environment} -- coverage run -p -m unittest discover -v traitsui",
+        "edm run -e {environment} -- coverage run -p -m unittest discover -v traitsui",
         "edm run -e {environment} -- coverage run -p -m unittest -v integrationtests.test_all_examples",
     ]
 
--- a/etstool.py
+++ b/etstool.py
@@ -221,9 +221,10 @@ def test(runtime, toolkit, environment):
     else:
         environ["EXCLUDE_TESTS"] = "(wx|qt)"
 
+    parameters["integrationtests"] = os.path.abspath("integrationtests")
     commands = [
         "edm run -e {environment} -- coverage run -p -m unittest discover -v traitsui",
-        "edm run -e {environment} -- coverage run -p -m unittest -v integrationtests.test_all_examples",
+        "edm run -e {environment} -- coverage run -p -m unittest discover -v {integrationtests}",
     ]
 
     # We run in a tempdir to avoid accidentally picking up wrong traitsui
--- a/etstool.py
+++ b/etstool.py
@@ -109,7 +109,7 @@ dependencies = {
     "docutils"
 }
 
-# Additional toolkit-independent dependencies for testing
+# Additional toolkit-independent dependencies for demo testing
 test_dependencies = {
     "apptools",
     "chaco",
--- a/etstool.py
+++ b/etstool.py
@@ -213,7 +213,8 @@ def test(runtime, toolkit, environment):
         environ["EXCLUDE_TESTS"] = "(wx|qt)"
 
     commands = [
-        "edm run -e {environment} -- coverage run -p -m unittest discover -v traitsui"
+        "edm run -e {environment} -- coverage run -p -m unittest discover -v traitsui",
+        "edm run -e {environment} -- coverage run -p -m unittest -v integrationtests.test_all_examples",
     ]
 
     # We run in a tempdir to avoid accidentally picking up wrong traitsui
--- a/etstool.py
+++ b/etstool.py
@@ -109,6 +109,14 @@ dependencies = {
     "docutils"
 }
 
+# Additional toolkit-independent dependencies for testing
+test_dependencies = {
+    "apptools",
+    "chaco",
+    "h5py",
+    "pytables",
+}
+
 extra_dependencies = {
     # XXX once pyside2 is available in EDM, we will want it here
     'pyside2': set(),
@@ -156,6 +164,7 @@ def install(runtime, toolkit, environment, editable):
         dependencies
         | extra_dependencies.get(toolkit, set())
         | runtime_dependencies.get(runtime, set())
+        | test_dependencies
     )
 
     install_traitsui = "edm run -e {environment} -- pip install "
@@ -213,7 +222,7 @@ def test(runtime, toolkit, environment):
         environ["EXCLUDE_TESTS"] = "(wx|qt)"
 
     commands = [
-        "edm run -e {environment} -- coverage run -p -m unittest discover -v traitsui",
+        #"edm run -e {environment} -- coverage run -p -m unittest discover -v traitsui",
         "edm run -e {environment} -- coverage run -p -m unittest -v integrationtests.test_all_examples",
     ]
 
--- a/etstool.py
+++ b/etstool.py
@@ -224,7 +224,10 @@ def test(runtime, toolkit, environment):
     parameters["integrationtests"] = os.path.abspath("integrationtests")
     commands = [
         "edm run -e {environment} -- coverage run -p -m unittest discover -v traitsui",
-        "edm run -e {environment} -- coverage run -p -m unittest discover -v {integrationtests}",
+        # coverage run prevents local images to be loaded for demo examples
+        # which are not defined in Python packages. Run with python directly
+        # instead.
+        "edm run -e {environment} -- python -m unittest discover -v {integrationtests}",
     ]
 
     # We run in a tempdir to avoid accidentally picking up wrong traitsui
--- a/integrationtests/test_all_examples.py
+++ b/integrationtests/test_all_examples.py
@@ -218,7 +218,7 @@ def run_file(file_path):
         content = f.read()
 
     globals = {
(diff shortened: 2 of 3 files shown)


integrationtests/test_all_examples.py
```diff
@@ -0,0 +1,254 @@
...
+    os.path.join(DEMO, "Useful", "demo_group_size.py"),
+)
+skip_file_if(
+    lambda: True, "Require wx and is blocking.",
+    os.path.join(TUTORIALS, "view_multi_object.py"),
+)
+skip_file_if(
+    lambda: True, "Require wx and is blocking.",
+    os.path.join(TUTORIALS, "view_standalone.py"),
+)
```
kitchoi, line 103: These two examples seem to me quite old. Should they be removed? 
Nevertheless I try to make this PR purely about adding tests.

etstool.py
```diff
@@ -109,6 +109,14 @@
...
 }
 
+# Additional toolkit-independent dependencies for demo testing
+test_dependencies = {
+    "apptools",
+    "chaco",
+    "h5py",
+    "pytables",
+}
+
```
kitchoi, line 119: These dependencies aren't traitsui runtime dependencies. On the other hand, I think the demo examples and the demo application could well have been a separate application-like distribution that can depend on other distributions more freely.

integrationtests/test_all_examples.py
kitchoi: If there are no decisions for this, I will remove these two examples in a separate PR, partly because that is orthogonal and partly because then it can be reverted more easily should they need to be resurrected again.

etstool.py
corranwebster: Right.  On my to-do list has been to add a `demo` method to `etstool.py` and add those as dependencies there.  For now it is fine to add as testing dependencies.

integrationtests/test_all_examples.py
corranwebster: A curation needs to be done with the examples, which may involve removing some, but more likely just re-writing and modernizing.

The multi-object view is still valid, it just needs re-working an generalizing to use pyface GUI: the point is to show how to have multiple objects in the context of a view and to access their values).

Similarly the stand-alone example is showing that the view doesn't need to be part of the class.

kitchoi: Thank you! I referred to this comment in #862 so that this can be taken into account.

etstool.py
kitchoi: Thanks!

integrationtests/test_all_examples.py
```diff
@@ -0,0 +1,281 @@
...
+    DEMO,
+    TUTORIALS,
+]
+
+
+#: Mapping from filepath to a callable() -> (skipped: bool, reason: str)
+FILES_MAY_BE_SKIPPED = {}
+
+
+def skip_file_if(condition, reason, filepath):
```
ievacerny, line 53: Maybe I'm missing some intended capabilities, but I find this `skip_file_if` system too complex for what it does. Why not have a simple dictionary with structured entries, that I think would be easier to read, e.g.:
``` python
FILES_MAY_BE_SKIPPED = {
    os.path.join(DEMO, "demo.py"): ( 
        lambda: True,
        "Not a target file to be tested"
    )
}
```
`os.path.abspath` shouldn't be necessary as long as `EXAMPLES` is saved as an absolute path. And I don't think condition execution is needed at import time, it could be done within the subtest by modifying `is_skipped`:
```python
    skip_condition, reason = FILES_MAY_BE_SKIPPED[path]
    return skip_condition(), reason
```

integrationtests/test_all_examples.py
```diff
@@ -0,0 +1,281 @@
...
+    paths : list of str
+        List of Python file paths.
+    """
+    paths = []
+    for root, _, files in os.walk(directory):
+        for filename in files:
+            path = os.path.abspath(os.path.join(root, filename))
+            if is_python_file(path):
+                paths.append(path)
+    return sorted(paths)
```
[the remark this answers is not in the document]
kitchoi: Yes it is. We don't want the test order to be determined by the operating system. In cases of test interactions, it is important for the test order to consistent, deterministic and reproducible.

integrationtests/test_all_examples.py
```diff
@@ -0,0 +1,281 @@
...
+@skip_if_null
+class TestExample(unittest.TestCase):
+
+    def test_run(self):
+        check_special_files()
+        file_paths = chain.from_iterable(
+            get_python_files(source_dir) for source_dir in SOURCE_DIRS
+        )
+        for file_path in file_paths:
+            with self.subTest(file_path=file_path):
```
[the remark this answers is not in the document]
kitchoi: I don't think we want logging here. We typically want to use the test runner flag to determine verbosity. In most situations, tests should not unconditionally emit logs that will pollute the console (TraitsUI test suite is not in a clean state in that regard).

integrationtests/test_all_examples.py
kitchoi: Good point. Not sure why I did that. The intermediate wrapper is uncalled for.

integrationtests/test_all_examples.py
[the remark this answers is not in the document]
kitchoi: This might be relevant: https://bugs.python.org/issue35327

integrationtests/test_all_examples.py
kitchoi: The `os.path.abspath` is important actually. It removes the assumption that file path discovery must match the configuration for what files should be skipped.

ievacerny: I'm not sure I follow that. The assumption that the keys of `FILES_MAY_BE_SKIPPED` will match the paths collected during file discovery is there even with `os.path.abspath`. It's the assumption that both will be saved as absolute paths.

integrationtests/test_all_examples.py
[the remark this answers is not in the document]
ievacerny: The way I see it is that each example should have its own test and therefore should be logged just like other tests (not subtests). With that it would be very clear how many examples tests have run, in what order they were run, how many were skipped or have failed (if it's a big proportion of existing examples), which examples are being noisy, which ones have uncaught exceptions, etc... 

I agree that verbosity should be controlled via the test runner flag and after quick googling it seems that the flag can't be accessed from within a `TestCase`.

Given that, maybe `subTest` isn't the best approach here, maybe generating tests via a metaclass would work better?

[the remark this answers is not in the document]
kitchoi: Ah... I think that would be an overkill to workaround some missing feature of unittest test runner. Let's get the examples covered by tests and then we can make the test machinery better.

integrationtests/test_all_examples.py
```diff
@@ -45,133 +144,60 @@
...
-
-
-skip_file_if(
+SEARCHER = ExampleSearcher(source_dirs=SOURCE_DIRS)
+SEARCHER.skip_file_if(
     lambda: True, "Not a target file to be tested",
     os.path.join(DEMO, "demo.py"),
 )
-skip_file_if(
+SEARCHER.skip_file_if(
```
ievacerny, line 152: Nit: Naturally I would expect to read "'file' is skipped if 'condition' for 'reason'", so making `file_path` the first argument would be more readable.

integrationtests/test_all_examples.py
```diff
@@ -254,16 +280,10 @@ def run_file(file_path):
 class TestExample(unittest.TestCase):
```
ievacerny, line 280: Nit: Another header before this class would be good so that it's clear where "utility functions" end.

integrationtests/test_all_examples.py
```diff
@@ -0,0 +1,312 @@
...
+        Parameters
+        ----------
+        condition: callable() -> bool
+            The condition for skipping a file.
+        reason : str
+            Reason for skipping the file.
+        filepath : str
+            Path of the file which may be skipped from tests.
+        """
+        filepath = os.path.abspath(filepath)
```
kitchoi, line 61: If I did not have this line, the test will fail to skip the files appropriately, e.g. if the `filepath` given to `skip_file_if` is not an absolute path but `is_skipped` is called with an absolute path.

ievacerny: Yes, I know. But if `EXAMPLES` was saved as an absolute path, then all paths given to `skip_file_if` would also be absolute paths without this line.

ievacerny: Anyways, now that the code has been restructured, the registration via `skip_file_if` doesn't stand out as much, so it can stay as is, except for maybe changing the order of arguments (see comment above).

kitchoi: I see.

integrationtests/test_all_examples.py
kitchoi: Yep. Done.

PR #859, opened 2020-06-01
Move format_func, format_str and invalid traits from factory to editor

### LINKED ISSUE #790: format_func applied to Item's for which it was not specified
The example below on TraitsUI 6.1.3 produces the window:

image

where the second Item has had the `format_func` (only specified in the first Item) applied. A workaround is to specify `format_func=lambda s: s` in the second Item.

```python
from traits.api import *
from traitsui.api import *

class Foo(HasStrictTraits):
    thing = Str('a_thing')
    things = Str('one, two, three_four')

    traits_view = View(
        Item('thing', format_func=lambda s: s.replace('_', ' ').title()),
        Item('things')
    )

Foo().configure_traits()
```

ievacerny:
Closes #790 (by addressing points 2 and 3)

Moves `format_func`, `format_str` from `editor_factory.py` to `editor.py` and moves `string_value` conversion back to `editor.py`. Also moves `invalid` from `editory_factory.py` to `editor.py` where it is renamed to `invalid_trait_name`.

All three traits can still be set via the factory and `Item` specification, but for the sake of using them the traits on `editor` should be used instead (they should be the source of truth).

Note: there are no unit tests for `invalid` trait. To test it I used `text_editor_invalid.py` integration test.

### DIFF
--- a/traitsui/editor.py
+++ b/traitsui/editor.py
@@ -126,6 +126,10 @@ class Editor(HasPrivateTraits):
     #: Format string to use for formatting (used if **format_func** is not set)
     format_str = Str()
 
+    #: The extended trait name of the trait containing editor invalid state
+    #: status:
+    invalid_trait_name = Str()
+
     #: The current editor invalid state status:
     invalid = Bool(False)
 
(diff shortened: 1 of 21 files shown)


ievacerny:
Because of squash merge, this PR is showing commits from #848. The commits for this PR start with b3eeab1

PR #865, opened 2020-06-03
Make sure all UI in tests are disposed

### LINKED ISSUE #817: UI not closed at the end of tests
Stems from a comment on a PR: https://github.com/enthought/traitsui/pull/809#discussion_r423138475

There are lost of tests that don't close UI when they're done, which could lead to test interactions. It would be good to go through all the tests and fix this issue.

Example of such test:
https://github.com/enthought/traitsui/blob/098582414f8ebca4e27d56c6d87f8344a3a0e56b/traitsui/tests/test_labels.py#L196-L204

kitchoi:
Closes #817

This PR goes through all the existing tests to ensure `UI.dispose` is always called after every `HasTraits.edit_traits`, even if the test should fail. Most of these are achieved by using the `create_ui` context manager, however I have left equivalent code like this unchanged:
```
ui = obj.edit_traits()
self.addCleanup(ui.dispose)
```

~Still work-in-progress: This PR needs to be tested with the CI environment and I'd like to see if we could now relax some of the skips/workarounds for test interactions.~

Done. Ready for review.
Observation: There is nothing obviously different in the test results. What failed before still fail now. What succeeded before still succeed now.

### DIFF
--- a/traitsui/tests/editors/test_code_editor.py
+++ b/traitsui/tests/editors/test_code_editor.py
@@ -21,6 +21,7 @@ from traitsui.editors.code_editor import CodeEditor
 
 
 from traitsui.tests._tools import (
+    create_ui,
     skip_if_not_qt4,
     store_exceptions_on_all_threads,
 )
@@ -60,12 +61,10 @@ class TestCodeEditor(unittest.TestCase):
             return txt_ctrl.line_number_widget.isVisible()
 
         def test_line_numbers_visibility(show=True):
-            with store_exceptions_on_all_threads():
-                code_model = CodeModel()
-                code_view = CodeView(model=code_model, show_line_numbers=show)
-                ui = code_view.edit_traits()
+            code_model = CodeModel()
+            code_view = CodeView(model=code_model, show_line_numbers=show)
+            with store_exceptions_on_all_threads(), create_ui(code_view) as ui:
                 self.assertEqual(is_line_numbers_visible(ui), show)
-                ui.control.close()
 
         test_line_numbers_visibility(True)
         test_line_numbers_visibility(False)
@@ -76,10 +75,9 @@ class TestCodeEditor(unittest.TestCase):
         """
         from pyface import qt
 
-        with store_exceptions_on_all_threads():
-            code_model = CodeModel()
-            code_view = CodeView(model=code_model, style="readonly")
-            ui = code_view.edit_traits()
+        code_model = CodeModel()
+        code_view = CodeView(model=code_model, style="readonly")
+        with store_exceptions_on_all_threads(), create_ui(code_view) as ui:
             txt_ctrl = ui.control.findChild(qt.QtGui.QPlainTextEdit)
             self.assertTrue(txt_ctrl.isReadOnly())
 
@@ -94,5 +92,3 @@ class TestCodeEditor(unittest.TestCase):
             code_view.model = code_model2
             self.assertTrue(txt_ctrl.isReadOnly())
             self.assertEqual(txt_ctrl.toPlainText(), code_model.code)
-
-            ui.control.close()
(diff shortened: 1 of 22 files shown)


traitsui/tests/test_ui.py
```diff
@@ -190,25 +193,26 @@ def test_destroy_after_ok_qt(self):
...
-        self.assertEqual(control.deleteLater._n_calls, 1)
+            self.assertIsNone(ui.control)
+            self.assertEqual(control.deleteLater._n_calls, 1)
 
     @skip_if_null
     def test_no_spring_trait(self):
         obj = DisallowNewTraits()
-        ui = obj.edit_traits()
-        ui.dispose()
+        with store_exceptions_on_all_threads(), create_ui(obj):
```
kitchoi, line 215: The additional `store_exceptions_on_all_threads` here (and a couple of other tests) are drive-by and orthogonal. Now I am regretting mixing them in.  Will remove them again to keep this PR focus.

traitsui/tests/test_color_column.py
```diff
@@ -38,11 +42,9 @@ class TestColorColumn(TestCase):
...
         # Behaviour: column ui should display without error
 
-        with store_exceptions_on_all_threads():
-            d1 = MyEntry(name="a", value=2, color=(1.0, 0.3, 0.1))
-            d2 = MyEntry(name="b", value=3, color=(0.1, 0.0, 0.9))
-            data = MyData(data_list=[d1, d2])
-
-            ui = data.edit_traits()
-
-            ui.dispose()
```
kitchoi, line 48: This does not actually need changing, but it also does not hurt. Happy to revert if requested.

traitsui/tests/test_ui.py
```diff
@@ -162,25 +164,25 @@ def test_destroy_after_ok_wx(self):
         import wx
 
         foo = FooDialog()
-        ui = foo.edit_traits()
+        with create_ui(foo) as ui:
```
ievacerny, line 167: Doesn't this test (and the one below) need `store_exceptions_on_all_threads()`? Other tests might need it as well, but this might be better done in a separate PR

traitsui/tests/editors/test_code_editor.py
```diff
@@ -60,12 +61,10 @@ def is_line_numbers_visible(ui):
...
         def test_line_numbers_visibility(show=True):
-            with store_exceptions_on_all_threads():
-                code_model = CodeModel()
-                code_view = CodeView(model=code_model, show_line_numbers=show)
-                ui = code_view.edit_traits()
+            code_model = CodeModel()
+            code_view = CodeView(model=code_model, show_line_numbers=show)
+            with store_exceptions_on_all_threads(), create_ui(code_view) as ui:
                 self.assertEqual(is_line_numbers_visible(ui), show)
-                ui.control.close()
```
ievacerny, line 68: I don't think this action is performed in `ui.dispose()` so it might be better to leave it in the test for now (same in the test below).

traitsui/tests/test_color_column.py
ievacerny: There are more instances of this pattern in `ui_editors/test_data_frame.py`, `editors/test_tree_editor.py` (which are already being modified in this PR) and in `qt4/tests_test_ui_panel.py`. So if changing the pattern, I think it's best to change it everywhere.

kitchoi: I will revert this. I don't think this needs changing.

kitchoi: (On the other hand, I would also be in favour of making patterns more consistent, but maybe best done in a separate PR.)

traitsui/tests/test_ui.py
kitchoi: Agreed.

kitchoi: Opened https://github.com/enthought/traitsui/issues/876

Issue #870, opened 2020-06-04
TabularAdapter segfault on py3 + pyqt5


rahulporuri:
The following script can be run from any Python 3 environment (3.6.0, 3.6.5, 3.6.9 and 3.6.10) with any version of pyqt5 (5.6, 5.7.1 and 5.14.2).

Steps to reproduce
- run script
- drag-and-drop any of the items to the bottom of the window (white space in the window, where the element list ends)
- crash!

<details>

```
from traits.api import HasTraits, Instance, List, Property, Str
from traitsui.api import ModelView, TabularAdapter, TabularEditor, UItem, View


class AnObject(HasTraits):
    name = Str


class AModel(HasTraits):
    alist = List


class ATabularAdapter(TabularAdapter):

    model_view = Instance(ModelView)

    #### 'TabularAdapter' protocol ############################################

    columns = [('name', 'name')]

    name_text = Property(depends_on='item')

    def _get_name_text(self):
        return self.item.name


class AModelView(ModelView):

    model = Instance(AModel)

    def default_traits_view(self):

        body = UItem(
            'model.alist',
            editor=TabularEditor(
                adapter=ATabularAdapter(model_view=self),
                editable=True,
                operations=['move'],
                show_titles=False,
                vertical_lines=False,
                horizontal_lines=False,
                multi_select=True,
            ),
            springy=True,
        )

        return View(body, resizable=True)


def main():
    am = AModel()
    for idx in range(10):
        am.alist.append(AnObject(name=str(idx)))

    amv = AModelView(model=am)
    amv.configure_traits()


if __name__ == "__main__":
    main()
```

</details>

When you get it to segfault, if you enable the faulthandler, you should be able to see the following on the console - 

```
> edm run -e dc-cd -- python -Xfaulthandler a_model_view.py
QWindowsWindow::setGeometry: Unable to set geometry 87x192+916+419 on QWidgetWindow/'_StickyDialogClassWindow'. Resulting geometry:  148x192+916+419 (frame: 9, 38, 9, 9, custom margin: 0, 0, 0, 0, minimum size: 87x87, maximum size: 16777215x16777215).
Windows fatal exception: access violation

Current thread 0x00000d48 (most recent call first):
  File "C:\Users\rporuri\.edm\envs\dc-cd\lib\site-packages\traitsui\qt4\tabular_editor.py", line 464 in _multi_selected_rows_changed
[...]

PR #871, opened 2020-06-04
FIX: Make sure new_row is positive

jvkersch:
may fix -- I think -- #870 

`QModelIndex.row()` returns -1 when trying to append a row below existing rows (I didn't find any indication in the Qt docs why that should be the case, but I observed it in practice using the example from #870). This causes `TabularModel.beginInsertRows()` to try and insert a row at index -1, which causes a segfault. The fix is to adjust the row location to be non-negative, i.e. to make it equal to the number of rows already in the table, [as per the documentation](https://doc.qt.io/qt-5/qabstractitemmodel.html#beginInsertRows).

I've based my investigation entirely on the example from #870, I'm not sure if there are other places in the codebase that are susceptible to this.

@rahulporuri Could you give this a spin?

### DIFF
--- a/traitsui/qt4/tabular_model.py
+++ b/traitsui/qt4/tabular_model.py
@@ -344,6 +344,12 @@ class TabularModel(QtCore.QAbstractTableModel):
         """
         editor = self._editor
 
+        if new_row == -1:
+            # In some cases, the new row may be reported as -1 (e.g. when
+            # dragging and dropping a row at the bottom of existing rows). In
+            # that case, adjust to the number of existing rows.
+            new_row = self.rowCount(None)
+        
         # Sort rows in descending order so they can be removed without
         # invalidating the indices.
         current_rows.sort()


rahulporuri:
omg omg omg it actually worked on the script in #870 .

rahulporuri:
Tested this on the internal project where we noticed this error and we don't see the segfault on this branch! Thanks a lot Joris!

kitchoi:
Thank you @jvkersch  for the fix!

> QModelIndex.row() returns -1 when trying to append a row below existing rows

I think you are referring to this line:
https://github.com/enthought/traitsui/blob/96701ca9f2111735d9b842ff58a2418666bfbfe5/traitsui/qt4/tabular_model.py#L288

`parent.isValid()` returns false there if the item is dropped below the last item, indicating the row index is in fact invalid.  So for correctness, I think we shouldn't use `parent.row()` there but check the return result of `isValid()` and use `rowCount`.

On the other hand, I think the conversion from -1 to number of items in `moveRows` is worth keeping as a last resort / safety net, instead of crashing the application altogether.  I think we need a warning there to capture programming errors: caller of that function should ensure the row index is valid when it gets to that function. I'd be in favour of merging this and I will do a separate PR to address my own comments.

jvkersch:
@kitchoi That sounds good. I'm too unfamiliar with the codebase to make a strong recommendation, but I was looking at those two places as well, and I agree there's value in ensuring the right value is passed in as well as in checking that values are correct. 

If it makes things easier for you, feel free to close to this PR in favor of a new one, or to subsume it in a greater whole.

jvkersch:
Ahh, I didn't see this was merged in time. Never mind :)

kitchoi:
No problem. Thanks for the fix again!

PR #874, opened 2020-06-05
Add more TabularEditor tests

ievacerny:
Adds a few more tests for TabularEditor.

I wasn't trying to cover as much of the editor as in #869 so I was able to write these tests by just using model state. 

Just like #869 this also includes a minor bug fix in wx version of `TabularEditor` because it was easier to fix it than to open an issue.
The bug: in `_multi_selected_items_changed` instead of `self.value` the method was trying to use `self.values` which doesn't exist.

### DIFF
--- a/traitsui/tests/editors/test_tabular_editor.py
+++ b/traitsui/tests/editors/test_tabular_editor.py
@@ -122,6 +122,7 @@ def set_selected_single(editor, row):
 
         smodel = editor.control.selectionModel()
         mi = editor.model.index(row, 0)
+        # Add `Rows` flag to select the whole row
         smodel.select(
             mi, QItemSelectionModel.ClearAndSelect | QItemSelectionModel.Rows
         )
@@ -146,6 +147,7 @@ def set_selected_multiple(editor, rows):
         smodel = editor.control.selectionModel()
         for row in rows:
             mi = editor.model.index(row, 0)
+            # Add `Rows` flag to select the whole row
             smodel.select(
                 mi, QItemSelectionModel.Select | QItemSelectionModel.Rows
             )
--- a/traitsui/tests/editors/test_tabular_editor.py
+++ b/traitsui/tests/editors/test_tabular_editor.py
@@ -12,12 +12,18 @@
 import contextlib
 import unittest
 
+from pyface.gui import GUI
 from traits.api import Event, HasTraits, Instance, Int, List, Str
 from traits.testing.api import UnittestTools
 
 from traitsui.api import Item, TabularEditor, View
 from traitsui.tabular_adapter import TabularAdapter
-from traitsui.tests._tools import skip_if_null
+from traitsui.tests._tools import (
+    is_current_backend_wx,
+    is_current_backend_qt4,
+    skip_if_null,
+    store_exceptions_on_all_threads,
+)
 
 
 class Person(HasTraits):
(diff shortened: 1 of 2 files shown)


traitsui/tests/editors/test_tabular_editor.py
```diff
@@ -39,30 +45,326 @@ class Report(HasTraits):
...
+    if is_current_backend_wx():
+        editor.control.Select(row)
+
+    elif is_current_backend_qt4():
+        from pyface.qt.QtGui import QItemSelectionModel
+
+        smodel = editor.control.selectionModel()
+        mi = editor.model.index(row, 0)
+        smodel.select(
+            mi, QItemSelectionModel.ClearAndSelect | QItemSelectionModel.Rows
```
kitchoi, line 126: If you could comment on why we need the union of these two actions that will be very helpful for future developers who try to understand these tests. It is not obvious why the `Rows` flag is needed here.

traitsui/tests/editors/test_tabular_editor.py
```diff
@@ -39,30 +45,326 @@ class Report(HasTraits):
...
+
+    elif is_current_backend_qt4():
+        from pyface.qt.QtGui import QItemSelectionModel
+
+        clear_selection(editor)
+        smodel = editor.control.selectionModel()
+        for row in rows:
+            mi = editor.model.index(row, 0)
+            smodel.select(
+                mi, QItemSelectionModel.Select | QItemSelectionModel.Rows
```
kitchoi, line 150: Same here, it is not obvious why the `Rows` flag is needed. A comment will be very helpful.

traitsui/tests/editors/test_tabular_editor.py
```diff
@@ -39,30 +45,326 @@ class Report(HasTraits):
...
+            self.assertEqual(get_selected_rows(editor), [0, 1])
+            self.assertEqual(report.multi_selected, people[:2])
+
+            report.multi_selected = [people[2], people[0]]
+            gui.process_events()
+
+            self.assertEqual(sorted(get_selected_rows(editor)), [0, 2])
+            self.assertEqual(sorted(report.selected_rows), [0, 2])
+
+            # If there's a single invalid value, nothing is updated
```
kitchoi, line 298: Nit: This could be a separate unit test such that assertion failures above this line won't prevent this test from being run, which is probably more helpful for developers when making changes to the code being tested.

traitsui/tests/editors/test_tabular_editor.py
```diff
@@ -39,30 +45,326 @@ class Report(HasTraits):
...
+            invalid_person = Person(name="invalid", age=-1)
+            report.multi_selected = [people[2], invalid_person]
+            gui.process_events()
+
+            self.assertEqual(sorted(get_selected_rows(editor)), [0, 2])
+            self.assertEqual(sorted(report.selected_rows), [0, 2])
+
+            # Empty list clears selection
```
kitchoi, line 306: Nit: Same here, this could be a separate unit test.

traitsui/tests/editors/test_tabular_editor.py
ievacerny: The argument could apply to every single assertion group in this and other tests. I don't see anything special about this group so I would like to keep the test as is -- it's a reasonable size test that checks the behaviour of one specific function. And if somebody is working on that part of the code, they should make sure the whole test passes anyways.

kitchoi: The reason why I saw this block being separate is because it tests a separate code branch to the ones tested above, i.e. the `except` branch, verse the `else` branch in `_multi_selected_changed`.

kitchoi: It is okay keeping this test as is, which is why I mark this as Nit.

ievacerny: The same argument would still apply to more than these couple locations, so I'll keep them the same.

Issue #431, opened 2018-03-01
Lifecycle management under PyQt5 and Mac OS


corranwebster:
It looks like PyQt5 may be more finicky than other backaens on the Mac OS platform.  It seems that we can sometimes get late and/or spurious events firing after editors have been disposed of (so all Python references should have been dropped, but objects may still be live in memory).

Care should be taken to disconnect signals during `dispose` methods, and additionally it may make sense to guard against `self.control` being `None` in the signal handlers.

kitchoi:
Status update
-----

Some of the signals can be disconnected easily with little risk. Some of them are harder because they are made deep into the widget.  What I noticed was a lot of cycled references between toolkit widgets and TraitsUI editors. Where the connections are made by the editor, it is easy to disconnect them as part of the "dispose" method.

Where the connections are made by the widgets, then the options are either to expose a method on the widgets where the editor can call to disconnect the signal, or move the connection call back to the editors.

The following editors have their dispose implemented:
- drop_editor 
- text_editor 
- styled_date_editor 
- html_editor 
- file_editor 
- directory_editor 
- date_editor 
- date_range_editor  
- boolean_editor 
- button_editor 
- check_list_editor 
- datetime_editor 

The following editors are currently okay (nothing to do):
- title_editor 
- tuple_editor 
- null_editor 
- progress_editor 
- shell_editor 
- rgb_color_editor 
- key_binding_editor 
- image_editor 
- image_enum_editor 
- array_editor 
- array_view_editor 
- basic_editor_factory 
- custom_editor 
- data_frame_editor 
- clipboard 
- csv_list_editor

I could not get around to auditing or implementing the dispose for the following editors:
- code_editor 
- color_editor
- compound_editor
- enum_editor
- font_editor
- history_editor
- instance_editor
- list_editor
- list_str_editor
- range_editor
- search_editor
- set_editor
- table_editor
- tabular_editor
- time_editor
- tree_editor


Problems with dispose
----

We sometimes see `AttributeError` when we dispose of a UI (e.g. https://github.com/enthought/traitsui/issues/854). These AttributeErrors typically occur from an event handler. Disconnecting slots from signals are not enough because there are public slots overridden by TraitsUI. One example in Qt is resizeColumnToContents on QTableView.

When UI.dispose is called, the following occurs (as a very gross approximation):
(1) All the internal editors' dispose methods are called. In the editor's dispose, change handlers are removed and internal traits are set to None.
[...]

Issue #752, opened 2020-03-01
Wx Event Handler clean-up issues


corranwebster:
There are issues with the wxPython 4 backend where event handlers are not being cleaned up correctly.  It appears that all of the event handlers explicitly added by TraitsUI are being dealt with appropriately, but there appear to be some others which are internal to wxPython with aren't being cleaned up:

- scrollbars and some base Window objects seem to have an `EvtHandler` associated with them which needs to be cleaned up.
- the Html widget appears to have an `EvtHandler` associated with it that causes an error if it _is_ cleaned up.

This is easy to hack a fix for, but I suspect a deeper problem with finalization.

pbrod:
I have found that the call "control.FindWindowByName(name)" many times returns the wrong window/button in the tests. Then the wrong window/button is attempted disposed which often result in errors.  To find the correct window one must replace the call it with "control.FindWindowByName(name, control)".

Also to unbind correctly one must replace:
control.Unbind(wx.EVT_XXX, handler)" with "control.Unbind(wx.EVT_XXX, handler=handler)"

kitchoi:
Quite a number of editors call `Bind` but do not unbind the event handlers in their `dispose` methods, e.g. this:
https://github.com/enthought/traitsui/blob/64858b8c9ce9189a8c9cfde60610e4bd12f30962/traitsui/wx/editor_factory.py#L115-L116

This potentially causes some test interactions that we are seeing.
@corranwebster Are these missing `Unbind` part of this issue? Or should they be separate?

kitchoi:
> This potentially causes some test interactions that we are seeing.

For example: https://github.com/enthought/traitsui/issues/851

ievacerny:
`ListStrEditor` is suffering from this. It seems to be unbinding everything that it binds within TraitsUI but the error is still raised when closing this basic example UI:
```
class Foo(HasTraits):
    values = List(Str)
    view = View(Item('values', editor=ListStrEditor()))

demo = Foo(values=["one", "two", "three"])
demo.configure_traits()
```
```
Event loop failed to close cleanly:
Traceback (most recent call last):
  File "/Users/icernyte/Documents/repos/traitsui/traitsui/wx/view_application.py", line 139, in __init__
    start_event_loop_wx(self)
  File "/Users/icernyte/.edm/envs/traitsui-test-3.6-wx/lib/python3.6/site-packages/pyface/util/guisupport.py", line 115, in st
art_event_loop_wx
    app.MainLoop()
[...]

PR #877, opened 2020-06-08
Implement dispose for Qt TextEditor 

kitchoi:
Part of #431 

This PR implements the missing `dispose` method in Qt TextEditor.

In the process of testing the dispose method, I discovered a bug:
```
class Foo(HasTraits):
    value = Float()

view = View(
    Item('value', editor=TextEditor(auto_set=False), style="custom"),
)
Foo().configure_traits(view=view)
```

Launching the GUI fails with:
```
...
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/ui_panel.py", line 874, in _add_items
    editor.prepare(inner)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/editor.py", line 264, in prepare
    self.init(parent)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/text_editor.py", line 84, in init
    control.editingFinished.connect(self.update_object)
AttributeError: 'QTextEdit' object has no attribute 'editingFinished'
```

In order to test the new dispose for the custom editor, I also had to fix this bug and add more tests for the behaviour not previously tested.
I can also confirm the behaviour on `examples/demo/Standard_Editors/TextEditor_demo.py` has not changed with this PR.

### DIFF
--- a/traitsui/qt4/text_editor.py
+++ b/traitsui/qt4/text_editor.py
@@ -72,17 +72,18 @@ class SimpleEditor(Editor):
 
         self._signals = []
         if wtype == QtGui.QTextEdit:
+            control.textChanged.connect(self.update_object)
             self._signals.append((control.textChanged, self.update_object))
         else:
             # QLineEdit
             if factory.auto_set and not factory.is_grid_cell:
+                control.textEdited.connect(self.update_object)
                 self._signals.append((control.textEdited, self.update_object))
             else:
+                control.editingFinished.connect(self.update_object)
                 self._signals.append(
                     (control.editingFinished, self.update_object)
                 )
-        for signal, handler in self._signals:
-            signal.connect(handler)
 
         placeholder = self.factory.placeholder
 
--- a/traitsui/qt4/text_editor.py
+++ b/traitsui/qt4/text_editor.py
@@ -13,7 +13,6 @@
 """ Defines the various text editors for the PyQt user interface toolkit.
 """
 
-
 from pyface.qt import QtCore, QtGui
 
 from traits.api import TraitError
@@ -28,7 +27,6 @@ from .editor import Editor
 from .editor_factory import ReadonlyEditor as BaseReadonlyEditor
 
 from .constants import OKColor
-
 
 
 class SimpleEditor(Editor):
@@ -104,10 +102,9 @@ class SimpleEditor(Editor):
     def dispose(self):
         """ Disposes of the contents of an editor.
         """
-        if self.control is not None:
-            while self._signals:
-                signal, handler = self._signals.pop()
-                signal.disconnect(handler)
+        while self._signals:
+            signal, handler = self._signals.pop()
+            signal.disconnect(handler)
 
         super().dispose()
 
--- a/traitsui/qt4/text_editor.py
+++ b/traitsui/qt4/text_editor.py
@@ -72,16 +72,19 @@ class SimpleEditor(Editor):
         if factory.password:
             control.setEchoMode(QtGui.QLineEdit.Password)
 
-        if factory.auto_set and not factory.is_grid_cell:
-            if wtype == QtGui.QTextEdit:
-                control.textChanged.connect(self.update_object)
-            else:
-                control.textEdited.connect(self.update_object)
-
+        self._signals = []
+        if wtype == QtGui.QTextEdit:
+            self._signals.append((control.textChanged, self.update_object))
         else:
-            # Assume enter_set is set, otherwise the value will never get
-            # updated.
-            control.editingFinished.connect(self.update_object)
+            # QLineEdit
+            if factory.auto_set and not factory.is_grid_cell:
+                self._signals.append((control.textEdited, self.update_object))
+            else:
+                self._signals.append(
+                    (control.editingFinished, self.update_object)
+                )
+        for signal, handler in self._signals:
+            signal.connect(handler)
 
         placeholder = self.factory.placeholder
 
@@ -100,15 +103,9 @@ class SimpleEditor(Editor):
 
     def dispose(self):
         if self.control is not None:
-            control = self.control
-            if self.factory.auto_set and not self.factory.is_grid_cell:
-                if isinstance(control, QtGui.QTextEdit):
-                    control.textChanged.disconnect(self.update_object)
-                else:
-                    control.textEdited.disconnect(self.update_object)
-
-            else:
-                control.editingFinished.disconnect(self.update_object)
+            while self._signals:
+                signal, handler = self._signals.pop()
+                signal.disconnect(handler)
 
         super().dispose()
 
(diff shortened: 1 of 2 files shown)


traitsui/tests/editors/test_text_editor.py
```diff
@@ -99,3 +156,85 @@ def test_text_editor_custom_style_placeholder(self):
...
+        with create_ui(foo, dict(view=view)) as ui:
+            editor, = ui.get_editors("name")
+
+            set_text(editor, "NEW")
+            gui.process_events()
+
+            key_press_return(editor)
+            gui.process_events()
+
+            self.assertEqual(foo.name, "NEW\n")
```
kitchoi, line 240: When https://github.com/enthought/traitsui/issues/752 is also fixed, these tests can be pulled out of the `skip_if_not_qt4` decorator and be run against wx. In the mean time, doing so may introduce more test interactions.

kitchoi: Actually it might be easier if I pull these tests out now and skip on wx. Later when we get to #752, it should only require removing the skip and implementing the simulation functions (e.g. `key_press_return`) for wx.

traitsui/tests/editors/test_text_editor.py
```diff
@@ -99,3 +154,92 @@ def test_text_editor_custom_style_placeholder(self):
...
+class TestTextEditor(unittest.TestCase):
+    """ Tests that can be run with any toolkit as long as there is an
+    implementation for simulating user interactions.
+    """
+
+    def check_editor_init_and_dispose(self, style, auto_set):
+        # Smoke test to test setup and tear down of an editor.
+        foo = Foo()
+        view = get_view(style=style, auto_set=auto_set)
+        with create_ui(foo, dict(view=view)):
```
kitchoi, line 171: After opening #876, I forgot about that too... fixing...

traitsui/qt4/text_editor.py
```diff
@@ -72,16 +70,19 @@ def init(self, parent):
         if factory.password:
             control.setEchoMode(QtGui.QLineEdit.Password)
 
-        if factory.auto_set and not factory.is_grid_cell:
-            if wtype == QtGui.QTextEdit:
-                control.textChanged.connect(self.update_object)
-            else:
-                control.textEdited.connect(self.update_object)
-
+        self._signals = []
```
corranwebster, line 73: `_signals` should probably be declared as a trait (`List()` is probably good enough).

kitchoi: Now knowing what sort of things traits List does, I am less keen about bringing in `TraitListObject` for this...

kitchoi:
> I'm split between "disconnecting should be the exact mirror of connecting" and "this is absolutely the way that all editors should work!"

Me too... On the one hand, copying-and-pasting / compiling the setup code in one's head is in fact more consistent with other editors. On the other hand, it is easier to maintain and ensure correctness if we maintain a list of signals to pop in dispose.

kitchoi:
On the topic of making disconnection code being an exact mirror of connection VS keeping track of a list of connected signal, I find myself wanting to do the latter again for the `connect` calls embedded in `traitsui.qt4.instance_editor.CustomEditor.init`.

kitchoi:
From offline discussion: Let's make `connect` followed by an `append` action to the list instead of collecting the signal in advance before connecting them in batch.

kitchoi:
Sigh... in light of https://github.com/enthought/traitsui/issues/884, I went to find out if there are other editors subclassing `traitsui.qt4.text_editor.SimpleEditor`, and there are, e.g. `traitsui.qt4.file_editor.SimpleEditor`.
This simple `file_editor` does not implement its `dispose` method and will inherit whatever is in `traitsui.qt4.text_editor.SimpleEditor`.

It is okay because `self._signals` is going to be None per `HasPrivateTraits` magic and therefore falsy, but still, all these subclassing fragile.

Turns out `TextEditor` is subclassed by quite a few other editors, which makes it a bad candidate to be worked on early. I'd propose deferring this PR until we have better test coverage for other editors.

kitchoi:
> Turns out TextEditor is subclassed by quite a few other editors

These are its subclasses:
`traitsui.qt4.csv_list_editor.SimpleEditor`
`traitsui.qt4.csv_list_editor.CustomEditor`
`traitsui.qt4.drop_editor.SimpleEditor`
`traitsui.qt4.file_editor.SimpleEditor`

traitsui/qt4/text_editor.py
kitchoi: Done declaring `_signals` as a private trait:  This is also done in `traitsui.editor.Editor`, this is good because (1) there is a natural place for documentation, (2) we can actually gain benefits from the validating `Tuple(Any, Callable)` and (3) avoid `HasPrivateTraits` returning `None` in the unexpected places.

traitsui/qt4/text_editor.py
```diff
@@ -48,6 +46,13 @@ class SimpleEditor(Editor):
     #: Function used to evaluate textual user input:
     evaluate = evaluate_trait
 
+    # -- private trait definitions ------------------------------------------
+
+    #: A list of Qt signals connected.
+    #: First item in the tuple is the Qt signal. The second item is the event
+    #: handler.
+    _signals = List(Tuple(Any, Callable))
```
ievacerny, line 54: Suggestion on the name - maybe it would be better to use `connections_to_remove` or something similar? Because (1) the list doesn't contain just signals, it contains signals and handlers; (2) there might be connections that shouldn't be disconnected in `dispose` method (e.g. to signal `destroyed`) and the name would make it clear that those connections shouldn't be appended to the list.

traitsui/qt4/text_editor.py
```diff
@@ -72,16 +77,20 @@ def init(self, parent):
         if factory.password:
             control.setEchoMode(QtGui.QLineEdit.Password)
 
-        if factory.auto_set and not factory.is_grid_cell:
-            if wtype == QtGui.QTextEdit:
-                control.textChanged.connect(self.update_object)
-            else:
-                control.textEdited.connect(self.update_object)
-
+        self._signals = []
```
ievacerny, line 80: Why is this line in the `init` needed? Wouldn't `_signals` be an empty list without it (as per default `List` value)?

kitchoi: Not needed. It was code evolution. Will remove.

traitsui/qt4/text_editor.py
kitchoi: Agreed with (1). Just wanted to clear, we still want this variable to be private right, i.e. with a preceding underscore?  If (2) does come up, then traitsui will have more flexibility to rename an internal private variable.

ievacerny: I think it should be private

kitchoi:
Update: the following editors subclassing `TextEditor` are now covered in tests for UI disposal

traitsui.qt4.csv_list_editor.SimpleEditor
traitsui.qt4.csv_list_editor.CustomEditor
traitsui.qt4.drop_editor.SimpleEditor
traitsui.qt4.file_editor.SimpleEditor

IOW, this PR is good to go in.

PR #881, opened 2020-06-08
Implement dispose method for CheckListEditor (Qt)

kitchoi:
Part of #431

This PR does the following for the Qt implementation of `CheckListEditor`:
- `CustomEditor` no longer subclasses `SimpleEditor` in order to reuse the `list_updated` method. This relationship makes it difficult to implement `dispose` as the two editors need to perform different cleanup. The `list_updated` logic is now refactored out to a base class.  Then it is easier to see what needs to be disconnected and cleaned up in each of these concrete subclasses.
- Added `dispose_child` (naming to be confirmed?) in `traitsui.qt4.editor.Editor` which will be called when a child widget in a layout is removed.  The default implementation does nothing, and the `CheckListEditor` reimplements this method. This should be done for `EnumEditor` as well (not in this PR, but will be part of #431 as well).

See #882 too for an alternative approach.

Note: The refactored `BaseCheckListEditor` actually shares **identical** code with the implementation in wx. This should be further refactored into a toolkit-agnostic base class or mixin. This should be done in a separate PR though.

### DIFF
--- a/traitsui/qt4/check_list_editor.py
+++ b/traitsui/qt4/check_list_editor.py
@@ -40,37 +40,16 @@ logger = logging.getLogger(__name__)
 capitalize = lambda s: s.capitalize()
 
 
-# -------------------------------------------------------------------------
-#  'SimpleEditor' class:
-# -------------------------------------------------------------------------
-
-
-class SimpleEditor(EditorWithList):
-    """ Simple style of editor for checklists, which displays a combo box.
+class BaseCheckListEditor(EditorWithList):
+    """ Base class that implement the common logic of simple and custom
+    CheckListEditor.
     """
-
-    # -------------------------------------------------------------------------
-    #  Trait definitions:
-    # -------------------------------------------------------------------------
 
     #: Checklist item names
     names = List(Str)
 
     #: Checklist item values
     values = List()
-
-    def init(self, parent):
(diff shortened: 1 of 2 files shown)


traitsui/qt4/check_list_editor.py
```diff
@@ -132,11 +151,42 @@ def update_editor(self):
...
+        the control layout.
+
+        Parameters
+        ----------
+        child : QWidget
+        """
+        if self._mapper is not None:
+            child.clicked.disconnect(self._mapper.map)
+            self._mapper.removeMappings(child)
+
```
kitchoi, line 189: An alternative to this is to maintain a list of signals inside `CheckListEditor`, like in #877. I might try that now...

kitchoi: See alternative PR: https://github.com/enthought/traitsui/pull/882

kitchoi: One thing I did not like about `dispose_child` is that it is only useful if all the children have the same type and the same function such that the same disposal code can be applied to everyone of them. This is unlikely to be generally useful: A layout can have a mixture of widgets.
(Preferring #882 to this one)

kitchoi:
Closing this in favour of #882. Sorry for the noise.

PR #882, opened 2020-06-08
Alternative approach to implement dispose for Qt CheckListEditor

kitchoi:
This PR presents an alternative approach to https://github.com/enthought/traitsui/pull/881
~One of these two can merge, but not both :)~.  I have unilaterally closed #881 in favour of this.

~Like in #881, this is also done here:~
- ~`CustomEditor` no longer subclasses `SimpleEditor` in order to reuse the `list_updated` method. This relationship makes it difficult to implement `dispose` as the two editors need to perform different cleanup. The `list_updated` logic is now refactored out to a base class.  Then it is easier to see what needs to be disconnected and cleaned up in each of these concrete subclasses.~ (Edited: This is removed)

Instead of adding `dispose_child` to the editor base class's `clear_layout`, the approach in this PR keeps lists of connected signals.

In particular, `append` is called immediately after `connect`. This is particularly important because `rebuild_editors` is called from a change handler, and it is possible for it to fail halfway.

### DIFF
--- a/traitsui/qt4/check_list_editor.py
+++ b/traitsui/qt4/check_list_editor.py
@@ -40,9 +40,13 @@ logger = logging.getLogger(__name__)
 capitalize = lambda s: s.capitalize()
 
 
-class BaseCheckListEditor(EditorWithList):
-    """ Base class that implements the common logic of simple and custom
-    CheckListEditor.
+# -------------------------------------------------------------------------
+#  'SimpleEditor' class:
+# -------------------------------------------------------------------------
+
+
+class SimpleEditor(EditorWithList):
+    """ Simple style of editor for checklists, which displays a combo box.
     """
 
     # -------------------------------------------------------------------------
@@ -54,6 +58,27 @@ class BaseCheckListEditor(EditorWithList):
 
     #: Checklist item values
     values = List()
+
+    def init(self, parent):
+        """ Finishes initializing the editor by creating the underlying toolkit
+            widget.
+        """
+        self.create_control(parent)
+        super(SimpleEditor, self).init(parent)
+
+    def create_control(self, parent):
+        """ Creates the initial editor control.
+        """
+        self.control = QtGui.QComboBox()
+        self.control.activated[int].connect(self.update_object)
+
+    def dispose(self):
+        """ Disposes of the contents of an editor.
+        """
+        if self.control is not None:
+            self.control.activated[int].disconnect(self.update_object)
+
+        super().dispose()
 
     def list_updated(self, values):
         """ Handles updates to the list of legal checklist values.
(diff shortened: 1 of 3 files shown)


traitsui/qt4/check_list_editor.py
```diff
@@ -103,6 +86,42 @@ def list_updated(self, values):
...
+        self.create_control(parent)
+        super(SimpleEditor, self).init(parent)
+
+    def dispose(self):
+        """ Disposes of the contents of an editor.
+        """
+        if self.control is not None:
+            self.control.activated[int].disconnect(self.update_object)
+
+        super().dispose()
```
kitchoi, line 117: Only added `dispose` for `SimpleEditor`, the `init` and `create_control` are the same as before.

traitsui/qt4/check_list_editor.py
```diff
@@ -132,11 +151,41 @@ def update_editor(self):
...
+            signal.disconnect(handler)
+
+        while self._connections_to_dispose:
+            signal, handler = self._connections_to_dispose.pop()
+            signal.disconnect(handler)
+
+        if self._mapper is not None:
+            self._mapper = None
+
+        super().dispose()
```
kitchoi, line 187: In the other PRs, I decided to just call `Editor.dispose(self)` and leave #884 for another day. I will do that here too to be consistent.

kitchoi:
Sorry for the late changes: Pulling out a base class for this one editor is not hard and does not look fiddly. However, turns out this is quite a common pattern across many editors and many small tasks add up into a big one, incurring more risk. At the end I have decided to defer #884 to be solved separately.

traitsui/qt4/check_list_editor.py
```diff
@@ -137,6 +145,13 @@ class CustomEditor(SimpleEditor):
         boxes.
     """
 
+    #: List of tuple(signal, slot) to be disconnected in dispose
+    _connections_to_dispose = List(Tuple(Any, Callable))
+
+    #: List of tuple(signal, slot) to be disconnected while rebuilding the
+    #: editor.
+    _connections_to_rebuild = List(Tuple(Any, Callable))
+
```
kitchoi, line 154: I will add a `init` here to do exactly what `SimpleEditor` does, but call `EditorWithList.init(self)` directly. This will read correctly with `dispose` and shield `CustomEditor.init` from changes in `SimpleEditor.init`.  The `init` and `dispose` of these two classes should be decoupled.

traitsui/qt4/check_list_editor.py
```diff
@@ -146,10 +168,34 @@ def create_control(self, parent):
 
         self._mapper = QtCore.QSignalMapper()
         self._mapper.mapped[str].connect(self.update_object)
+        self._connections_to_dispose.append(
```
ievacerny, line 171: I see only one connection in this list and looks like it could be disconnected without the list, so maybe `_connections_to_dispose` is unnecessary?

kitchoi: Yep, agreed. I reverted this back to just calling disconnect.
Thanks!

Issue #879, opened 2020-06-08
Deprecate values_trait in ButtonEditor?


kitchoi:
The Qt implementation of the `ButtonEditor` takes into account of the `values_trait` in the editor factory:
https://github.com/enthought/traitsui/blob/25e6f83aec9c483b189ec43037440dd71bc4745d/traitsui/editors/button_editor.py#L53-L57

The user manual says more or less the same thing:
> Use values_trait to specify the name of the trait on the object that contains the list of possible values. If this is set, then the value, label, and label_value traits are ignored; instead, they will be set from this list. When this button is clicked, the value set will be the one selected from the drop-down.

But there are no implementations for this trait in wx `ButtonEditor` or the custom version of qt `ButtonEditor`. I also could work out how this is supposed to work: I got as far as getting a button but there is no drop-down. The description of this trait also looks similar to what `EnumEditor` does.

All these signs point to this question: Can we deprecate `values_trait` on `ButtonEditor` instead of trying to fix/maintain it?

kitchoi:
This issue was motivated by #431. In particular, the `on_trait_change` put up for `values_traits` is not taken down in `dispose`:
https://github.com/enthought/traitsui/blob/f919dbd2044bcbff7280e5580286e93caa8a153a/traitsui/qt4/button_editor.py#L56-L61

PR #885, opened 2020-06-09
Reverse on_trait_change hook in Qt ButtonEditor's dispose

kitchoi:
Part of #431

This PR adds missing dispose logic for Qt's ButtonEditor.

### DIFF
--- a/traitsui/qt4/button_editor.py
+++ b/traitsui/qt4/button_editor.py
@@ -168,6 +168,8 @@ class CustomEditor(SimpleEditor):
         self.set_tooltip()
 
     def dispose(self):
+        """ Disposes of the contents of an editor.
+        """
         if self.control is not None:
             self.control.clicked.disconnect(self.update_object)
 
--- a/traitsui/qt4/button_editor.py
+++ b/traitsui/qt4/button_editor.py
@@ -75,6 +75,19 @@ class SimpleEditor(Editor):
     def dispose(self):
         """ Disposes of the contents of an editor.
         """
+
+        if self.factory.values_trait:
+            self.object.on_trait_change(
+                self._update_menu,
+                self.factory.values_trait,
+                remove=True,
+            )
+            self.object.on_trait_change(
+                self._update_menu,
+                self.factory.values_trait + "_items",
+                remove=True,
+            )
+
         if self.control is not None:
             self.control.clicked.disconnect(self.update_object)
         super(SimpleEditor, self).dispose()
@@ -153,3 +166,10 @@ class CustomEditor(SimpleEditor):
         self.sync_value(self.factory.label_value, "label", "from")
         self.control.clicked.connect(self.update_object)
         self.set_tooltip()
+
+    def dispose(self):
+        if self.control is not None:
+            self.control.clicked.disconnect(self.update_object)
+
+        # enthought/traitsui#884
+        Editor.dispose(self)
--- a/traitsui/qt4/button_editor.py
+++ b/traitsui/qt4/button_editor.py
@@ -173,5 +173,6 @@ class CustomEditor(SimpleEditor):
         if self.control is not None:
             self.control.clicked.disconnect(self.update_object)
 
+        # FIXME: Maybe better to let this class subclass Editor directly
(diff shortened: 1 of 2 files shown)


traitsui/tests/editors/test_button_editor.py
```diff
@@ -85,3 +89,39 @@ def test_simple_button_editor(self):
...
+        with store_exceptions_on_all_threads():
+            with create_ui(instance, dict(view=view)):
+                pass
+
+            # It is okay to mutate trait after the GUI is disposed.
+            instance.values = ["Item3"]
+
+    def test_simple_editor_values_trait_init_and_dispose(self):
+        # Smoke test to check init and dispose when values_trait is used.
+        self.check_editor_values_trait_init_and_dispose(style="simple")
```
kitchoi, line 123: This test failed before fixing `SimpleEditor.dispose`.

traitsui/qt4/button_editor.py
```diff
@@ -153,3 +166,12 @@ def init(self, parent):
...
         self.set_tooltip()
+
+    def dispose(self):
+        """ Disposes of the contents of an editor.
+        """
+        if self.control is not None:
+            self.control.clicked.disconnect(self.update_object)
+
+        # enthought/traitsui#884
+        Editor.dispose(self)
```
ievacerny, line 177: Nit: I would add a `FIXME` to the comment so that it clearly stands out as something to be fixed.

traitsui/tests/editors/test_button_editor.py
```diff
@@ -85,3 +89,39 @@ def test_simple_button_editor(self):
...
+                style=style,
+            ),
+        )
+
+    def check_editor_values_trait_init_and_dispose(self, style):
+        # Smoke test to check init and dispose when values_trait is used.
+        instance = ButtonTextEdit(values=["Item1", "Item2"])
+        view = self.get_view(style=style)
+        with store_exceptions_on_all_threads():
+            with create_ui(instance, dict(view=view)):
```
ievacerny, line 115: Is this okay without `gui.process_events()`? There are quite a few qt methods called in the code.

traitsui/qt4/button_editor.py
```diff
@@ -75,6 +75,19 @@ def init(self, parent):
     def dispose(self):
         """ Disposes of the contents of an editor.
         """
+
+        if self.factory.values_trait:
```
ievacerny, line 79: There are also connections made between `action.triggered` and a lambda function in `_update_menu`. They will be disconnected when `action` objects are destroyed. I wonder if it's worth adding explicit disconnects for the case when the objects aren't destroyed for some reason.

traitsui/tests/editors/test_button_editor.py
kitchoi: I am about to open a PR to add GUI event queue processing prior to all disposal of UIs and convert all tests to do that. For this test, nothing external to the editor's code causes new events in between `init` and `dispose`, so this is okay, but we should do this flush more generally.

I also noticed that the Qt implementation for `GUI.process_events` does not actually do what it says in the tin if one reads it without thinking/reading harder about how Qt works.  That is going to be addressed in that separate PR mentioned above.

traitsui/qt4/button_editor.py
kitchoi: Considering that the feature does not seem to work, I am leaning towards the answer "not worth it". See https://github.com/enthought/traitsui/issues/879

traitsui/tests/editors/test_button_editor.py
kitchoi: Could I NOT add this `gui.process_events()` here? For the following reasons:
(1) `create_ui` should do that clean up, not here
(2) If I add it here as a "just-in-case" measure, it would be confusing to others whether it is generally needed for all the tests that just calls `create_ui` for opening and closing the UI. 
(3) There are no race conditions for the setup and teardown of this simple button editor as far as I can see.

Confusion is one of the drawbacks of this kind of defensive programming. For (3), I think it is worth avoiding that confusion.

PR #886, opened 2020-06-09
Implement dispose for DateEditor (Qt)

kitchoi:
Part of #431 

This PR implements the `dispose` method for Qt DateEditor.

The custom DateEditor is also subclassed by `DateRangeEditor` and `StyledDateEditor`. I checked `DateRangeEditor` does not need changes because it basically just calls `super().init(parent)` and so it makes sense to carry on using the inherited `dispose`.

`StyledDateEditor.init` has much in common with the base but it did not call the base class `init`. The option is (1) keep the `init` as is but add a `dispose` and then call `traitsui.editor.Editor.dispose` there, or (2) reuse the base class `init` so that it makes sense to inherit the base class `dispose` too. I opt-ed for the second option.

Tests for `DateRangeEditor` already cover init and dispose. New tests are added for `StyledDateEditor` to confirm the init-dispose works. I did some manual testing there too.

### DIFF
--- a/traitsui/tests/editors/test_styled_date_editor.py
+++ b/traitsui/tests/editors/test_styled_date_editor.py
@@ -19,7 +19,7 @@ class DateModel(HasTraits):
     styles_mapping = Dict(Str, Instance(CellFormat))
 
 
-def example_model():
+def get_example_model():
     return DateModel(
         special_days={
             "public-holidays": [datetime.date(2020, 1, 1)],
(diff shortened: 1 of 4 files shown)


traitsui/tests/editors/test_styled_date_editor.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+
+# StyledDateEditor is currently only implemented for Qt
+@skip_if_not_qt4
+class TestStyledDateEditor(unittest.TestCase):
+
+    def test_init_and_dispose(self):
+        # Smoke test to test init and dispose.
+        instance = example_model()
```
ievacerny, line 42: Nit: `get_example_model()` would read better. Now this looks too similar to the instantiation of a class.

PR #887, opened 2020-06-09
Implement dispose for HTMLEditor (Qt)

kitchoi:
Part of #431 

This PR implements the `dispose` method of `HTMLEditor` (Qt).
Confirmed that no other editors in traitsui subclass `HTMLEditor` (Qt)

### DIFF
--- a/traitsui/qt4/html_editor.py
+++ b/traitsui/qt4/html_editor.py
@@ -62,6 +62,14 @@ class SimpleEditor(Editor):
         self.base_url = self.factory.base_url
         self.sync_value(self.factory.base_url_name, "base_url", "from")
 
+    def dispose(self):
+        """ Disposes of the contents of an editor.
+        """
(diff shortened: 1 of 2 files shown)


kitchoi:
The CI failure is caused by the Qt5 egg being built against a different version of libxml2, which only surfaces whenever there is a test that imports `PyQt5.QtWebKit`.
There are two possible workarounds: (1) Use a newer OSX image on Travis, (2) Add `libxml2` to the dependencies in `etstools.py`, but then I think the Qt5 egg is going to be fixed soon so we could also just wait a bit.

kitchoi:
It might be available now. Let me try rebuilding the CI job.

kitchoi:
Either because the build was too old, or I am running into sign-in issues with Travis-CI org again, I can't rebuild the previous run, I am going to use the close and reopen hack...

Issue #889, opened 2020-06-10
Simple FileEditor cannot be launched in Wx


kitchoi:
To reproduce:
```
from traits.api import *
class FileModel(HasTraits):
    filepath = File()

m = FileModel()
m.configure_traits()
```

Output:
```
15:44:18: Debug: Adding duplicate image handler for 'Windows bitmap file'
15:44:18: Debug: Adding duplicate animation handler for '1' type
15:44:18: Debug: Adding duplicate animation handler for '2' type
15:44:18: Debug: Adding duplicate image handler for 'Windows bitmap file'
15:44:18: Debug: Adding duplicate animation handler for '1' type
15:44:18: Debug: Adding duplicate animation handler for '2' type
15:44:18: Debug: Adding duplicate image handler for 'Windows bitmap file'
15:44:18: Debug: Adding duplicate animation handler for '1' type
15:44:18: Debug: Adding duplicate animation handler for '2' type
Traceback (most recent call last):
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/wx/view_application.py", line 152, in OnInit
    args=self.args,
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/view.py", line 462, in ui
    ui.ui(parent, kind)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/ui.py", line 246, in ui
    self.rebuild(self, parent)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/wx/toolkit.py", line 125, in ui_live
    ui_live.ui_live(ui, parent)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/wx/ui_live.py", line 52, in ui_live
    _ui_dialog(ui, parent, BaseDialog.NONMODAL)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/wx/ui_live.py", line 88, in _ui_dialog
    BaseDialog.display_ui(ui, parent, style)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/wx/ui_base.py", line 71, in display_ui
    ui.owner.init(ui, parent, style)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/wx/ui_live.py", line 203, in init
    sw = panel(ui, window)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/wx/ui_panel.py", line 272, in panel
    panel, content[0], ui
[...]

Issue #884, opened 2020-06-09
Awkward editor disposal when editors are defined by subclassing another


kitchoi:
This pattern is seen in a number of editors: a concrete editor is defined by subclassing another concrete editor. By concrete editors, I mean a subclass of `traitsui.editor.Editor` that is intended to be instantiated directly to be used in `traitsui.ui.UI._editors`. 

For example, there is a concrete editor `traitsui.qt4.text_editor.SimpleEditor`, which is subclassed by another concrete editor `traitsui.qt4.text_editor.CustomEditor`.

This makes implementing the `dispose` method awkward when trying to clean up widgets upon UI disposal.

e.g.
`traitsui.qt4.button_editor.CustomEditor` subclasses `traitsui.qt4.button_editor.SimpleEditor` in order to reuse the `update_object` and `update_editor` defined there.

`SimpleEditor` has `init` and `dispose` methods. However `CustomEditor` only overrides `SimpleEditor`'s `init`, and does not implement a `dispose`.

The `SimpleEditor.dispose` is in fact incorrect: It needs to remove the on_trait_change handlers put up in `SimpleEditor.init`. However, such removal is NOT needed by `CustomEditor`.  Consequently, one may need to override `dispose` in `CustomEditor`, but then it will also need to call `traitsui.editor.Editor.dispose`.  Simply calling `super().dispose()`  will end up calling `SimpleEditor.dispose` which will try to remove change handlers that don't exist.

Proposed solutions (they are not mutually exclusive):
- Avoid subclassing editors that are intended to be a leaf class: Refactor the common logic to a base class that is subclassed by the concrete editors.
- Move the logic in `traitsui.editor.Editor.dispose` out to another maybe-private method so that it is always run without relying on subclass to always call it. Its docstring says it is intended to be overridden to provide *additional clean-up*. If the base class implementations always needs to be run, it might make sense for the base class to ensure that happens.
- More tools for disposal: Maybe we need a method for adding cleanup steps in a stack, similar to `unittest.TestCase.addCleanup`. `unittest.TestCase.setUp` will be analogous to `traitsui.editor.Editor.init`. `unittest.TestCase.tearDown` is analogous to `traitsui.editor.Editor.dispose`.

kitchoi:
There are probably a lot of subclasses of concrete editors in the wild (i.e. outside of traitsui), and they might follow the same pattern of overriding `init` without overriding the `dispose`.

[...]

Issue #888, opened 2020-06-10
Qt IconButton let its owner be responsible for connecting signals


kitchoi:
`traitsui.qt4.helper.IconButton` extends `QtGui.QPushButton`, and its `__init__` does two things:
(1) Configure the visual things, e.g. setting the size limit
(2) Connect a slot to its `clicked` signal.

The widget would be more reusable if it does not perform the second step, but instead let its caller (e.g. `FileEditor`) does the signal connections. This will make the signal connection/disconnection more obvious too.

It could just be me: I don't normally expect instantiating the widget should have the side-effect of connecting any signals.

PR #890, opened 2020-06-10
Implement dispose for FileEditor (Qt)

kitchoi:
Part of #431

This PR implements the `dispose` method for Qt's simple and custom FileEditor. New tests are added to ensure launching and disposal don't fail.

As in other PRs, to avoid creating more test interactions in the mean time for wx test suite, the new tests are skipped on wx until #752 is fixed as well.
I did try running them against wx, and discovered #889.

Not sure why these editors need to subclass the `SimpleEditor` in `text_editor`. This subclassing makes `dispose` awkward. I am deferring this refactoring/cleanup to #884 as this is a common pattern for a lot of editors.


traitsui/qt4/file_editor.py
```diff
@@ -48,16 +54,36 @@ def init(self, parent):
...
+            self._connections_to_remove.append(
+                (control.editingFinished, self.update_object)
+            )
 
         button = IconButton(QtGui.QStyle.SP_DirIcon, self.show_file_dialog)
+        # Call to connect should be moved out of IconButton
+        # (enthought/traitsui#888)
+        self._connections_to_remove.append(
+            (button.clicked, self.show_file_dialog)
+        )
```
kitchoi, line 73: I think it might be better to move this inside `IconButton` and expose it as a separate method, because then the signal `clicked` is mentioned in the same place. I think the `IconButton` should behave like `QPushButton` and not try to connect signals at instantiation.
Moving the logic into `IconButton`...

kitchoi: Now I remember why I didn't do it: It was because that requires `IconButton` to hold a reference to the editor's method, creating a reference cycle. 
On balance I am going to skip this disconnection because (1) `connect` and `disconnect` should be co-located, having them separate risk erroneous inconsistency in the future - attempt to disconnect an unconnected signal will cause Qt to raise an error and (2) having them co-located without modifying the public interface of `IconButton` will create reference cycle, and is not a very good pattern anyway.

This should be dealt within #888

ievacerny: Should this line be removed then?

kitchoi: Yes. Somehow I only managed to remove the comment 🤦 . Thank you!

PR #893, opened 2020-06-10
Wrap DatetimeEditor tests with store_exceptions_on_all_threads

kitchoi:
The `store_exceptions_on_all_threads` does a few things, one of those is to capture exceptions 
 from trait change and then re-raise. 
This is needed if an error occurs in the change handlers for `minimum_datetime` and `maximum_datetime` on the DatetimeEditor.

For example if the `if self.control is None` was removed from the change handler, we get errors logged on the console but the tests still pass. Wrapping the test with the context manager make them fail in that scenario.

Motivated by #431 which also concerns the hooking and unhooking of trait change handlers. 

Solves part of #876

### DIFF
--- a/traitsui/tests/editors/test_datetime_editor.py
+++ b/traitsui/tests/editors/test_datetime_editor.py
@@ -43,7 +43,8 @@ class TestDatetimeEditorQt(GuiTestAssistant, unittest.TestCase):
     def test_datetime_editor_simple(self):
         view = get_date_time_simple_view(DatetimeEditor())
         instance = InstanceWithDatetime(date_time=datetime.datetime.now())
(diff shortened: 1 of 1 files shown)


traitsui/tests/editors/test_datetime_editor.py
```diff
@@ -52,7 +54,8 @@ def test_datetime_editor_simple_with_minimum_datetime(self):
         )
         view = get_date_time_simple_view(editor_factory)
         instance = InstanceWithDatetime()
-        with self.launch_editor(instance, view) as editor:
+        with store_exceptions_on_all_threads(), \
```
ievacerny, line 57: There's one more test above - `test_datetime_editor_simple`

PR #895, opened 2020-06-11
Fix crash when number of columns changes

### LINKED ISSUE #894: TabularEditor crashes when reducing number of columns in adapter
I have a custom TabularAdapter that changes dynamically the number of columns in a TabularEditor based on the size of the underlying variable.

When I reduce the number of columns, I get this crash:

```
  File "C:\WinPython-64bit-3.8.3.0\python-3.8.3.amd64\lib\site-packages\traitsui\qt4\tabular_editor.py", line 913, in columnResized
  File "C:\WinPython-64bit-3.8.3.0\python-3.8.3.amd64\lib\site-packages\traitsui\qt4\tabular_editor.py", line 894, in resizeColumnsToContents
    widths = compute_column_widths(
  File "C:\WinPython-64bit-3.8.3.0\python-3.8.3.amd64\lib\site-packages\traitsui\helper.py", line 168, in compute_column_widths
    widths[i] = user_width
IndexError: list assignment index out of range
```

It seems the issue is with the user width array which is still of the wrong size.
I am submitting a PR that fixes this by resetting the user widths when there is a mismatch in size.

FedeMiorelli:
Fixes #894

### DIFF
--- a/traitsui/qt4/tabular_editor.py
+++ b/traitsui/qt4/tabular_editor.py
@@ -875,6 +875,7 @@ class _TableView(QtGui.QTableView):
         else:
             return self.horizontalHeader().sectionSizeHint(column)
 
+
     def resizeColumnsToContents(self):
         """ Reimplemented to support proportional column width specifications.
 
@@ -895,6 +896,9 @@ class _TableView(QtGui.QTableView):
             requested.append(width)
             min_widths.append(self.sizeHintForColumn(column))
 
+        if self._user_widths is not None and len(self._user_widths) != len(self._editor.adapter.columns):
+            self._user_widths = None
+
         widths = compute_column_widths(
             available_space, requested, min_widths, self._user_widths
         )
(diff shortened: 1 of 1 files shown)


kitchoi:
@FedeMiorelli Thank you very much for the bug report and the fix!  I can confirm that this fixes the bug, and more than that. I think you are right that the `user_widths` needs resetting. That really solves the bulk of the problem.

There is a different place where we can reset the `user_widths` whenever the adapter `columns` changes, and that might be a better place than `columnResized`, which is called whenever the user manually adjusts the column widths as well.

It is possible that I have missed something. If you don't mind, could you please try the branch in https://github.com/enthought/traitsui/pull/897 and see if that also resolves the issue? 

Thanks again!

FedeMiorelli:
@kitchoi Thanks for providing a better solution! Your branch works fine for me.
Cheers

kitchoi:
@FedeMiorelli  #897 is now merged with you as a co-author. Thank you very much for your contribution!  I believe this PR can now be closed, is that alright?

FedeMiorelli:
Thanks @kitchoi yes please go ahead and close this!
Cheers

kitchoi:
Thank you!  Closing.

PR #896, opened 2020-06-11
Implement dispose for DropEditor (Qt)

kitchoi:
Part of #431 

The PR adds open-and-close tests for DropEditor which are already passing without the new dispose method.

A `dispose` method is still added to ensure correctness: DropEditor uses the simple `TextEditor`'s `init` only if `readonly` is false. The new `dispose` to ensure it does not call the TextEditor dispose when the TextEditor `init` is not actually used. See #884.

(Orthogonal note 1: I tried but I still could not figure out how DropEditor is meant to be used.)
(Orthogonal note 2: The `_DropEventFilter` in this module is used by the Qt InstanceEditor.)

### DIFF
--- a/traitsui/qt4/drop_editor.py
+++ b/traitsui/qt4/drop_editor.py
@@ -22,6 +22,7 @@ from pyface.qt import QtGui, QtCore
 # traitsui.editors.drop_editor file.
 from traitsui.editors.drop_editor import ToolkitEditorFactory
 
+from .editor import Editor as _BaseEditor
 from .text_editor import SimpleEditor as Editor
 from .constants import DropColor
 from .clipboard import PyMimeData, clipboard
@@ -55,6 +56,15 @@ class SimpleEditor(Editor):
         self.control.installEventFilter(drop_event_filter)
 
         self.control._qt4_editor = self
+
+    def dispose(self):
+        """ Disposes of the content of an editor.
+        """
+        if self.factory.readonly:
+            # enthought/traitsui#884
+            _BaseEditor.dispose(self)
+        else:
+            super(SimpleEditor, self).dispose()
 
     def string_value(self, value):
         """ Returns the text representation of a specified object trait value.
--- a/traitsui/tests/editors/test_drop_editor.py
+++ b/traitsui/tests/editors/test_drop_editor.py
@@ -29,10 +29,20 @@ class Model(HasTraits):
 class TestDropEditor(unittest.TestCase):
     """ Test DropEditor. """
 
-    def test_init_dispose(self):
+    def test_init_dispose_editable(self):
 
         obj = Model()
-        view = View(Item("value", editor=DropEditor()))
+        view = View(Item("value", editor=DropEditor(readonly=False)))
+        with store_exceptions_on_all_threads():
+                with create_ui(obj, dict(view=view)):
+                    pass
+                # Mutating value after UI is closed should be okay.
+                obj.value = "New"
+
+    def test_init_dispose_readonly(self):
+
+        obj = Model()
+        view = View(Item("value", editor=DropEditor(readonly=True)))
         with store_exceptions_on_all_threads():
                 with create_ui(obj, dict(view=view)):
                     pass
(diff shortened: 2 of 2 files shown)


traitsui/qt4/drop_editor.py
```diff
@@ -56,6 +57,15 @@ def init(self, parent):
 
         self.control._qt4_editor = self
 
+    def dispose(self):
+        """ Disposes of the content of an editor.
+        """
+        if self.factory.readonly:
+            # enthought/traitsui#884
+            _BaseEditor.dispose(self)
```
ievacerny, line 65: I wonder if it would be better to call `super(Editor, self).dispose()` without relying on what exactly that parent class is.

traitsui/tests/editors/test_drop_editor.py
```diff
@@ -0,0 +1,50 @@
...
+@skip_if_not_qt4
+class TestDropEditor(unittest.TestCase):
+    """ Test DropEditor. """
+
+    def test_init_dispose_editable(self):
+
+        obj = Model()
+        view = View(Item("value", editor=DropEditor(readonly=False)))
+        with store_exceptions_on_all_threads():
+                with create_ui(obj, dict(view=view)):
```
ievacerny, line 37: These two could be combined into one `with` statement and indentation is off here.

traitsui/qt4/drop_editor.py
kitchoi: If the design was right, we should have been able to just call `super().dispose()`. Unfortunately we can't. The `self.factory.readonly` branch in `init` does not use the `SimpleEditor.init`, so it would be wrong to call `super(Editor, self).dispose()` for this branch. If the `SimpleEditor.init` changes, its `SimpleEditor.dispose` will need to change, but that updated `SimpleEditor.dispose` may not be compatible with `DropEditor` any more.   I strongly doubt this editor should be subclassing the simple text editor. The purposes of these two editors look unrelated. It is possible that the subclassing was done "to reuse some code", causing unrelated components to be coupled. This problem is relevant for so so many editors and I'd in favour of decoupling them.

ievacerny: Sorry, my comment wasn't clear. I meant that maybe we should use `super(Editor, self).dispose()` to call the `dispose` of what is now imported as `_BaseEditor` (the parent class of `.text_editor.SimpleEditor` aka `Editor`). From here it doesn't matter what the parent class of `.text_editor.SimpleEditor` is as long as its `dispose` method is called so I think we could do without the assumption here that the parent class is `_BaseEditor`.

As long as the parent class of `.text_editor.SimpleEditor` doesn't change, the behaviour of both statements should be equivalent.

kitchoi: I am not sure I understand... This is not making assumption what the subclass of `traitsui.qt4.text_editor.SimpleEditor` is. The `DropEditor` should independently have subclassed `traitsui.qt4.editor.Editor` (which is base class for PyQt editors for Traits-based UIs) and have wanted to use call the `dispose` resolved on `traitsui.qt4.editor.Editor`. Currently `traitsui.qt4.editor.Editor.dispose` is not implemented, and it inherits the one from `traitsui.editor.Editor.dispose`. If `traitsui.qt4.editor.Editor.dispose` is later implemented, this particular branch in `DropEditor.dispose` would want to use that. What `traitsui.qt4.editor.text_editor.SimpleEditor.dispose` does should not matter because `traitsui.qt4.editor.text_editor.SimpleEditor.init` isn't used in this branch.

ievacerny: Oh, okay. I misunderstood the intentions of the code. I thought the goal was to simply avoid `traitsui.qt4.text_editor.SimpleEditor` and to stick with its inheritance structure, whatever it would be (and importing `_BaseEditor` just because it happens to be the next class in that structure).

Since the goal is to call `traitsui.qt4.editor.Editor.dispose()` independently, the code is fine as it is 👍

kitchoi: I see. I think I might have a comment in one of these PRs that says "skip this parent class ...". Maybe that was what caused the confusion. I may try to see if I can track that comment down and make it less confusing.

traitsui/tests/editors/test_drop_editor.py
ievacerny: Was wrong, they can't be combined into one. Thanks for fixing the indentation!

PR #897, opened 2020-06-11
Alternative fix to TabularAdapter crashes when column number reduces

kitchoi:
Many thanks to @FedeMiorelli tracking down the cause to issue #894.  The issue is indeed resolved by resetting the user widths.

This PR proposes resetting the user-defined widths in a slightly different place.

This PR actually does two things:
(1) Invalidating the view's `_user_widths` cache when the adapter columns is changed.
(2) Workaround an error that occurs when the UI is disposed. I believe the error is caused by header resize signal still being connected after the factory is removed from the editor. (See #431). The second test fails with this error if the `if self._editor.factory is not None` check is removed:
```
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/tests/editors/test_tabular_editor.py", line 402, in test_adapter_columns_changes_reduce_columns
    editor.adapter.columns = [("Name", "name")]
  File "/Users/kchoi/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/contextlib.py", line 88, in __exit__
    next(self.gen)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/tests/_tools.py", line 66, in store_exceptions_on_all_threads
    raise exceptions[0]
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/tabular_editor.py", line 929, in columnResized
    if (not self._editor.factory.auto_resize):
AttributeError: 'NoneType' object has no attribute 'auto_resize'
```

### DIFF
--- a/traitsui/qt4/tabular_editor.py
+++ b/traitsui/qt4/tabular_editor.py
@@ -247,7 +247,7 @@ class TabularEditor(Editor):
         # Rebuild the editor columns and headers whenever the adapter's
         # 'columns' changes:
         self.on_trait_change(
-            self.update_editor, "adapter.columns", dispatch="ui"
+            self._adapter_columns_updated, "adapter.columns", dispatch="ui"
         )
 
     def dispose(self):
@@ -266,7 +266,7 @@ class TabularEditor(Editor):
             self.refresh_editor, "adapter.+update", remove=True
         )
         self.on_trait_change(
-            self.update_editor, "adapter.columns", remove=True
+            self._adapter_columns_updated, "adapter.columns", remove=True
         )
 
         self.adapter.cleanup()
@@ -391,6 +391,19 @@ class TabularEditor(Editor):
         """
         if not self.factory.multi_select:
             self.selected_column = self.column_clicked.column
+
+    def _adapter_columns_updated(self):
+        """ Update the view when the adapter columns trait changes.
+        Note that this change handler is added after the UI is instantiated,
+        and removed when the UI is disposed.
+        """
+        # Invalidate internal state of the view related to the columns
+        n_columns = len(self.adapter.columns)
+        if (self.control is not None
+                and self.control._user_widths is not None
+                and len(self.control._user_widths) != n_columns):
+            self.control._user_widths = None
+        self.update_editor()
 
     def _update_changed(self):
         self.update_editor()
--- a/traitsui/qt4/tabular_editor.py
+++ b/traitsui/qt4/tabular_editor.py
@@ -40,9 +40,9 @@ from traits.api import (
     Property,
     TraitListEvent,
 )
+
 from traitsui.tabular_adapter import TabularAdapter
 from traitsui.helper import compute_column_widths
-from traitsui.qt4 import toolkit
 from .editor import Editor
 from .tabular_model import TabularModel
 
@@ -270,8 +270,6 @@ class TabularEditor(Editor):
         )
 
         self.adapter.cleanup()
-
-        toolkit.destroy_control(self.control)
 
         super(TabularEditor, self).dispose()
 
@@ -928,7 +926,8 @@ class _TableView(QtGui.QTableView):
             if self._user_widths is None:
                 self._user_widths = [None] * len(self._editor.adapter.columns)
             self._user_widths[index] = new
-            if not self._editor.factory.auto_resize:
+            if (self._editor.factory is not None
+                    and not self._editor.factory.auto_resize):
                 self.resizeColumnsToContents()
 
     @contextmanager
--- a/traitsui/qt4/tabular_editor.py
+++ b/traitsui/qt4/tabular_editor.py
@@ -926,7 +926,8 @@ class _TableView(QtGui.QTableView):
             if self._user_widths is None:
                 self._user_widths = [None] * len(self._editor.adapter.columns)
             self._user_widths[index] = new
-            if not self._editor.factory.auto_resize:
+            if (self._editor.factory is not None
+                    and not self._editor.factory.auto_resize):
                 self.resizeColumnsToContents()
 
     @contextmanager
--- a/traitsui/qt4/tabular_editor.py
+++ b/traitsui/qt4/tabular_editor.py
@@ -40,9 +40,9 @@ from traits.api import (
     Property,
     TraitListEvent,
 )
-
 from traitsui.tabular_adapter import TabularAdapter
 from traitsui.helper import compute_column_widths
+from traitsui.qt4 import toolkit
 from .editor import Editor
 from .tabular_model import TabularModel
 
@@ -270,6 +270,8 @@ class TabularEditor(Editor):
         )
 
         self.adapter.cleanup()
+
+        toolkit.destroy_control(self.control)
 
         super(TabularEditor, self).dispose()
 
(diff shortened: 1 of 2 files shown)


kitchoi:
Thank you @FedeMiorelli for confirming that this branch resolves the issue (see https://github.com/enthought/traitsui/pull/895#issuecomment-643104961).

If we do merge this PR, I'd like to propose including @FedeMiorelli in the merge commit as a coauthor, because this fix would not have been here had it not because of @FedeMiorelli's initial investigation and proposed solution. @FedeMiorelli, would that be okay with you?

FedeMiorelli:
@kitchoi Sure, happy to help!

kitchoi:
Thank you @FedeMiorelli!

traitsui/qt4/tabular_editor.py
```diff
@@ -913,7 +926,8 @@ def columnResized(self, index, old, new):
             if self._user_widths is None:
                 self._user_widths = [None] * len(self._editor.adapter.columns)
             self._user_widths[index] = new
-            if not self._editor.factory.auto_resize:
+            if (self._editor.factory is not None
```
kitchoi, line 929: Actually I can fix this properly than working around it.

traitsui/qt4/tabular_editor.py
```diff
@@ -266,11 +266,13 @@ def dispose(self):
             self.refresh_editor, "adapter.+update", remove=True
         )
         self.on_trait_change(
-            self.update_editor, "adapter.columns", remove=True
+            self._adapter_columns_updated, "adapter.columns", remove=True
         )
 
         self.adapter.cleanup()
 
+        toolkit.destroy_control(self.control)
```
kitchoi, line 274: Backtracking on this: I need to think about harder about why this fixes the error. It is strange that we should need to call `destroy_control` here, because I'd expect the UI to do this for all the editors, and in fact, `traitsui.ui.UI.finish` calls this.  I will resurrect the workaround because that is a less risky thing to do, and would make it easier for this PR to be backported to 7.0.1 bug fix release.

traitsui/qt4/tabular_editor.py
kitchoi: Decided not to eventually so this PR is more trivial and incur lower risk for it to be backported to the bug fix release.

kitchoi, line 929: @FedeMiorelli This `if self._editor.factory is not None` check was also made by you in #895. May I ask if this is also needed in the production environment, e.g. running a desktop application manually? Or is it only for tests to pass?  I am trying to understand if there is a more fundamental problem lurking behind. Thank you very much!

FedeMiorelli: @kitchoi I had added this check because I'd get a crash when closing the dialog, so while in production desktop application use.

kitchoi: @FedeMiorelli Thank you very much. It is good to know this is affecting production use. This definitely worth some investigation. I will look into this.

kitchoi: I tried removing this check and then tested this branch against a few simple examples, removing columns and resizing them, and then closing the dialog, but I still could not reproduce the crash.

@FedeMiorelli if you have a simple example that can produce the crash specific to this check, and if you could provide us with the example, that would be very helpful. Without a way to reproduce this in a production environment, I am unable to investigate further nor can I open a separate issue for this. 

The fact that this check is needed in a production environment indicates some deeper issues. However it is not clear to me at this point whether this is an issue in TraitsUI or somewhere else.

FedeMiorelli: @kitchoi I have created a minimal case to reproduce the crash when columns are removed, which occurs if the user resizes some column. In this simplified code I have not been able to repeat the other crash however, maybe it was triggered by some other component in the much more complex code this was embedded in.  Sorry if this is not helping!

By the way maybe I could ask a small advice here: you can see I am hacking the TabularAdapter to add a reference to the object (line 15) which i then use to access the column names property. I'm sure there is a smarter / more correct way to do so. I thought I could access the self.object field in the TabularAdapter but it is always None for me.
```

import os, sys
import numpy as np
from traits.api import HasStrictTraits, Int, Property, cached_property, Any
from traitsui.api import View, Group, HGroup, Item, EnumEditor, TextEditor, RangeEditor, TabularEditor, TabularAdapter



class DataTableAdapter ( TabularAdapter ):
    columns      = Property(depends_on='parentObject.DataTableColumns')
    parentObject = Any
    font         = 'Consolas 9'
    alignment    = 'left'
    
    def len ( self, object, trait ):
        self.parentObject = object
        if object is not None and object.DataTable is not None:
            return len(object.DataTable)
        else:
            return 0
    
    @cached_property
    def _get_columns ( self ):
        if self.parentObject is None:
            return []
        else:
            return self.parentObject.DataTableColumns

        
class TestClass(HasStrictTraits):

    NCols = Int(1, label='Columns')
    DataTable = Property(depends_on='NCols')
    DataTableColumns = Property(depends_on='DataTable')

    @cached_property
    def _get_DataTable(self):
        rows = []
        for i in range(4):
            row = np.random.rand(self.NCols)
            rows.append(row)
        
        return rows

    @cached_property
    def _get_DataTableColumns(self):
        if len(self.DataTable) == 0:
            return []
        return [str(i+1) for i in range(len(self.DataTable[0]))]

    traits_view = View(Group(Item('NCols', editor=RangeEditor(low=0, high=10, mode='spinner')),
                                    HGroup(Item('DataTable', editor=TabularEditor(adapter=DataTableAdapter(), drag_move=False, stretch_last_section=False, editable=False, operations=[]),
                                           show_label=False, springy=True), visible_when='len(DataTable)>0'),
                        ),
                         
                   resizable=True,
                   kind='livemodal',
                   width=800, height=300,
                   buttons = [ 'OK', 'Cancel' ],
                  )


if __name__=='__main__':
    TestClass().configure_traits()
```

kitchoi: @FedeMiorelli Thank you very much for the example! It is really useful to see how others use the adapter and to know that the attribute error crash is not easily reproducible.

As for the issue with having to create a reference from the adapter back to the object, I wonder if the use of `default_traits_view` is useful here. Instead of defining a static `traits_view`, you can define a view dynamically with a method. Then the `TestClass` can own the adapter and update its columns directly. e.g.:
```
# I added these imports:
from traits.api import Instance, on_trait_change

class DataTableAdapter ( TabularAdapter ):
    font         = 'Consolas 9'
    alignment    = 'left'

    def len ( self, object, trait ):
        if object is not None and object.DataTable is not None:
            return len(object.DataTable)
        else:
            return 0


class TestClass(HasStrictTraits):

    NCols = Int(1, label='Columns')
    DataTable = Property(depends_on='NCols')

    adapter = Instance(DataTableAdapter, kw=dict(columns=["1"]))

    @cached_property
    def _get_DataTable(self):
        rows = []
        for i in range(4):
            row = np.random.rand(self.NCols)
            rows.append(row)

        return rows

    @on_trait_change("NCols")
    def _update_adapter_columns(self):
        self.adapter.columns = [str(i + 1) for i in range(self.NCols)]

        adapter = self.adapter
        return View(Group(Item('NCols', editor=RangeEditor(low=0, high=10, mode='spinner')),
                        HGroup(Item('DataTable', editor=TabularEditor(adapter=adapter, drag_move=False, stretch_last_section=False, editable=False, operations=[]),
                                show_label=False, springy=True), visible_when='len(DataTable)>0'),
            ),
            kind="livemodal",
            resizable=True,
            width=800, height=300,
            buttons = [ 'OK', 'Cancel' ],
        )
```

FedeMiorelli: Thanks a lot @kitchoi I learned something new!

kitchoi: @FedeMiorelli My pleasure!  I hope this helps and thanks again for looking into this!

PR #900, opened 2020-06-16
Fix unexpected format_func in RangeEditor bug

### LINKED ISSUE #898: format_func may be unexpected by some editors
Currently the `RangeEditor_demo.py` fails to launch, with the following error:
```
  File "examples/demo/Standard_Editors/RangeEditor_demo.py", line 97, in <module>
    demo.configure_traits()
  File "/Users/kchoi/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/traits/has_traits.py", line 2088, in configure_traits
    args,
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/toolkit.py", line 232, in view_application
    context, view, kind, handler, id, scrollable, args
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/view_application.py", line 84, in view_application
    context, view, kind, handler, id, scrollable, args
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/view_application.py", line 126, in __init__
    args=self.args,
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/view.py", line 462, in ui
    ui.ui(parent, kind)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/ui.py", line 246, in ui
    self.rebuild(self, parent)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/toolkit.py", line 162, in ui_live
    ui_live.ui_live(ui, parent)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/ui_live.py", line 43, in ui_live
    _ui_dialog(ui, parent, BaseDialog.NONMODAL)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/ui_live.py", line 65, in _ui_dialog
    BaseDialog.display_ui(ui, parent, style)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/ui_base.py", line 284, in display_ui
    ui.owner.init(ui, parent, style)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/ui_live.py", line 225, in init
    self.add_contents(panel(ui), bbox)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/ui_panel.py", line 262, in panel
    _fill_panel(panel, content, ui)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/ui_panel.py", line 295, in _fill_panel
    gp = _GroupPanel(item, ui, suppress_label=True)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/ui_panel.py", line 601, in __init__
    layout = self._add_items(content, inner)
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/qt4/ui_panel.py", line 854, in _add_items
    ui, object, name, item.tooltip, None
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/editors/range_editor.py", line 305, in custom_editor
    ui, object, name, description, parent
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/editor_factory.py", line 155, in custom_editor
    invalid_trait_name=self.invalid,
TypeError: CustomEnumEditor() got an unexpected keyword argument 'format_func'
```

Looks related to https://github.com/enthought/traitsui/pull/859

The example seems to work fine with v7.0.0.

ievacerny:
Closes #898

`SimpleEnumEditor` and `CustomEnumEditor` in `range_editor.py` (both qt and wx) are not subclasses of `Editor` themselves. Instead they are functions that return an `EnumEditor` via `traitsui.editors.enum_editor.ToolkitEditorFactory`.

The PR modifies the signatures of these functions to match what `EditorFactory` expects (adding `format_func`, `format_str` and `invalid_trait_name`). These additional arguments aren't passed to the actual editor to match the behaviour as it was before the bug.

### DIFF
--- a/traitsui/qt4/range_editor.py
+++ b/traitsui/qt4/range_editor.py
@@ -720,8 +720,7 @@ class RangeTextEditor(TextEditor):
 
 
 def SimpleEnumEditor(
-    parent, factory, ui, object, name, description, format_func=None,
-    format_str='', invalid_trait_name=''
+    parent, factory, ui, object, name, description, **kwargs
 ):
     return CustomEnumEditor(
         parent, factory, ui, object, name, description, "simple"
@@ -729,8 +728,7 @@ def SimpleEnumEditor(
 
 
 def CustomEnumEditor(
-    parent, factory, ui, object, name, description, style="custom",
-    format_func=None, format_str='', invalid_trait_name=''
+    parent, factory, ui, object, name, description, style="custom", **kwargs
 ):
     """ Factory adapter that returns a enumeration editor of the specified
     style.
--- a/traitsui/qt4/range_editor.py
+++ b/traitsui/qt4/range_editor.py
@@ -719,14 +719,18 @@ class RangeTextEditor(TextEditor):
 # -------------------------------------------------------------------------
 
 
-def SimpleEnumEditor(parent, factory, ui, object, name, description):
+def SimpleEnumEditor(
+    parent, factory, ui, object, name, description, format_func=None,
+    format_str='', invalid_trait_name=''
+):
     return CustomEnumEditor(
         parent, factory, ui, object, name, description, "simple"
     )
 
 
 def CustomEnumEditor(
-    parent, factory, ui, object, name, description, style="custom"
+    parent, factory, ui, object, name, description, style="custom",
+    format_func=None, format_str='', invalid_trait_name=''
 ):
     """ Factory adapter that returns a enumeration editor of the specified
     style.
--- a/traitsui/tests/editors/test_range_editor.py
+++ b/traitsui/tests/editors/test_range_editor.py
@@ -1,0 +1,50 @@
+import unittest
+
+from traits.api import HasTraits, Int
+from traitsui.api import RangeEditor, UItem, View
+from traitsui.tests._tools import (
+    create_ui,
+    skip_if_null,
+    store_exceptions_on_all_threads,
+)
+
+
+class RangeModel(HasTraits):
+
+    value = Int()
+
+
+@skip_if_null
+class TestRangeEditor(unittest.TestCase):
+
+    def check_range_enum_editor_format_func(self, style):
+        # RangeEditor with enum mode doesn't support format_func
+        obj = RangeModel()
+        view = View(
+            UItem(
+                "value",
+                editor=RangeEditor(
+                    low=1, high=3,
+                    format_func=lambda v: "{:02d}".format(v),
+                    mode="enum"
+                ),
+                style=style,
+            )
+        )
+
+        with store_exceptions_on_all_threads(),\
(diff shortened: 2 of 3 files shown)


traitsui/qt4/range_editor.py
```diff
@@ -719,14 +719,18 @@ def update_object(self):
 # -------------------------------------------------------------------------
 
 
-def SimpleEnumEditor(parent, factory, ui, object, name, description):
+def SimpleEnumEditor(
+    parent, factory, ui, object, name, description, format_func=None,
+    format_str='', invalid_trait_name=''
+):
```
kitchoi, line 725: Perhaps we should use `**kwargs` here. I am not a fan of using `**kwargs` myself but this function is trying to imitate the instantiation of a subclass of `HasTraits`, so it makes sense for the function signature to match.

ievacerny: I'm a bit worried that `**kwargs` might end up masking potential future bugs. I've added a couple of tests, so now similar problems of signatures not matching (if not using `kwargs`) should be caught early. And given that these "fake" editors are quite hidden, maybe it's better for things to break early rather than accepting everything quietly?

kitchoi: If I was to re-design these factories, I would not use `**kwargs` for exact the reason that is too lenient. Back to what we have here...

These functions are trying to duck type a subclass of HasTraits, here:
https://github.com/enthought/traitsui/blob/ab90190dded4c70b2ba39e9323b6ef424c926dd4/traitsui/editors/range_editor.py#L237-L238

When it gets to this point:
https://github.com/enthought/traitsui/blob/0881435aed3d8390a6fd7cf795b2f6df679b9085/traitsui/editor_factory.py#L146
One assumes a subclass of `HasTraits`. The implicit interface there is therefore the signature of `(*args, **kwargs)`.  Not accepting `**kwargs` broke that implicit contract consistent across the code base.

kitchoi: While we are at this, it would safe to add `*` before `factory`. This is a new feature of Python 3 that we can now use because Python 2 is dropped. Yay.
Only the `parent` is expected to be passed onto `init`:
https://github.com/enthought/traitsui/blob/0881435aed3d8390a6fd7cf795b2f6df679b9085/traitsui/editor_factory.py#L91-L95

kitchoi: Except... I don't know if this `SimpleEnumEditor` is used externally outside of the `editor_factory`. Adding `*` would be backward incompatible. Peril of having all the modules and classes with public names... 
I am sorry. Please ignore the `*` suggestion. I will open a separate issue to confirm. That is less important than the bug fix.

Issue #854, opened 2020-05-29
Linux qt test interactions: remaining connection to old ImageEnumModel instance


ievacerny:
I've been seeing some test interactions where `data` method of an old `ImageEnumModel` instance is called in a different test (and fails because `self._editor.factory` is `None`) https://travis-ci.org/github/enthought/traitsui/jobs/692588809

I don't know what exactly isn't being cleaned up and what is the exact connection that remains. I'm failing to reproduce this locally on a Linux VM so I can't debug this any further, but it's happening consistently on Linux CI.

Just for reference, there is a similar issue for wx test interactions on windows (https://github.com/enthought/traitsui/issues/851)

kitchoi:
Thank you for the issue!

Here is the test failure just in case Travis-CI causes us to lose the log:
```
======================================================================
ERROR: test_simple_editor_more_cols (traitsui.tests.editors.test_image_enum_editor.TestSimpleImageEnumEditor)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/.edm/envs/traitsui-test-3.6-pyside2/lib/python3.6/site-packages/traitsui/tests/editors/test_image_enum_editor.py", line 314, in test_simple_editor_more_cols
    self.setup_gui(enum_edit, view)
  File "/home/travis/.edm/envs/traitsui-test-3.6-pyside2/lib/python3.6/contextlib.py", line 88, in __exit__
    next(self.gen)
  File "/home/travis/.edm/envs/traitsui-test-3.6-pyside2/lib/python3.6/site-packages/traitsui/tests/_tools.py", line 66, in store_exceptions_on_all_threads
    raise exceptions[0]
  File "/home/travis/.edm/envs/traitsui-test-3.6-pyside2/lib/python3.6/site-packages/traitsui/qt4/image_enum_editor.py", line 301, in data
    index = mi.row() * self._editor.factory.cols + mi.column()
AttributeError: 'NoneType' object has no attribute 'cols'
```
(Job was run on this commit: 0620721122f7767bf43ea512b9f4e738753bab6b)

kitchoi:
FTR, this [commit](https://github.com/enthought/traitsui/pull/865/commits/7466c03f9cabb377d63ea59845224d0a111312cb) tries to remove the skips after ensuring all `UI.dispose` are called in other tests (assuming I had got that right). That did not help with this issue.

ievacerny:
[...]

PR #902, opened 2020-06-17
Add process_cascade_events to the internal test tools

kitchoi:
Attribute errors such as https://github.com/enthought/traitsui/issues/854 are caused by events being processed after `UI.dispose` is called. Because various UI attributes are removed in `dispose`, the processing of these left-over events will cause errors.

This issue applies to assertions in the test body as well, which is why tests will typically call `GUI.process_events` prior to checking states. For more complex UIs, an event could easily cause a cascade of new events. For most testing purposes, we want to simulate the production environment where the GUI event loop is continuously running and we normally want all those cascade of events to be processed before we do assertions or dispose the UI.

The wx implementation of `pyface.gui.GUI.process_events` will process all cascade events, so it is enough to call `pyface.gui.GUI.process_events` just once before assertions and in between cleanup actions.

For Qt, the implementation of `pyface.gui.GUI.process_events` calls `QtCore.QCoreApplication.processEvents` which only processes the events that are in the queue when the function is called, excluding events that are posted by those events.

This PR introduces a function called `process_cascade_events` in the internal testing tools module so that we can process all cascaded events in Qt (and Wx). I will apply this function to all existing tests (in fact, I will open a draft PR that depends on this one to experiment the outcome with CI).

This function should probably be added to `pyface.gui.GUI`, once this is reviewed, I will open a separate PR to `pyface`, but it will be a feature there, rather than an internal utility function.


traitsui/tests/_tools.py
```diff
@@ -131,6 +133,37 @@ def filter_tests(test_suite, exclusion_pattern):
...
+        # if there are pending events (`hasPendingEvents` is deprecated).
+        # The minus-one is to account for precision differences between Python
+        # and Qt, false positive triggers another redundant run that should
+        # return immediately so that is fine.
+        while start is None or duration_millisecs >= _TOLERANCE_MILLISECS - 1:
+            start = time.time()
+            QtCore.QCoreApplication.processEvents(
+                QtCore.QEventLoop.AllEvents, _TOLERANCE_MILLISECS
+            )
+            duration_millisecs = (time.time() - start) * 1000
```
kitchoi, line 162: Yep this is ugly and I'd be very happy if that is a neater solution.

kitchoi: There is `flush` but Qt documentation says it is deprecated. https://doc.qt.io/qt-5/qcoreapplication-obsolete.html#flush

traitsui/tests/test_tools.py
```diff
@@ -0,0 +1,131 @@
...
+        self.assertEqual(q_object.n_events, max_n_events)
+
+    @skip_if_not_wx
+    def test_wx_process_events_process_all(self):
+
+        def cleanup(wx_handler):
+            GUI.process_events()
+            wx_handler.Destroy()
+
+        wx_handler = DummyWxHandler(max_n_events=10)
```
ievacerny, line 122: Nit: could we have a local variable `max_n_events` as in `test_qt_process_events_process_all`?

traitsui/tests/test_tools.py
```diff
@@ -0,0 +1,131 @@
...
+
+        # then
+        self.assertEqual(q_object.n_events, max_n_events)
+
+    @skip_if_not_wx
+    def test_wx_process_events_process_all(self):
+
+        def cleanup(wx_handler):
+            GUI.process_events()
+            wx_handler.Destroy()
```
ievacerny, line 120: Is the `GUI.process_events()` here in case of a test failure?

traitsui/tests/_tools.py
```diff
@@ -131,6 +133,37 @@ def filter_tests(test_suite, exclusion_pattern):
...
+        from pyface.qt import QtCore
+        start = None
+
+        # Qt won't raise if there are still events to be processed before
+        # the time limit is reached. There are no other safe way to tell
+        # if there are pending events (`hasPendingEvents` is deprecated).
+        # The minus-one is to account for precision differences between Python
+        # and Qt, false positive triggers another redundant run that should
+        # return immediately so that is fine.
+        while start is None or duration_millisecs >= _TOLERANCE_MILLISECS - 1:
```
ievacerny, line 157: 🤖 flake8 is complaining about undefined variable `duration_millisecs `

traitsui/tests/test_tools.py
```diff
@@ -0,0 +1,131 @@
+#  Copyright (c) 2008-20, Enthought, Inc.
```
ievacerny, line 1: Nit: shouldn't this file be named `test__tools.py`?

traitsui/tests/test_tools.py
kitchoi: Yep. I will add a comment here to explain that.

PR #903, opened 2020-06-17
TST: Process all cascade events in tests

### LINKED ISSUE #851: Windows wx test interactions: remaining connection to old TextEditor instance
I found this while investigating errors on the following windows CI job https://ci.appveyor.com/project/EnthoughtOSS/traitsui/builds/33168227/job/j10vrs43kwn1qb0i

The root of the cause comes from these two tests and the error can be replicated when only the two of them are run together:
```
test_text_check_list_object_list (traitsui.tests.editors.test_check_list_editor.TestTextCheckListEditor) ... ok
test_text_check_list_object_str (traitsui.tests.editors.test_check_list_editor.TestTextCheckListEditor) ... Uncaught exception:
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\wx\check_list_editor.py", line 263, in update_object
    self.value = value
UnboundLocalError: local variable 'value' referenced before assignment
ERROR
```

The first test passes fine as there were no other check list `TextEditors` before it. In the second test, the error is raised during the call https://github.com/enthought/traitsui/blob/64858b8c9ce9189a8c9cfde60610e4bd12f30962/traitsui/wx/ui_base.py#L98
because the `update_object` method of the check list `TextEditor` from the **previous** test is called. The control of that object is already removed (it is `None`), therefore, value assignment fails and this `UnboundLocalError` is raised.

I am not entirely sure what exactly the remaining connection to the old check list `TextEditor` is and why this is only happening on windows, but it seems to be related to the binding of https://github.com/enthought/traitsui/blob/64858b8c9ce9189a8c9cfde60610e4bd12f30962/traitsui/wx/editor_factory.py#L115
Adding a `dispose` method to the editor that unbinds this specific event seems to fix the issue.

kitchoi:
This is a draft PR as it has dependency on #902. This perhaps gives some context for how the function added in #902 could be used.

This PR:
- Include the changes in #902 (will be removed once that is merged first)
- Apply the `process_cascade_events` function to all tests in TraitsUI.
 
Opened to test behaviour on CI.

Closes #851 as the workaround can now be removed and the test that used to fail without the workaround is still passing.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -213,7 +213,7 @@ def test(runtime, toolkit, environment):
         environ["EXCLUDE_TESTS"] = "(wx|qt)"
 
     commands = [
-        "edm run -e {environment} -- coverage run -p -m unittest discover -v traitsui"
+        "edm run -e {environment} -- coverage run -p -m unittest traitsui.tests.editors.test_image_enum_editor.TestSimpleImageEnumEditor.test_simple_editor_combobox"
     ]
 
     # We run in a tempdir to avoid accidentally picking up wrong traitsui
--- a/traitsui/qt4/boolean_editor.py
+++ b/traitsui/qt4/boolean_editor.py
@@ -43,6 +43,11 @@ class SimpleEditor(Editor):
         self.control.stateChanged.connect(self.update_object)
         self.set_tooltip()
 
+    def dispose(self):
+        if self.control is not None:
+            self.control.stateChanged.disconnect(self.update_object)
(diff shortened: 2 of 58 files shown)


kitchoi:
Closing... as this has slowly evolved into a branch does too many things for experimenting test outcome with CI. I will open separate PRs that are more focused. Sorry for the noise.

Issue #907, opened 2020-06-18
Crash when using UI.dispose() with a TableEditor nested in a ListEditor


xamcost:
Hi there !

I ran into a python crash upon calling `traitsui.ui.UI.dispose()`, which happens when a `ListEditor` is using to display a list of `HasTraits` objects featuring a view using `TableEditor`. This occurs with qt4 and qt5. I used traits 6.0 and traitsui 7. The snippet below reproduces this crash:

```Python3
from traits.api import List, HasTraits, Str
from traitsui.api import ListEditor, TableEditor, UItem, View
from traitsui.table_column import ObjectColumn


class Row(HasTraits):
    col_ = Str()


class ViewWithTableEditor(HasTraits):

    table = List([Row(col_0="1")])

    def traits_view(self):
        view = View(
            UItem(
                'table',
                editor=TableEditor(
                    columns=[ObjectColumn(name='col_0')],
                )
            )
        )
        return view


class MainView(HasTraits):

    tabs = List([ViewWithTableEditor(), ViewWithTableEditor()])

    def traits_view(self):
        view = View(
                UItem('tabs',
                      editor=ListEditor(
                          use_notebook=True,
                      ),
                      style='custom'),
            )
        return view


view = MainView()
ui = view.edit_traits()
ui.dispose()
```
It leads to the following error message:

```
Traceback (most recent call last):
  File "/.edm/envs/msp-dev-3.6/lib/python3.6/site-packages/traitsui/qt4/table_editor.py", line 1059, in resizeEvent
    self.resizeColumnsToContents()
  File "/.edm/envs/msp-dev-3.6/lib/python3.6/site-packages/traitsui/qt4/table_editor.py", line 1176, in resizeColumnsToContents
    width = max(base_width, self.sizeHintForColumn(column_index))
  File "/.edm/envs/msp-dev-3.6/lib/python3.6/site-packages/traitsui/qt4/table_editor.py", line 1127, in sizeHintForColumn
    if editor.factory.sortable and not editor.factory.reorderable:
AttributeError: 'NoneType' object has no attribute 'sortable'
```

Any ideas what is happening ? Thank you in advance for your help !

xamcost:
On another aspect, it is interesting to note that testing an instance of the above `MainView` class with pyface's `GuiTestAssistant` leads to different behaviors depending on the Qt version used. Consider the following code:

```Python3
from unittest import TestCase

[...]

PR #910, opened 2020-06-18
[WIP] Unset editor attribute later using the GUI event queue

kitchoi:
The many errors we have seen in the past relate to getting an attribute error because a GUI event is processed after several editor's attributes are set to None in in `dispose`.

This PR attempts to fix this by asking the GUI event loop to unset the attributes in `dispose`, i.e. delay unsetting the attributes.

### DIFF
--- a/traitsui/editor.py
+++ b/traitsui/editor.py
@@ -20,6 +20,8 @@
 from contextlib import contextmanager
 from functools import partial
 
+from pyface.api import GUI
+
 from traits.api import (
     Any,
     Bool,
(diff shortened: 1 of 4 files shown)


kitchoi:
This was a preliminary test to see if unsetting the traits later might have catastrophic outcome. While the tests did not show such outcome, it does not mean there won't be, but at least this seems to be an avenue worth pursuing further. More investigation is needed to confirm this is indeed the right thing to do. Closing for now.

kitchoi:
FTR, unsetting the traits later seem to resolve the error seen in #907

PR #914, opened 2020-06-22
Add cron job to install/test against packages from source

rahulporuri:
At the moment, traits is the only package installed from source as pyface is always installed from source on CI. The changes are almost the same as that in traits PR https://github.com/enthought/traits/pull/479/


rahulporuri:
Note that there's already a weekly cron job for traitsui - it just behaves differently now.

etstool.py
```diff
@@ -133,6 +138,9 @@
 }
 
 
+github_url_fmt = "git+http://github.com/enthought/{0}.git#egg={0}"
```
kitchoi, line 141: This is not used.

rahulporuri: corrected.

rahulporuri:
thanks for the review @kitchoi

.travis.yml
```diff
@@ -42,6 +42,7 @@ before_install:
   - edm install -y wheel click coverage
 install:
   - for toolkit in ${TOOLKITS}; do edm run -- python etstool.py install --runtime=${RUNTIME} --toolkit=${toolkit} || exit; done
+  - if [[ ${TRAVIS_EVENT_TYPE} == 'cron' ]]; then for toolkit in ${TOOLKITS}; do edm run -- python etstool.py install --runtime=${RUNTIME} --toolkit=${toolkit} --source || exit; done; fi
```
kitchoi, line 45: I only realized this later that this causes `install` to be called twice on the cronjob.
Options:
(1) in `etstool.py`, set the default of `--source` based on some environment variable, e.g. `default=os.environ.get("IS_CRON", False)`
(2) compose the option flags in the `travis.yml` based on the environment variable..., e.g. `if [[${TRAVIS_EVENT_TYPE} == 'cron' ]] FLAGS="--source"` something like that.

rahulporuri: yes. sorry. i wasn't trying to optimize CI time. I'll look into this this week (before i start adding more cron jobs which are wasteful).

PR #915, opened 2020-06-24
Fix Theme pickling

### LINKED ISSUE #825: Pickling Theme errors on v7.0.0
We recently upgraded traitsui: `6.1.3-5  -->  7.0.0-1`. The following code fails on the newer version but runs fine on the older:

```
>>> import pickle
>>> from traitsui.theme import Theme
>>> theme = Theme()
>>> pickle.dumps(theme)
  File "<stdin>", line 1, in <module>
  File "...traits/has_traits.py", line 1230, in __reduce_ex__
    return (__newobj__, (self.__class__,), self.__getstate__())
  File "...traits/has_traits.py", line 1194, in __getstate__
    result = self.trait_get(transient=is_none)
  File "...traits/has_traits.py", line 1298, in trait_get
    value = getattr(self, name, missing)
  File "...traitsui/theme.py", line 69, in _get_content_color
    import wx
ModuleNotFoundError: No module named 'wx'
```

#### Full context

We have a tree editor for which some nodes have color icons. The color nodes specifically fail to be draggable (into an IPython shell for example) because of the pickling failure above.

ievacerny:
Closes #825 

PR adds toolkit checks before wx imports

### DIFF
--- a/traitsui/theme.py
+++ b/traitsui/theme.py
@@ -76,7 +76,7 @@ class Theme(HasPrivateTraits):
 
                 self._content_color = color
 
-            return self._content_color
+        return self._content_color
 
     def _set_content_color(self, color):
         self._content_color = color
@@ -93,18 +93,18 @@ class Theme(HasPrivateTraits):
 
                 self._label_color = color
 
-            return self._label_color
+        return self._label_color
 
     def _set_label_color(self, color):
         self._label_color = color
 
     @cached_property
     def _get_image_slice(self):
+        if self.image is None:
+            return None
+
         if ETSConfig.toolkit == "wx":
             from traitsui.wx.image_slice import image_slice_for
-
-            if self.image is None:
-                return None
 
             return image_slice_for(self.image)
 
--- a/traitsui/theme.py
+++ b/traitsui/theme.py
@@ -18,9 +18,8 @@
 """ Defines 'theme' related classes.
 """
 
-
-
 from traits.api import HasPrivateTraits, Property, cached_property
+from traits.etsconfig.api import ETSConfig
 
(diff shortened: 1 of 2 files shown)


traitsui/theme.py
```diff
@@ -66,45 +65,48 @@ def __init__(self, image=None, **traits):
...
+                color = wx.BLACK
+                islice = self.image_slice
+                if islice is not None:
+                    color = islice.content_color
 
-            self._content_color = color
+                self._content_color = color
 
-        return self._content_color
+            return self._content_color
```
kitchoi, line 79: This `return self._content_color` should be dedented so that even if wx is not available, the value given by the setter can be retrieved via the getter.  Maybe worth adding a test for this?

traitsui/theme.py
```diff
@@ -66,45 +65,48 @@ def __init__(self, image=None, **traits):
...
+                color = wx.BLACK
+                islice = self.image_slice
+                if islice is not None:
+                    color = islice.label_color
 
-            self._label_color = color
+                self._label_color = color
 
-        return self._label_color
+            return self._label_color
```
kitchoi, line 96: Same here.

traitsui/theme.py
```diff
@@ -66,45 +65,48 @@ def __init__(self, image=None, **traits):
...
     @cached_property
     def _get_image_slice(self):
-        from traitsui.wx.image_slice import image_slice_for
+        if ETSConfig.toolkit == "wx":
+            from traitsui.wx.image_slice import image_slice_for
 
-        if self.image is None:
-            return None
+            if self.image is None:
+                return None
```
kitchoi, line 107: This two lines should be up and out of the wx block?

Issue #920, opened 2020-06-25
Unable to test with TableEditor with wx on Appveyor


kitchoi:
In two of the recent PRs (https://github.com/enthought/traitsui/pull/915, https://github.com/enthought/traitsui/pull/919), we started seeing this in the Appveyor build with wx:

```
======================================================================
ERROR: test_filtered_table_editor (traitsui.tests.editors.test_table_editor.TestTableEditor)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_table_editor.py", line 288, in test_filtered_table_editor
    filter = ui.get_editors("values")[0].filter
AttributeError: 'SimpleEditor' object has no attribute 'filter'
======================================================================
ERROR: test_table_editor_select_cell (traitsui.tests.editors.test_table_editor.TestTableEditor)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_table_editor.py", line 489, in test_table_editor_select_cell
    selected = editor.selected_cell
AttributeError: 'SimpleEditor' object has no attribute 'selected_cell'
======================================================================
ERROR: test_table_editor_select_cell_index (traitsui.tests.editors.test_table_editor.TestTableEditor)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_table_editor.py", line 542, in test_table_editor_select_cell_index
    selected = editor.selected_cell_index
AttributeError: 'SimpleEditor' object has no attribute 'selected_cell_index'
======================================================================

... (a couple similar ones)
```

The error looks like a symptom of a failure to import `traitsui.wx.table_editor` (see #864)

Unfortunately the build on Appveyor for wx is allowed to fail, which makes it harder to detect this sort of things early.

kitchoi:
This may be the earliest commit where this started to happen: https://github.com/enthought/traitsui/pull/853/commits/717c9d3bf6928b33685ff6a6cb1ad1e0d38355b2 
Keep digging...

kitchoi:
[...]

Issue #864, opened 2020-06-02
Editor base class is used as fallback when import fails, leading to obscure errors


kitchoi:
The tests in `traitsui.wx.tests.test_table_editor` (when run on their own, not with the entire test suite) produce an error like this:
```
======================================================================
ERROR: test_filtered_table_editor (traitsui.tests.editors.test_table_editor.TestTableEditor)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/tests/editors/test_table_editor.py", line 288, in test_filtered_table_editor
    filter = ui.get_editors("values")[0].filter
AttributeError: 'SimpleEditor' object has no attribute 'filter'
```

The `SimpleEditor` there is actually an instance of `traitsui.wx.editor_factory.SimpleEditor`, where we expected `traitsui.wx.table_editor.TableEditor`.
Somehow, the failure to import `traitsui.wx.table_editor` (see #863) has led to the `SimpleEditor` from `editor_factory` to be used instead.

This leads to rather obscure failure mode. I am not sure why the fallback is necessary.  If it is not necessary, I would propose removing it and let import errors like this to be propagated truthfully.

Note that this issue is about the more general behaviour of using some default editor as a fallback. The test failures are separate issues.

PR #921, opened 2020-06-25
Maint: Add raise_to_debug when failed to import a toolkit specific editor

kitchoi:
Motivated by #864 and #920, it is hard to debug when a different editor is used in place of the intended one.
This PR adds `raise_to_debug` to allow exceptions otherwise swallowed to be propagated in development environment.
We might want to add this for all types of editors.


kitchoi:
Hmmm loads of things fail if we let errors otherwise suppressed to propagate, and I could not get to the Appveyor build I was interested in because the build setting stops at the first failure. Looks like I will need to get my own Windows laptop out.
Closing.

Issue #866, opened 2020-06-03
Curate demos, examples and tutorials


corranwebster:
There is a lot of material in the examples directory, however it is not well organized and the point of may examples is not clearly explained.  Some effort should be spent to curate the examples, removing duplicates, modernizing code, and improving explanations.

ievacerny:
So far all demos in `examples/demo/Standard_Editors` have been fixed up and warnings about broken demos added where relevant.

A list of demos that have been fixed up in `examples/demo/Advanced`:
- `Auto_update_TabularEditor_demo.py`
- `Date_editor_demo.py`
- `Date_range_editor_demo.py`
- `List_editors_demo.py `
- `Multi_select_string_list.py`
- `NumPy_array_tabular_editor_demo.py`
- `Property_List_demo.py`
- `String_list_ui_editor.py`
- `Table_editor_with_checkbox_column.py`
- `Table_editor_with_context_menu_demo.py`
- `Table_editor_with_progress_column.py`
- `Tabular_editor_demo.py`
- `Tabular_editor_with_context_menu_demo.py`
- `Time_editor_demo.py`

I briefly looked at `examples/demo/Advanced/Auto_editable_readonly_table_cells.py` and I think this demo is not worth fixing up - the functionality doesn't seem to exist in Qt implementation and Wx implementation is broken with multiple different errors. It might be best to just delete it.

ievacerny:
A potential minor improvement to `List_editors_demo.py`:

> One additional thing - the tabs in `List_edtors_demo.py` with notebook mode `ListEditor` are not consistent between wx and qt - in wx they are document style tabs that can be moved around; qt has dialog style tabs that are not movable.
> 
> In issue #814 it is mentioned that document style tabs are now supported by qt `ListEditor` but I'm struggling to find that functionality. @corranwebster are they actually supported and is it possible to switch the demo to use them instead of the dialog style tabs?

This was originally a comment in #965 but the PR has already been merged.

kitchoi:
> This example made my computer hang to the point I had to restart: examples/demo/Advanced/Table_editor_with_live_search_and_cell_editor.py

This was fixed by #979

PR #924, opened 2020-06-25
Maintenance of Standard_Editors/TextEditor_demo.py

ievacerny:
Addresses #866 

This PR deletes the largely duplicate demo in `Standard_Editors/Popup_versions/TextEditor_demo.py`. It also does maintenance of the remaining demo.

A couple things I'm not sure about and would like opinions:

1. The remaining demo has been updated 9 years ago to remove the editor styles that aren't "useful". But I think it might be beneficial to show all styles so that the users themselves can select what is useful. And the information on which styles are recommended can be recorded in the description of the demo.

2. There is a `FIXME` suggesting that `text` style editor for passwords should show the typed characters. Is this a useful feature that deserves its own issue or should I just remove the `FIXME`?


kitchoi:
> There is a FIXME suggesting that text style editor for passwords should show the typed characters. Is this a useful feature that deserves its own issue or should I just remove the FIXME?

This FIXME seems very out-of-date. I don't see an issue with the current password style field: The characters should be hidden and they are now. Or is there an issue that I misunderstood?

examples/demo/Standard_Editors/TextEditor_demo.py
```diff
@@ -55,19 +53,21 @@ class TextEditorDemo(HasTraits):
...
         Item('password', style='simple', label='Simple'),
         # custom and text style are the same as simple, not shown.
+        # readonly style is not useful for passwords.
         label='Password'
     )
 
     # The view includes one group per data type. These will be displayed
     # on separate tabbed panels:
-    traits_view = View(
+    view = View(
```
kitchoi, line 62: Hmmmm I think defining a `default_traits_view` method is the recommended style nowadays. I might be mistaken though. @corranwebster Could you confirm please?

ievacerny: From offline discussion - static `view` definition is preferred for demos

kitchoi: Offline discussion: Not true. Both static and dynamic ways are good.
I am not sure whether it should be `traits_view` or `view` though.

ievacerny: Yes, sorry. Both are good, but for simple cases static `view` definition is preferred.

As for `traits_view`, I think it's best to rename it because of https://github.com/enthought/traits/issues/705

kitchoi: Apparently one can call it anything they like and the last one standing is used. Here is the relevant magic:
https://github.com/enthought/traits/blob/e73cae932cca147661a8f2aedb0489044cbe9e2e/traits/has_traits.py#L498-L500

kitchoi: I believe `traits_view` will need to be whitelisted for https://github.com/enthought/traits/issues/705 because it is explicitly advertised by traits:
https://github.com/enthought/traits/blob/e73cae932cca147661a8f2aedb0489044cbe9e2e/traits/has_traits.py#L105-L106
https://github.com/enthought/traits/blob/e73cae932cca147661a8f2aedb0489044cbe9e2e/traits/has_traits.py#L1704-L1709

So we have two views defined on this class, one is called `traits_view` and the other one called `view`. The one assigned to `traits_view` will be used. 
Just to add to that fun, if `default_traits_view` is also defined, then `default_traits_view` trumps the rest.

This is getting a bit orthogonal. Since the original example use `traits_view` and the `view` name is never advertised by the API (looks like), shall we use `traits_view` here?

ievacerny: Okay, I'll use `traits_view` in all demos and if a time comes for that to change, renaming should be easy via find-replace

PR #925, opened 2020-06-25
Increase precision offset for process_cascade_events and Windows

kitchoi:
While Linux/OSX typically have decent timing precision (in the order of microseconds), Windows has a worse [temporal resolution](https://docs.microsoft.com/en-us/windows-hardware/drivers/kernel/high-resolution-timers).

This PR increases the precision tolerance in `process_cascade_events` to account for the larger precision error on Windows. Again false positive is okay as long as the next redundant run returns fast enough to exit the loop, so it is fine for Linux/OSX to receive the same treatment.

### DIFF
--- a/traitsui/tests/_tools.py
+++ b/traitsui/tests/_tools.py
@@ -151,11 +151,13 @@ def process_cascade_events():
         # Qt won't raise if there are still events to be processed before
         # the time limit is reached. There are no other safe way to tell
         # if there are pending events (`hasPendingEvents` is deprecated).
-        # The minus-one is to account for precision differences between Python
+        # The offset is to account for precision differences between Python
         # and Qt, false positive triggers another redundant run that should
         # return immediately so that is fine.
+        # The precision is worse on Windows, typically around ~15 milliseconds.
+        # Give it a 1% error offset, which should be more than enough.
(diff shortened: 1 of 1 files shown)


traitsui/tests/_tools.py
```diff
@@ -151,11 +151,13 @@ def process_cascade_events():
...
         # if there are pending events (`hasPendingEvents` is deprecated).
-        # The minus-one is to account for precision differences between Python
+        # The offset is to account for precision differences between Python
         # and Qt, false positive triggers another redundant run that should
         # return immediately so that is fine.
+        # The precision is worse on Windows, typically around ~15 milliseconds.
+        # Give it a 1% error offset, which should be more than enough.
         while (start is None
-                or (time.time() - start) * 1000 >= _TOLERANCE_MILLISECS - 1):
+                or (time.time() - start) * 1000 >= _TOLERANCE_MILLISECS - 50):
```
itziakos, line 160: FYI, you can get better time resolution using `timeit.default_timer`. The returned timer offer much better resolution on windows.

kitchoi: That's a good idea. I will use the `default_timer`.

PR #926, opened 2020-06-26
Maintenance of Standard_Editors/TableEditor_demo.py

ievacerny:
Addresses #866 

This PR deletes the largely duplicate demo in `Standard_Editors/Popup_versions/TableEditor_demo.py`. It also does maintenance of the remaining demo.

The demo itself does mostly work as described in [documentation](https://docs.enthought.com/traitsui/traitsui_user_manual/factories_advanced_extra.html#tableeditor) if run with wx backend. However, there are lots of errors related to wx upgrade, sorting doesn't work at all and it's difficult to use. When demo is run using qt backend, a lot of functionality is missing because it was never implemented in qt. A warning has been added about the broken state of the demo.

More notable differences between remaining and deleted demo:
* Old demo had a gender (`Enum`) field. It was shown but none of the values were explicitly set. I decided not to carry it into the remaining demo as I don't think it adds much (and gender can be a sensitive topic).
* Search was `RuleTableFilter` instead of `EvalTableFilter`. Given that search doesn't work with qt editor, I decided to ignore this difference.


ievacerny:
The new demo tests are already worth it - they caught my mistake with the imports!

examples/demo/Standard_Editors/TableEditor_demo.py
```diff
@@ -38,56 +41,60 @@ class Employee(HasTraits):
...
+        ObjectColumn(name='last_name', width=0.20),
+        ExpressionColumn(
+            label='Full Name',
+            width=0.30,
+            expression="'%s %s' % (object.first_name, "
+            "object.last_name )"
+        ),
+        ObjectColumn(name='age', width=0.10, horizontal_alignment='center'),
+        ObjectColumn(name='phone', width=0.20)
+            ],
```
kitchoi, line 59: Overly indented?

examples/demo/Standard_Editors/TableEditor_demo.py
```diff
@@ -2,34 +2,37 @@
 #  License: BSD Style.
 
 """
+WARNING
+-------
```
kitchoi, line 6: This does look pretty prominent... Just wondering if you meant for it to look this big.
Screenshot 2020-06-26 at 17 29 51

Smaller options would be `**WARNING**` (without `----` in the next line), or something like that.

ievacerny: I meant it to be drawing attention and the smaller option seemed to blend in a bit too much.

A warning block would be ideal but they don't seem to work with the demo app. Maybe this "definition" pattern draws enough attention:

kitchoi: Yep. That looks like a good compromise.

PR #929, opened 2020-06-28
Empty traitsui-stubs package

rahulporuri:
Just an empty traitsui-stubs package with empty stub files for the relevant modules in the traitsui package. The empty stub pyi files only have a copyright header and nothing else.

The setup, readme, license, manifest have been pulled and modified as needed from the traits-stubs package https://github.com/enthought/traits/tree/master/traits-stubs .

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,5 +12,8 @@ build/
 dist/
 docs/build/
 
+# ignore mypy cache
+.mypy_cache/
+
 # Ignore the auto-generated version file
 traitsui/_version.py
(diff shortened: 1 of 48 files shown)


corranwebster:
I'm curious about the "why" of this?  Is there something that having this does that running `mypy` with `--ignore-missing-imports` doesn't achieve?

rahulporuri:
> I'm curious about the "why" of this?

laying the groundwork to slowly start adding type stubs for traitsui classes.

Issue #953, opened 2020-07-01
File dialog issues


ievacerny:
Issues found by going through `Standard_Editors/File_Dialog` demos on MacOS system.

- [ ] 1. Wx file dialog closure doesn't work properly. If either OK or Cancel are clicked, it seems that the dialog doesn't close properly as the main demo window remains unresponsive (beeps on click). Could potentially be related to issues with FileEditor (#889) or improper disconnect of event handlers (#752). Closing the dialog via system "x" goes back to a responsive main demo window, however, closing the demo window doesn't actually quit the demo. I'm guessing this is because of the dialog still hanging around somewhere. `Ctrl+C` right now is the only way to fully quit the demo. This happens with wx 4.0 and 4.1.

- [ ] 2. History is advertised as a big benefit of TraitsUI file dialog over system's dialog, however, it doesn't seem to work. Qt implementation always selects current working directory while wx implementation opens on the root system folder. Wx implementation history might be affected by issue (1).

- [ ] 3. Dialogs are affected by #944 

- [ ] 4. Using `FileInfo` extension doesn't show file size. This is because of errors in `commatize` helper function that could be replaced by locale specific number formatting.

- [ ] 5. Issue with using filters (I think they might be related to saving preferences as behaviour differs sometimes):
  - In `File_Open_with_TextInfo_Extension.py` demo launched with wx toolkit individual files aren't shown. This is a serious problem as the whole point of the demo is to show the contents of files.
  - In `File_Open_with_ImageInfo_Extension.py` demo launched with qt toolkit individual files aren't shown. Again, this beats the purpose of the demo as no image file can be selected to check its infromation.

- [ ] 6. Qt `File_Open_with_TextInfo_Extension.py` demo doesn't show contents of python files due to the following error:
  ```
  Exception occurred in traits notification handler for object: <traitsui.file_dialog.TextInfo object at 0x12c096308>, trait: file_name, old value: /Users/icernyte/Documents/repos/traitsui/appveyor.yml, new value: /Users/icernyte/Documents/repos/traitsui/setup.py
  Traceback (most recent call last):
    File "/Users/icernyte/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/traits/trait_notifiers.py", line 578, in _dispatch_change_event
      self.dispatch(handler, *args)
[...]

PR #954, opened 2020-07-01
Maintenance of Standard_Editors/File_Dialog demos

ievacerny:
Addresses #866

This PR mainly does stylistic and flake8 fixes to demos in `Standard_Editors/File_Dialog`. It also adds warning about all these demos being broken (see issue #953)

A few more significant changes:
- replace "Traits" references to "TraitsUI"
- replace `io.open` with Python built-in `open` and use it as a context manager
- remove the use of `commatize` function in demo code because it raises errors. Use integer string formatting instead

### DIFF
--- a/examples/demo/Standard_Editors/File_Dialog/File_Open_with_Custom_Extension.py
+++ b/examples/demo/Standard_Editors/File_Dialog/File_Open_with_Custom_Extension.py
@@ -15,9 +15,6 @@ from traitsui.api import View, VGroup, HGroup, Item
 
 from traitsui.file_dialog import open_file, MFileDialogModel
 
-from traitsui.helper import commatize
-
-from io import open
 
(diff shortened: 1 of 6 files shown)


examples/demo/Standard_Editors/File_Dialog/File_Open_with_Custom_Extension.py
```diff
@@ -46,30 +43,29 @@ class LineCountInfo(MFileDialogModel):
...
             if getsize(self.file_name) > 10000000:
                 return 'File too big...'
 
-            fh = open(self.file_name, 'rU', encoding='utf8')
-            data = fh.read()
-            fh.close()
-        except:
+            with open(self.file_name, 'r', encoding='utf8') as fh:
+                data = fh.read()
+        except OSError:
```
kitchoi, line 56: 👍

Issue #951, opened 2020-07-01
process cascade events occasional failure with PyQt and Appveyor


kitchoi:
The following failure occurs a few times on [#932](https://github.com/enthought/traitsui/pull/932) but it looks obviously unrelated:
```
======================================================================
FAIL: test_qt_process_events_process_all (traitsui.tests.test__tools.TestProcessEventsRepeated)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-pyside2\lib\site-packages\traitsui\tests\test__tools.py", line 113, in test_qt_process_events_process_all
    self.assertEqual(q_object.n_events, max_n_events)
AssertionError: 2 != 10
```

It is not reproducible on an actual Windows machine.

kitchoi:
It is weird that we are seeing more of this lately, we got another one from: https://github.com/enthought/traitsui/pull/955

kitchoi:
One with [PyQt5](https://ci.appveyor.com/project/EnthoughtOSS/traitsui/builds/33875471/job/2qahd9ewmfo8odrs), so it is not just PySide2.

kitchoi:
This is the first one I have seen a different error message where the actual number of events wasn't `2`, but this was with [PyQt4 and a Travis CI build on OSX](https://travis-ci.org/github/enthought/traitsui/jobs/705719573):
```
======================================================================
FAIL: test_qt_process_events_process_all (traitsui.tests.test__tools.TestProcessEventsRepeated)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/travis/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/traitsui/tests/test__tools.py", line 110, in test_qt_process_events_process_all
    self.assertEqual(q_object.n_events, 1)
AssertionError: 3 != 1
```
(Note that this is on #971, and "pyqt" means PyQt4)

kitchoi:
Turns out this test failure is very reproducible with PyQt4 and Mac OSX! (Discovered via #971)

kitchoi:
For reference, if I install Qt 4.8.7, and change `process_cascade_events` to this instead:
```
        from pyface.qt import QtCore
        event_loop = QtCore.QEventLoop(QtCore.QCoreApplication.instance())
        for _ in range(20):
            result = event_loop.processEvents(QtCore.QEventLoop.AllEvents)
            print(result)
```
[...]

PR #959, opened 2020-07-03
TST: Modify test for processing cascade events to use different event types

kitchoi:
[Qt documentation](https://doc.qt.io/qt-5/eventsandfilters.html#sending-events): 
> postEvent() posts the event on a queue for later dispatch. The next time Qt's main event loop runs, it dispatches all posted events, with some optimization. For example, if there are several resize events, they are compressed into one. The same applies to paint events: QWidget::update() calls postEvent(), which eliminates flickering and increases speed by avoiding multiple repaints.

This is another hypothesis: Because the QObject is going to receive events that have the same event type, derived from the same class with the same constructor, Qt has managed to "optimize" them into one?

For #951

### DIFF
--- a/traitsui/tests/test__tools.py
+++ b/traitsui/tests/test__tools.py
@@ -31,6 +31,9 @@ if is_current_backend_qt4():
 
     from pyface.qt import QtCore
 
+    class CustomEvent(QtCore.QEvent):
+        pass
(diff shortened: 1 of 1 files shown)


kitchoi:
Not this either. :(

Issue #814, opened 2020-05-11
Tabbed Group doesn't match WxPython appearance on Qt/OS X


corranwebster:
This is an old issue, but the `Tabbed` group on WxPython uses Pyface Dock tabs, which are "document-style" tabs (all rendered on the Python side):

![image](https://user-images.githubusercontent.com/600761/81566918-7758b800-9393-11ea-9467-c64191be5be1.png)

However on Qt on Mac it uses "dialog-style" tabs which have different connotations:

![image](https://user-images.githubusercontent.com/600761/81567206-e9310180-9393-11ea-9912-af03b15c0418.png)

When the Qt backend was written there was no alternative, but now we can specify document style tabs. This is already supported by the List Editor, but we should shift `Tabbed` group to use the same.

For those less usual situations where the older style makes sense (eg. at the top-level of a dialog) there may be a case for a flag to permit native dialog style tabs, but this should perhaps be done in a secondary way to allow support on wxpython as well (using wx.Notebook, for example).

PR #965, opened 2020-07-03
Maintenance of TableEditor demos in demos/Advanced

ievacerny:
Addresses #866

These are mostly flake8, formatting and other minor stylistic fixes for `TableEditor` demos in `demos/Advanced`. There are two more `TableEditor` demos that are not dealt with in this PR: `Table_editor_with_live_search_and_cell_editor.py` and `Auto_editable_readonly_table_cells.py` .

`List_editors_demo.py` is broken because of differences between `TableEditor` implementations.

More important changes:
- Added `sortable=False` in demos where this feature isn't important to avoid `TableEditor` implementation differences
- Added `show_toolbar=True` to `List_editors_demo.py` because a lot of the functionality added there wasn't accessible. Now it is accessible but the differences between implementation are more obvious.
- Simplified an expression in demo with checkboxes so that it's easier for beginner programmers to understand.

### DIFF
--- a/examples/demo/Advanced/List_editors_demo.py
+++ b/examples/demo/Advanced/List_editors_demo.py
@@ -2,6 +2,13 @@
 #  License: BSD Style.
 
 """
+**WARNING**
+
+  This demo might not work as expected and some documented features might be
+  missing.
+
+-------------------------------------------------------------------------------
+
 This shows the three different types of editor that can be applied to a list
 of objects:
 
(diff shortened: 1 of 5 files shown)


ievacerny:
One additional thing - the tabs in `List_edtors_demo.py` with notebook mode `ListEditor` are not consistent between wx and qt - in wx they are document style tabs that can be moved around; qt has dialog style tabs that are not movable.

In issue https://github.com/enthought/traitsui/issues/814 it is mentioned that document style tabs are now supported by qt `ListEditor` but I'm struggling to find that functionality. @corranwebster are they actually supported and is it possible to switch the demo to use them instead of the dialog style tabs?

PR #971, opened 2020-07-06
Bring Qt4 back to CI

### LINKED ISSUE #821: Revert #815
Unfortunately, it looks like we can't quickly deprecate Qt4.  Let's revert #815 and corresponding changes in Pyface.

kitchoi:
Closes #821 

This PR reverts #815 and #818 but not literally, because `pyside` does not support Python 3 so that does not need to be brought back.

With that `traitsui-test-3.6-pyqt` development environment refers to the PyQt4 environment again! (It is PyQt5 on master currently).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,16 +15,14 @@ env:
 
 matrix:
   include:
-    - env: RUNTIME=3.6 TOOLKITS="null pyside2"
-    - env: RUNTIME=3.6 TOOLKITS="pyqt"
+    - env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5 pyside2"
     - env: RUNTIME=3.6 TOOLKITS="wx"
     - os: osx
-      env: RUNTIME=3.6 TOOLKITS="null pyqt pyside2"
+    - env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5 pyside2"
     - os: osx
       env: RUNTIME=3.6 TOOLKITS="wx"
   allow_failures:
     - env: RUNTIME=3.6 TOOLKITS="wx"
-    - env: RUNTIME=3.6 TOOLKITS="pyqt"
     - os: osx
       env: RUNTIME=3.6 TOOLKITS="wx"
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ matrix:
     - env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5 pyside2"
     - env: RUNTIME=3.6 TOOLKITS="wx"
     - os: osx
-    - env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5 pyside2"
+      env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5 pyside2"
     - os: osx
       env: RUNTIME=3.6 TOOLKITS="wx"
   allow_failures:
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -15,16 +15,14 @@ env:
 
 matrix:
   include:
-    - env: RUNTIME=3.6 TOOLKITS="null pyside2"
-    - env: RUNTIME=3.6 TOOLKITS="pyqt"
+    - env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5 pyside2"
```
rahulporuri, line 18: don't we need to mark `pyqt5` as an allowed failure anymore?

appveyor.yml
```diff
@@ -12,7 +12,11 @@ environment:
     - RUNTIME: '3.6'
       TOOLKITS: "null"
     - RUNTIME: '3.6'
-      TOOLKITS: "pyside2 pyqt"
+      TOOLKITS: "pyqt"
+    - RUNTIME: '3.6'
+      TOOLKITS: "pyqt5"
+    - RUNTIME: '3.6'
+      TOOLKITS: "pyside2"
```
rahulporuri, line 19: any specific reason you chose to split them up into multiple jobs instead of running tests for all toolkits in the same job?

etstool.py
```diff
@@ -125,7 +125,8 @@
 extra_dependencies = {
     # XXX once pyside2 is available in EDM, we will want it here
     'pyside2': set(),
-    'pyqt': {'pyqt5'},
+    'pyqt': {'pyqt<4.12'},  # FIXME: build of 4.12-1 appears to be bad
```
rahulporuri, line 128: 4.12.1-1 is available. Do we want to try using that?

appveyor.yml
kitchoi: Yes... because I "messed up" the `appveyor-run.cmd` in https://github.com/enthought/traitsui/pull/940
The old `appveyor-run.cmd` took me a while to understand and I misunderstood it.  If I did not split it up, the current `appveyor-run.cmd` will only run test for the first toolkit in the space separated string.
I am not keen on reverting the `appveyor-run.cmd` just so that we can lump all these into one build. It is at best an optimization for maybe almost 2 mins of build time.  Splitting them up is more readable, with the benefit that now we can run tests on wx even if the test fails for an earlier toolkit.

.travis.yml
kitchoi: The pyqt5 build was allowed to fail as a result of #815 (https://github.com/enthought/traitsui/issues/821#issuecomment-649736045)

I don't remember the exact reason why they are linked. The tests are passing now with Qt5 so it would be good to move it out of the allow-to-fail group.

etstool.py
kitchoi: I don't know... this is the part where I went: "I am not sure if I could do things differently, let's do what we did before".

kitchoi: I will try using `4.12.1` in a separate PR and see if it changes test behaviour.

rahulporuri: yep yep. the changes doesn't have to be in this PR but I just wanted to see if 4.12.1 is any better/safer/less crashy.

appveyor.yml
rahulporuri: fine by me

.travis.yml
rahulporuri: 👍

etstool.py
kitchoi: I just tried 4.12.1, and a test in `TestViewApplication` will segfault. The clean up code looks suspicious.
But the rest of the tests seem to behave well.

PR #972, opened 2020-07-07
TST: Expand test to provide more information for diagnosing

kitchoi:
For #951, it was not clear whether it was the test itself being problematic, or whether the test is doing the right thing but the test objective fails.
This PR expands the failure message for the tests on `process_cascade_events`, in order to provide more information for diagnosing the problem, when it does fail.

### DIFF
--- a/traitsui/tests/test__tools.py
+++ b/traitsui/tests/test__tools.py
@@ -123,8 +123,8 @@ class TestProcessEventsRepeated(unittest.TestCase):
 
         n_left_behind_events = q_object.n_events - actual
         msg = (
-            "Expected {max_n_events} events processed on the objects after "
-            "running process_cascade_events and zero events left. "
+            "Expected {max_n_events} events processed on the objects and zero "
+            "events left on the queue after running process_cascade_events. "
             "Found {actual} processed with {n_left_behind_events} left "
             "behind.".format(
                 max_n_events=max_n_events,
--- a/traitsui/tests/test__tools.py
+++ b/traitsui/tests/test__tools.py
@@ -115,7 +115,7 @@ class TestProcessEventsRepeated(unittest.TestCase):
         # then
         actual = q_object.n_events
 
-        # If process_cascade_events fails for genuine reasons, then there
+        # If process_cascade_events did not do what it promises, then there
         # are still pending tasks left. Run process events at least the same
         # number of times as max_n_events to verify
         for _ in range(max_n_events):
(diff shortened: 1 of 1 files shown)


kitchoi:
There are times when I actually want the test to fail so I can see what's going on... :/
I want to make sure this change did not magically fix the issue without understanding why, so I will go and try to make it fail by... rebuilding!

traitsui/tests/test__tools.py
```diff
@@ -113,7 +113,27 @@ def cleanup(q_object):
...
+            "running process_cascade_events and zero events left. "
+            "Found {actual} processed with {n_left_behind_events} left "
+            "behind.".format(
+                max_n_events=max_n_events,
+                actual=actual,
+                n_left_behind_events=n_left_behind_events,
+            )
+        )
+        self.assertEqual(n_left_behind_events, 0, msg)
+        self.assertEqual(actual, max_n_events)
```
kitchoi, line 136: Missing a `msg` here.

traitsui/tests/test__tools.py
```diff
@@ -113,7 +113,31 @@ def cleanup(q_object):
...
+                actual=actual,
+                n_left_behind_events=n_left_behind_events,
+            )
+        )
+        self.assertEqual(n_left_behind_events, 0, msg)
+
+        # If the previous assertion passes but this one fails, that means some
+        # events have gone missing, and that would likely be a problem for the
+        # test setup, not for the process_cascade_events.
+        self.assertEqual(actual, max_n_events, msg)
```
kitchoi, line 140: Strictly speaking, the original assertion with `actual == max_n_events` has made an assumption about the test setup being correct. If one needs to be pedantic, an assertion error here means the prerequisite of the test has not been met and it should result in a test error instead of a test failure. I think the comment is good enough to explain what this assertion error implies.

kitchoi:
5+ builds (lost count) and the said test has not failed once... it is weird and I hope this hasn't fixed the problem because if it does I would never understand why it happened in the first place. Ready for review nonetheless. If the test fails again later, we will get more information upon failure.

rahulporuri:
I'm out of my depth on this PR. I have no clue what a `DummyObject` is and what happens to `n_events` when we do `processEvents` on the application. I'll need a day to get a better hang of this code and approve.

kitchoi:
No worries. This PR was really just about adding more information to the test failure message because the error had not been reproducible locally. This was before I realized that the error is actually pretty reproducible with Qt 4, but we had got rid of Qt4 for sometime.

PR #975, opened 2020-07-07
Expand caution notes and skip test for process_cascade_events on certain conditions

kitchoi:
After some digging, the issue in #951 was due to:

- `QtCore.QEventLoop.processEvents` may misreport whether it has processed some pending events on certain platforms, build conditions and Qt version.
- The ability to process cascade events relies on looping over the result of `QEventLoop.processEvents`: If it returns true, run it again. However, when it erroneously returns false, the loop can break too early, leaving cascaded events unprocessed.

This has been observed on Qt5 + Appveyor, and Qt4 + OSX.

This PR:
- Expand docstring of `process_cascade_events` to note the situation. At worst, it is as good as `processEvents` which process only the current snapshots of events. At best, it can achieve what we often wanted in tests.
- Skip the tests for the problematic Qt version and platform combination. The tests are still run on Linux.
- Rewrote the function to do away with the timer. Using the returned value from `QEventLoop.processEvents` in a loop with a time limit is essentially what `QCoreApplication.processEvents(flags, max_time)` is doing. Here we remove the timer part.

Required for #971 to bring Qt4 back to CI

### DIFF
--- a/traitsui/tests/test__tools.py
+++ b/traitsui/tests/test__tools.py
@@ -11,6 +11,7 @@
 
 """ Tests for traitsui.tests._tools """
 
+import sys
 import unittest
 
 from pyface.api import GUI
@@ -18,6 +19,7 @@ from pyface.api import GUI
 from traitsui.tests._tools import (
     is_current_backend_qt4,
     is_current_backend_wx,
+    is_mac_os,
     skip_if_not_qt4,
     skip_if_not_wx,
     process_cascade_events,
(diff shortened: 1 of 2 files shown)


kitchoi:
> Skip the tests for the problematic Qt version and platform combination. 

I unskipped Qt5 + Appveyor for now. Because I'd still like to have a confirmation that the error seen on Appveyor has the same symptom compared to the error I can reproduce on OSX. For that, #972 is needed to provide more information when the test fails on Appveyor. Once we have that information on record, we can proceed to skip the test on Qt5 + Appveyor.

traitsui/tests/test__tools.py
```diff
@@ -87,6 +89,14 @@ class TestProcessEventsRepeated(unittest.TestCase):
     def test_qt_process_events_process_all(self):
         from pyface.qt import QtCore
 
+        if QtCore.__version_info__ < (5, 0, 0) and is_mac_os:
```
rahulporuri, line 92: any reason why we can't use the `is_qt4` util? (i dont remember now if that's in pyface or traitsui or both)

rahulporuri: and if we use the `is_qt4` util, we can simply move this `skipTest` from inside the method to a decorator.

kitchoi: `is_qt4` is from pyface. I could import `is_qt4` from pyface and there it does just mean Qt4 only. But then in `traitsui.tests._tools` , the `is_current_toolkit_qt4` actually means both Qt4 and Qt5 and is used in a lot of places in traitsui. Currently, all `is_whatever_qt4` in traitsui means both Qt4 and Qt5. We need to sort this out. But before that I want to avoid mixing more meaning for these `is_whatever_qt4` within traitsui.

kitchoi: #978

PR #985, opened 2020-07-10
Fix invalid value in Item not being used

### LINKED ISSUE #983: 'invalid' in Item no longer used for dynamically setting error state
In the following example:
```
from traits.api import Bool, HasTraits, Property, Str
from traitsui.api import TextEditor

class Person(HasTraits):
    name = Str()

    invalid = Property(Bool, depends_on='name')

    def _get_invalid(self):
        # Name is valid iff it doesn't consist entirely of whitespace.
        stripped_name = self.name.strip()
        return stripped_name == ''

    traits_view = View(
        Item('name', invalid='invalid'),
    )

if __name__ == '__main__':
    Person().configure_traits()
```

The text box is expected to be red if all the characters in the field are whitespaces.
This works in 7.0.0 but no longer works on master.

(The example above is adapted from `traitsui/integrationtests/ui/text_editor_invalid.py`. If instead of setting `invalid` on `Item`, set it on the editor factory, then the invalid state works.)

Believed to be a regression from #859, we should fix this before the 7.0.1 release.

kitchoi:
Closes #983 
Revert #859

The issue in #790 is fixed by upgrading to traits 6.1 (see [PR in traits](https://github.com/enthought/traits/pull/1032)).  Attempt to fix the issue independently causes breakages as shown in #983. So this PR reverts those changes.  Users who would like #790 being fixed will have to upgrade traits to >= 6.1 instead.

It might be easier to review this commit-by-commit. This PR:
(1) Add a regression test for #790, because this PR will revert the PR that aims at fixing the issue independently of traits. This test passes on master even with traits 6.0
(2) Add failing tests for #983
(3) Revert #859. This allows the tests added for #983 to pass. The test added for #790 still passes if traits version is 6.1, but fails if traits version is less than 6.1
(4) Skip the test added for #790 if the traits in the environment has version < 6.1. (On CI we have traits 6.1, but traits 6.0 is still supported for anyone installing traitsui 7.x from PyPI and therefore the test should account for that)
(5) Add `packaging` in the test dependencies for parsing version strings.

This PR will need to be backported for 7.0.1 release.

### DIFF
--- a/traitsui/tests/test_editor.py
+++ b/traitsui/tests/test_editor.py
@@ -402,8 +402,8 @@ class TestEditor(GuiTestAssistant, unittest.TestCase):
 
         editor.dispose()
 
-    # Test synchronizing built-in trait values between from factory
-    # to editor.
+    # Test synchronizing built-in trait values between factory
+    # and editor.
 
     def test_factory_sync_invalid_state(self):
         # Test when object's trait that sets the invalid state changes,
--- a/traitsui/tests/test_editor.py
+++ b/traitsui/tests/test_editor.py
@@ -144,6 +144,9 @@ class UserObject(HasTraits):
     #: An event user value
     user_event = Event()
 
+    #: A state that is to be synchronized with the editor.
+    invalid_state = Bool()
+
 
 def create_editor(
         context=None,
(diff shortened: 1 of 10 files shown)


traitsui/tests/test_editor.py
```diff
@@ -399,6 +402,31 @@ def test_parse_extended_name(self):
 
         editor.dispose()
 
+    # Test synchronizing built-in trait values between from factory
```
rahulporuri, line 405: spelling - "from factory to editor" or "between factory and editor"

PR #987, opened 2020-07-10
Backport bug fix to 7.0.1

kitchoi:
This PR:

- backports bug fix from #985
- updates the changelog accordingly for 7.0.1

Note that this is targeting the `maint/7.0` branch.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -21,8 +21,6 @@ Fixes
   (#848)
 * Fix demo application description (#850)
 * Fix auto-add functionality in ListStrModel (#860)
-* Move format_func, format_str and invalid traits from factory to editor
-  (#859, #900)
 * Fix segmentation fault for TabularModel (#871, #873)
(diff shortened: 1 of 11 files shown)


CHANGES.txt
```diff
@@ -21,8 +21,6 @@ Fixes
   (#848)
 * Fix demo application description (#850)
 * Fix auto-add functionality in ListStrModel (#860)
-* Move format_func, format_str and invalid traits from factory to editor
-  (#859, #900)
```
rahulporuri, line 25: i think we only reverted #859 , not #900 as well. right?

kitchoi: That's right.
The issue fixed by #900 was directly caused by #859 though. With #859 reverted, #900 isn't fixing anything visible to the users. We did not need to revert #900 either, because the change there can stay in its own right (it is correcting a function signature trying to ducktype the editor class's `__init__`).

PR #991, opened 2020-07-11
FIX: Enforce underscore for click command

ku-ya:
Applying a quick fix for [Command names with underscores must be invoked with dashes instead when upgrading from 6.7 to 7.0](https://github.com/pallets/click/issues/1123)

This will make the documentation command consistent: ```python etstools.py test_all```

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -148,7 +148,11 @@ environment_vars = {
 }
 
 
-@click.group()
+def normalize(name):
+    return name.replace("_", "-")
+
+
+@click.group(context_settings={"token_normalize_func": normalize})
 def cli():
     pass
 


rahulporuri:
> Do we need to change things elsewhere, such as in docstrings or README files to account for underscore-vs-hyphen

there are references to `test_clean` and `test_all` in the module docstring.

rahulporuri:
Ahh. I misunderstood this PR. This PR doesn't force users to only use `-` commands. This PR ensures that users can continue using both `_` and `-` commands i.e. on master `test_clean` won't work but `test-clean` does. On this PR, `test_clean` and `test-clean` both work as expected.

Issue #986, opened 2020-07-10
Occasional error from QAbstractItemModel events upon test tear down


kitchoi:
The following error occurs once from this travis build (https://travis-ci.org/github/enthought/traitsui/jobs/706873557) :
```
======================================================================
FAIL: test_run (test_all_examples.TestExample) (file_path='/home/travis/build/enthought/traitsui/examples/demo/Advanced/ListStrEditor_demo.py')
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/enthought/traitsui/integrationtests/test_all_examples.py", line 293, in test_run
    run_file(file_path)
AttributeError: 'NoneType' object has no attribute 'auto_add'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/travis/build/enthought/traitsui/integrationtests/test_all_examples.py", line 300, in test_run
    file_path, exc, message
AssertionError: Executing /home/travis/build/enthought/traitsui/examples/demo/Advanced/ListStrEditor_demo.py failed with exception 'NoneType' object has no attribute 'auto_add'
 Traceback (most recent call last):
  File "/home/travis/build/enthought/traitsui/integrationtests/test_all_examples.py", line 293, in test_run
    run_file(file_path)
  File "/home/travis/build/enthought/traitsui/integrationtests/test_all_examples.py", line 277, in run_file
    exec(content, globals)
  File "<string>", line 50, in <module>
  File "/home/travis/build/enthought/traitsui/integrationtests/test_all_examples.py", line 238, in replaced_configure_traits
    process_cascade_events()
  File "/home/travis/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/contextlib.py", line 88, in __exit__
    next(self.gen)
  File "/home/travis/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/traitsui/tests/_tools.py", line 67, in store_exceptions_on_all_threads
    raise exceptions[0]
  File "/home/travis/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/traitsui/qt4/list_str_model.py", line 53, in rowCount
    if editor.factory.auto_add:
AttributeError: 'NoneType' object has no attribute 'auto_add'

```

This is similar to the ones seen in https://github.com/enthought/traitsui/issues/854 for `ImageEnumEditor` and the one worked around in https://github.com/enthought/traitsui/pull/897 for `TabularEditor`.
[...]

PR #994, opened 2020-07-13
Unset ListStrModel from the ListStrEditor view upon dispose

kitchoi:
Related to #431 and resolves part of #986.
The error seen in #986 is revealed by unskipping the test in https://github.com/enthought/traitsui/pull/974.
As it turns out, events are still received by the `ListStrModel` used in the view of `ListStrEditor` after the UI is disposed.
The issue occurs most frequently in Qt4.

This PR attempts to resolve this by doing the same thing in #974: Unset the item model from the view in the editor dispose.
Since the error is sporadic, I might need to rebuild this a few time to verify the issue is fixed. 

This does not completely _fix_ #986 because there are other instances of `QAbstractItemModel` which we are not unsetting.

### DIFF
--- a/traitsui/qt4/list_str_editor.py
+++ b/traitsui/qt4/list_str_editor.py
@@ -123,6 +123,7 @@ class _ListStrEditor(Editor):
 
         if factory.title or factory.title_name:
             header_view = QtGui.QHeaderView(QtCore.Qt.Horizontal, self.control)
+            self._header_view = header_view
             header_view.setModel(self.model)
             header_view.setMaximumHeight(header_view.sizeHint().height())
             if is_qt5:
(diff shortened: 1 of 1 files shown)


traitsui/qt4/list_str_editor.py
```diff
@@ -512,6 +520,11 @@ def __init__(self, editor):
         # Configure context menu behavior
         self.setContextMenuPolicy(QtCore.Qt.CustomContextMenu)
 
+    def _dispose(self):
```
rahulporuri, line 523: i don't have enough context to know why `dispose` is private.

kitchoi: The class is also private, so it really doesn't matter what we call it here.

kitchoi: Except the instance is a public attribute on `ListStrEditor`. There are no clear guideline as to what is public to users of traitsui and what is protected. People import and subclass so freely so I'd prefer making things obviously private if we can. No one should be calling this except for the editor. Keeping it private for now.

PR #998, opened 2020-07-14
Add main function and EAM metadata for the demo application

### LINKED ISSUE #908: Make demo application installable from EDM app
This issue depends on #904 as it requires adding an entry point to the new package setup.py
See https://docs.enthought.com/eam/application_bundles.html#application-bundles

kitchoi:
Closes #908

This PR:
- Add a new main function for launching the demo application. Running this function will launch the GUI, but the GUI will show nothing interesting.
- Add EAM metadata and expose this via package entry point.
- Modify the demo application to allow for a placeholder node to be shown when there is no data. Without this change the main function would have to supply a temporary directory to the demo application. I think it makes sense for the application to fall back to some no-data-available view if it finds nothing to be shown.

Note that there exists another function for launching the demo application:
https://github.com/enthought/traitsui/blob/0a4a6f418d01f4d6e24a84eb5995202217c8e66c/ets-demo/etsdemo/app.py#L1157-L1163
I am deliberately creating a new one for because we will eventually move to using entry points. Projects wish to maintain a similar [`tutor.py`](https://github.com/enthought/traitsui/blob/01d9e5d8c898b5ac60519d3e39c9b92b0c03e21d/examples/tutorials/tutor.py) will need to call a different function later.
In fact, I might remove this function in a separate PR as the function is currently untested.


ets-demo/etsdemo/main.py
```diff
@@ -0,0 +1,27 @@
...
+
+""" The main function for launching the demo application.
+"""
+
+from etsdemo.app import Demo, DemoEmpty
+
+
+def _create_demo():
+    """ Create the demo object with everything setup ready to be launched. """
+    return Demo(model=DemoEmpty())
```
kitchoi, line 19: This function looks silly now but this will grow and we can test this function without requiring GUI.

ets-demo/etsdemo/tests/test_info.py
```diff
@@ -0,0 +1,44 @@
...
+"""
+
+import os
+import subprocess
+import unittest
+
+try:
+    import eam   # noqa: #F401
+except ImportError:
+    EAM_EXISTS = False
```
rahulporuri, line 21: i don't think `eam` can be an optional dependency so i'm not sure why we want to have a try/import here.

ets-demo/setup.py
```diff
@@ -281,8 +282,21 @@ def get_long_description():
         long_description=get_long_description(),
         long_description_content_type="text/x-rst",
         packages=setuptools.find_packages(include=["etsdemo", "etsdemo.*"]),
+        package_data={
+            "etsdemo": [
+                "images/*.png",
```
rahulporuri, line 287: i'm not sure if there is a manifest file in `ets-demo` but that we will probably want to update that as well.

ets-demo/etsdemo/tests/test_info.py
kitchoi: Why can't it be an optional dependency?
It is currently only a test dependency. Without eam, the application can still run.

ets-demo/setup.py
kitchoi: By adding the image in the `package_data`, this will include the file in the source distribution as well, so we might not need a manifest file yet?

ets-demo/etsdemo/tests/test_info.py
rahulporuri: >  It is currently only a test dependency. Without eam, the application can still run.

Ohh. Ha. Fair enough. Sorry.

Issue #862, opened 2020-06-02
Merge demo and tutorial examples


kitchoi:
From this https://github.com/enthought/traitsui/pull/808#discussion_r424535947:

> There are two entry points because before this year there were two completely different sources: this demo.py and a separate tutor.py that included a completely parallel and differently architected demo application.

This is now a merging of the two codebases so that it can handle both sets of examples. At some point the traits demo and tutor examples should be merged (with likely a lot of clean-up!)

This is likely an issue that requires a stream of PRs, not just one.
Opened for tracking purposes.

ievacerny:
A drive by comment: Traits only has `tutorials` so it might be good to only have `tutorials` in TraitsUI as well (merge and delete `demo`).

kitchoi:
On the two examples `view_standalone.py` and `view_multi_object.py` in tutorials, see [this comment](https://github.com/enthought/traitsui/pull/853#discussion_r434702889)

> A curation needs to be done with the examples, which may involve removing some, but more likely just re-writing and modernizing.
> 
> The multi-object view is still valid, it just needs re-working an generalizing to use pyface GUI: the point is to show how to have multiple objects in the context of a view and to access their values).
> 
> Similarly the stand-alone example is showing that the view doesn't need to be part of the class.

PR #1008, opened 2020-07-15
Move demo examples into the package as package data

### LINKED ISSUE #970: Include example files as package data for TraitsUI
Related to #966 

This issue is for making TraitsUI's examples (files in the current `examples` folder) of the package data so that it can be loaded by the demo application upon installation.

Care should be taken to ensure that we **don't** end up including a Python `examples` package that can be imported. The files are data. They are not packages.

kitchoi:
Closes #970

This PR
- Moves the `examples/demo` folder to `traitsui/examples/demo` and adds the content of this folder as package data.
- Update tests for launching and closing the GUI defined in the examples. 

With this PR, the demo examples are installed along with traitsui, and can be contributed to the demo application later.

The tests for these examples are useful again because they help me to include nested folders and files which I might otherwise have missed. The tests are not fool-proof, because they merely launch and close the GUI. A lot of the internal logics are not tested. I tried to manually check that the most deeply nested folder is included.

A symbolic link is created in the `examples` folder for discovery purposes. One downside of this move is that GitHub ~should not~ cannot resolve symbolic link to folders, so anyone browsing the files on GitHub will need to manually navigate to the correct folder. The opposite approach (having a symbolic link in the package) does not work with setuptools.

Note:
I selected `examples/demo` and excluded `examples/tutorials` because the content in `demo` is clearly written with the demo application in mind. Examples in `demo` have descriptive docstrings to be displayed in the application, and there is a config file and an image designed for this use case. Whereas for the `tutorials` folder, a lot of them don't have the `if __name__ == "__main__"` guard to prevent a GUI to be opened as soon as one clicks on an example from the demo application. It is quite annoying when a user browses the content lightly.
The examples in `examples/tutorials` should be curated and be merged with `examples/demo`. See #866 and #862.


setup.py
```diff
@@ -172,9 +172,18 @@ def additional_commands():
...
         maintainer_email='enthought-dev@enthought.com',
-        package_data=dict(traitsui=['extras/images/*', 'image/library/*.zip',
-                                    'images/*', 'wx/images/*',
-                                    'qt4/images/*']),
+        package_data=dict(
+            traitsui=[
+                'examples/demo/*',
+                'examples/demo/*/*',
+                'examples/demo/*/*/*',
+                'examples/demo/*/*',
```
rahulporuri, line 180: `examples/demo/*/*` included twice?

rahulporuri: i'm not sure what changes need to happen in `MANIFEST.in`

setup.py
```diff
@@ -172,9 +172,18 @@ def additional_commands():
         license='BSD',
         maintainer='ETS Developers',
         maintainer_email='enthought-dev@enthought.com',
-        package_data=dict(traitsui=['extras/images/*', 'image/library/*.zip',
-                                    'images/*', 'wx/images/*',
-                                    'qt4/images/*']),
+        package_data=dict(
+            traitsui=[
+                'examples/demo/*',
```
rahulporuri, line 177: do you think there's any advantage to explicitly listing the different sub dirs in `examples/demo/` which we are including or is that unnecessary?

rahulporuri:
I just tried building an sdist from this branch locally and it works as we expect - the new folders are created and the examples/demo content is copied over.

setup.py
kitchoi: > examples/demo/*/* included twice?

Oops. Thanks!

> i'm not sure what changes need to happen in MANIFEST.in

Nothing I think. `MANIFEST.in` already has the top level `examples` folder. We are moving something that is already in the source distribution to the package data (which will also be included in the source distribution).

setup.py
kitchoi: Seems unnecessary for now since we do want to include everything in there. If there is a reason to exclude some subdirectories and include some other directories, we can revisit this.

PR #1009, opened 2020-07-16
PyFace to Pyface change (non-code)

aaronayres35:
As mentioned in the comments on issue #584 this is the first PR for fixing the comments, docstrings, documentation, etc.  
Another PR will handle the deprecation associated with PyFaceAction.


kitchoi:
The build failure is known to be sporadic, and https://github.com/enthought/traitsui/pull/1007 is going to skip the test that fails, given we have the error message on record (see details there).

kitchoi:
#1007 is merged, I will trigger another Appveyor build, which should be on a merge commit and the skip should be present.

PR #1018, opened 2020-07-17
Fix symlink path for examples/demo

kitchoi:
The symbolic link to the new location of the demo files should be relative to where the symbolic link file is.
This PR fixes this. Now GitHub can recognize the symbolic link being a directory (though it still won't bring you directly to the actual folder).

### DIFF
--- a/examples/demo
+++ b/examples/demo
@@ -1,1 +1,1 @@
-traitsui/examples/demo
+../traitsui/examples/demo


rahulporuri:
So i have confirmed that github shows the `examples/demo` file differently on master vs this branch. But, this doesn't work for me locally on windows.

Can I ask what command you used to create the symlink? (I don't know if there is a platform-agnostic way to create symlinks or if there is even a way to do that!)

kitchoi:
Does setting your local git config help? Something like this?: https://git-scm.com/docs/git-config#Documentation/git-config.txt-coresymlinks
This is not unlike the handling of the platform dependent line-ending character I guess...

rahulporuri:
@aaronayres35 i think you're on a mac. can you test this and see if this works as expected? (we can worry about cross platform-ness later. for now, we just need another eyeball on this.)

kitchoi:
For reference, the `examples` folder on the top-level is only included in the source distribution, and the symbolic link is only for making it easier for people browsing the repository on GitHub to discover example files . So it is not a huge deal if `cd` it does not work: The path is still written on the symbolic link.

PR #1023, opened 2020-07-17
Feat: Extend demo application to load entry points and accept multiple data sources

### LINKED ISSUE #966: Support contributing demo example data via entry point
After #908, the demo application will need to be able to load example files contributed by other projects.

This will involve creating an API for other projects to adhere to.
As a first cut, an entry point should return a path-like object for loading a directory of files and a display name for the data.

For future consideration, not part of this issue:
The API should be extensible so that options can be added.  Some ideas: we might want to extend the API so that projects can omit certain files under certain conditions at runtime, e.g. traitsui can choose to hide examples that require Wx in a Qt environment. There might be conditions that are settable by the users, e.g. traitsui can advertise a set of more mature examples and a set of less stable examples separately (see #949 as well), and the user might be allowed to select them?

### LINKED ISSUE #995: Extend demo application to support multiple sources of example folders
Currently the demo application only supports showing a single root directory containing example files. Given the demo application is to be deployed as a standalone application, it will need to support loading example files contributed from more than one packages (e.g. traits, chaco, envisage, see #966 for introducing entry points for packages to contribute example files). This issue is for making the demo application capable of showing more than one source of example files.

kitchoi:
Closes #966 and closes #995

This PR:
- Change the `DemoEmpty` node placeholder to something more meaningful: A simple node that can hold a list of children nodes. This allows us to support showing no data to showing many data sources.
- Add logic to load metadata from entry points, and then convert them into tree nodes displayed in the application. Note that we need to capture errors and log, because the demo app cannot control what is being contributed.
- Add such an entry point for TraitsUI, with that, the demo application will show TraitsUI demo files, because the application depends on TraitsUI too. 
- Remove the `traitsui/examples/demo/demo.py` script
- Add a new script in `examples/demo.py` (which is not in the package data but in source distribution) to achieve what the old `demo.py` is doing, but use the new code.
- Update the README documentation for `ets-demo` accordingly.

To try on this branch:
```
$ pip install -U <root-level-of-this-repo>/
$ pip install -U <root-level-of-this-repo>/ets-demo/
$ etsdemo
```


ets-demo/etsdemo/loader.py
```diff
@@ -0,0 +1,121 @@
...
+
+    try:
+        return DemoPath(
+            nice_name=response["name"],
+            name=response["root"],
+        )
+    except KeyError:
+        logger.exception("Failed to load response: %r", response)
+        return _get_placeholder_node(
+            response, "Unable to load data. See log for more info."
```
kitchoi, line 99: The log is not really accessible at the moment, so this is a lie. See #1026
I will remove the second half of the sentence for now.

examples/demo.py
```diff
@@ -0,0 +1,16 @@
...
+if __name__ == "__main__":
+    main(
+        [
+            {
+                "version": 1,
+                "name": "TraitsUI Demo",
+                "root": os.path.realpath(os.path.join(HERE, "demo")),
+            },
+        ]
+    )
```
kitchoi, line 16: This file replaces the previous `demo.py`.  But I am not sure we actually need to resurrect this file:
(1) This script requires `etsdemo` to have been installed.
(2) The content of the demo requires `traitsui` to have been installed.

With that, running `etsdemo` from the command line will pull up the same content.

The only difference this script will make is that it will only show TraitsUI's example. Whereas launching `etsdemo` might show `traits`, `pyface`, `chaco`, ... examples if these other projects have also added the corresponding entry point for `etsdemo`.

Since this script requires maintenance (making sure the path is right etc), perhaps we should remove this script and add an instruction to "run `etsdemo`" in the README.  If we really need to have a script to exclude examples from other packages, it is easy to add this script back.

rahulporuri:
@kitchoi im assuming this PR is ready for review

kitchoi:
> im assuming this PR is ready for review

Yes it is. Thank you!

rahulporuri:
something very weird is happening right now. on windows, when i run the app using `etsdemo`, selecting any of the actual demo script nodes crashes the app. on the other hand, if i run the app using `python -c "from etsdemo import main; main.main()"`, the app works as expected and i don't see any crashes.

kitchoi:
!  Let me see if I could reproduce.

rahulporuri:
this might be an issue with console script/gui script difference on windows - i added a duplicate console script locally and reinstalled etsdemo - and i was able to use `etsdemo` from the command prompt without any issues.

rahulporuri:
a few comments about the app in general : 

- nit : application name/title is "edit properties" at the moment

app-title-win

- nit : application icon is this weird people (?) icon at the moment, even though we are shipping the Enthought E icon, which is advertised as the icon to EDM GUI

app-icon-win

- nit : it'd be nicer if there were a default selection when the app is opened. If i understand correctly, there will always be a nice in the app (empty node or real data node) so we should be able to set the selection without any issues

- nit : the default layout of the app is slightly awkward. Instead of splitting the vertical screen space in half equally, it might be better to split 1/3 and 2/3 or 1/4 and 3/4. (this with the default selection will be a much nicer landing page)

- nit : the default tab selected under a demo is "Output". That feels secondary to the actual thing which is "Demo".

default-selection-in-subpane

kitchoi:
Agreed that those are nice enhancements to have.

> nit : application name/title is "edit properties" at the moment

That's something small enough that I can sneak into this PR.

> nit : application icon is this weird people (?) icon at the moment, even though we are shipping the Enthought E icon, which is advertised as the icon to EDM GUI

That is orthogonal enough that I would not sneak into this PR.

> nit : it'd be nicer if there were a default selection when the app is opened. If i understand correctly, there will always be a nice in the app (empty node or real data node) so we should be able to set the selection without any issues

Yes this can be done, but I would prefer this to be a separate issue.
With `hide_root=True`, there can be no nodes in the tree, so we will need to handle that and test accordingly.

> nit : the default layout of the app is slightly awkward. Instead of splitting the vertical screen space in half equally, it might be better to split 1/3 and 2/3 or 1/4 and 3/4. (this with the default selection will be a much nicer landing page)

I did give it a try, but to no avail. The TreeEditor isn't very flexible for setting widths, so I gave up. I am half suspecting the original demo script uses a large enough TraitsUI demo image to workaround this. Maybe a separate issue too.

rahulporuri:
@kitchoi feel free to move my comments to separete issues that we can handle later (maybe even add to backlog in the sprint)

kitchoi:
> something very weird is happening right now. on windows, when i run the app using etsdemo, selecting any of the actual demo script nodes crashes the app. 

I can reproduce it.  It can reproduce it with `pythonw -c "from etsdemo import main; main.main()"` as well.
Something about this application that makes it not compatible with pythonw.
This will be a separate blocking issue.

kitchoi:
> I can reproduce it. It can reproduce it with pythonw -c "from etsdemo import main; main.main()" as well.
Something about this application that makes it not compatible with pythonw.
This will be a separate blocking issue.

Opened #1032. This is an existing behaviour with pythonw.

kitchoi:
Merging master to this branch so the branch alone can be (manually) tested with Windows more easily.

ets-demo/etsdemo/loader.py
```diff
@@ -0,0 +1,121 @@
...
+            response,
+            "Unable to load data: Path {!r} not found.".format(root)
+        )
+
+    try:
+        return DemoPath(
+            nice_name=response["name"],
+            name=response["root"],
+        )
+    except KeyError:
```
rahulporuri, line 96: Is this because `name` might not exist? Wouldn't it be better to move both of these checks (`root` and `name`) to the top of the function and raise the exception there?

ets-demo/etsdemo/loader.py
```diff
@@ -0,0 +1,121 @@
...
+        Response returned from a distribution's entry point.
+    msg : str
+        User-facing message to show.
+
+    Returns
+    -------
+    node : DemoTreeNodeObject
+    """
+    return DemoVirtualDirectory(
+        nice_name=response.get("name", "(Empty)"),
```
rahulporuri, line 118: nit : refactor : can we just pass `nice_name` to the `_get_placeholder_node` function instead of passing `response`?

ets-demo/etsdemo/main.py
```diff
@@ -11,17 +11,60 @@
...
+_TITLE = "Enthought Tool Suite"
 
 
-def main():
-    """ Main function exposed to the entry point for launching the demo
-    application.
+def _create_demo(infos=None, title=_TITLE):
+    """ Create the demo object with everything setup ready to be launched.
+
+    Paramters
```
rahulporuri, line 25: nit : spelling "Parameters"

examples/demo.py
rahulporuri: let's go with the option to remove this script in it's entirety and document the use of `etsdemo` instead

setup.py
```diff
@@ -192,6 +192,9 @@ def additional_commands():
                 'qt = traitsui.qt4:toolkit',
                 'null = traitsui.null:toolkit',
             ],
+            'etsdemo_data': [
+                'demo = traitsui.extras._demo_info:info',
```
rahulporuri, line 196: IIUC, we aren't using the LHS of the entry point (`demo = ...`), only the RHS, which is what we're calling to load the resources. I'm not sure what other projects do with the LHS. Do they just ignore it?

ets-demo/etsdemo/loader.py
```diff
@@ -0,0 +1,121 @@
...
+    responses = []
+    for entry_point in pkg_resources.iter_entry_points(ENTRY_POINT_GROUP):
+        try:
+            loader = entry_point.load()
+        except ImportError:
+            logger.exception("Failed to load entry point %r", entry_point)
+            continue
+
+        try:
+            response = loader(request)
```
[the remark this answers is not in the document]
kitchoi: The `except Exception` is already maximally defensive. Or maybe I misunderstood what you meant here.

ets-demo/etsdemo/loader.py
kitchoi: "name" may not exist, "root" may not exist, ... 

In fact, we can't even be sure that the entry point will return a `dict`. So actually I need to wrap this with another `try-except-log`.

This is to deal with data outside the control of `etsdemo`, so these defensive `try-except-log` are justified.

setup.py
kitchoi: The LHS is not used by etsdemo. Whether the project defining the entry points wants to use LHS is up to the projects I guess.

kitchoi:
Thank you @rahulporuri 
I ended up rewriting the `response_to_node` function to make it more robust, sorry I have to make you review this again 😂

ets-demo/etsdemo/loader.py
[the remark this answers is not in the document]
rahulporuri: Sorry. I meant - do we want to define the signature of the `loader` right now and say that it should accept `*args, **kwargs` so that in the future, we can maintain backwards compatibility without needing updates in traitsui/pyface?

e.g. if the future, we pass a new argument to `loader` functions in `etsdemo` and we don't have to immediately update pyface/traitsui to accept the arg in their `loader` functions and make a new release.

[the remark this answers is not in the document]
kitchoi: Ah, sorry I neglected to explain the intention of this `request` parameter.

It is currently an empty dict as a placeholder, for exact the reason of backward compatibility. The intention is that if in the future we need to supply more information to the `loader` function, instead of changing the function call signature, we will specify a schema for the request, like we specify a schema for the response.  In other words, the `loader` will always have just one argument.

Does this sound good?

ets-demo/etsdemo/loader.py
```diff
@@ -0,0 +1,121 @@
+# (C) Copyright 2020 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
```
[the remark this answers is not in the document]
kitchoi: Opened #1045

kitchoi:
Hmmm Travis CI has finished and passed, but we are missing the webhook response again.

ets-demo/etsdemo/loader.py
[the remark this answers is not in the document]
kitchoi: This makes me think I should add a comment here...

kitchoi:
Merging now.

@rahulporuri Let me know if you think we should revisit the function call signature before the release. See this comment: https://github.com/enthought/traitsui/pull/1023#discussion_r458198143

rahulporuri:
still LGTM

Issue #1032, opened 2020-07-20
Demo application crashes if run with pythonw on Windows


kitchoi:
To reproduce, with the current master and on Windows:
```
pythonw traitsui\examples\demo\demo.py
```
Then click on a file node on the tree. There will be a dialog saying "Python has stopped working".

After #1023 goes in, it can be reproduced with a different command:
```
pythonw -c "from etsdemo.main import main; main()"
```

For now, we can workaround this by not using `gui_scripts` for the demo main function entry point.

rahulporuri:
> For now, we can workaround this by not using gui_scripts for the demo main function entry point.

I was able to get things working by adding both a duplicate `console_scripts` which points to the same thing. That fixed the issue - not sure what was originally causing this though. (Not a 100% clear why adding the duplicate `console_scripts` entry fixed things either)

kitchoi:
Just a hypothesis and I have not verified, it could be this:
https://github.com/enthought/traitsui/blob/498ed16986c5028a5bced17f128e8625de708725/ets-demo/etsdemo/app.py#L181-L188

The point of having `gui_scripts` to avoid the ugly command-line console that gets opened on Windows. The process runs graphical interface applications without launching a system shell - that might get in the way of the demo application trying to wrap out the stdout.

PR #1038, opened 2020-07-21
Switch to console_scripts to workaround crash on Windows

kitchoi:
This is not a fix but a workaround for #1032

Crashing on Windows is bad enough we should avoid that. The one downside is that if one tries to run this application from a desktop shortcut on Windows, it will result in a command line shell being open as well (which is a bit ugly), but that is better than crashing.

### DIFF
--- a/ets-demo/setup.py
+++ b/ets-demo/setup.py
@@ -292,7 +292,7 @@ if __name__ == "__main__":
         license="BSD",
         python_requires=">=3.5",
         entry_points={
-            "gui_scripts": [
+            "console_scripts": [
                 "etsdemo = etsdemo.main:main",
             ],
             "enthought_app_metadata": [


rahulporuri:
we don't need to remove `gui_scripts` and replace it with `console_scripts`. I added a duplicate `console_scripts` which is the same as `gui_scripts` and that works on windows both on the windows terminal and via the EDM GUI (and the cmd prompt doesn't show up when launched via EDM GUI).

i'm not a 100% sure how windows automagically infers which to use - python vs pythonw but it seems to figure out the right thing to do.

kitchoi:
> we don't need to remove gui_scripts and replace it with console_scripts. I added a duplicate console_scripts which is the same as gui_scripts and that works on windows both on the windows terminal and via the EDM GUI (and the cmd prompt doesn't show up when launched via EDM GUI).

I am not sure if that just makes the `gui_scripts` definition obsolete. The two seem conflicting to me :/

kitchoi:
From https://packaging.python.org/specifications/entry-points/#use-for-scripts

> The difference between console_scripts and gui_scripts only affects Windows systems. console_scripts are wrapped in a console executable, so they are attached to a console and can use sys.stdin, sys.stdout and sys.stderr for input and output. gui_scripts are wrapped in a GUI executable, so they can be started without a console, but cannot use standard streams unless application code redirects them. Other platforms do not have the same distinction.

Looks like the standard streams limitation is indeed the cause of the issue: the demo application is trying to patch those streams (see https://github.com/enthought/traitsui/issues/1032#issuecomment-661232274).
In this case we really do want to use the console option for Windows. For other platforms, it does not matter.

kitchoi:
> we don't need to remove gui_scripts and replace it with console_scripts. I added a duplicate console_scripts which is the same as gui_scripts and that works on windows both on the windows terminal and via the EDM GUI (and the cmd prompt doesn't show up when launched via EDM GUI).

One fun thing to check is if you swap the definition of the keys in the dictionary, whether that changes the behaviour. On Python 3.6+, dictionary is ordered.

kitchoi:
> we don't need to remove gui_scripts and replace it with console_scripts. I added a duplicate console_scripts which is the same as gui_scripts and that works on windows both on the windows terminal and via the EDM GUI (and the cmd prompt doesn't show up when launched via EDM GUI).
> 
> One fun thing to check is if you swap the definition of the keys in the dictionary, whether that changes the behaviour.

Actually I just tried setting both `gui_scripts` and `console_scripts` in either order on my Windows machine. They both crash for me. Only skipping `gui_scripts` works. Not sure why it behaves differently on your machine... 🤷

rahulporuri:
Sorry. You're correct. The fix is to replace `gui_scripts` with `console_scripts`. My confusion stemmed from naming both the console script and the gui script the same thing "etsdemo" whereas they should have been "etsdemo" and "etsdemo-ui" - and then testing "etsdemo-ui" crashes as expected.

PR #1041, opened 2020-07-21
TST: Resolve warnings from QItemSelectionModel with Qt5

kitchoi:
With Qt5, running the tests for the editors result in a lot of warnings like this:
```
$ python -m unittest traitsui.tests.editors
.....QItemSelectionModel: Selecting when no model has been set will result in a no-op.
QItemSelectionModel: Selecting when no model has been set will result in a no-op.
QItemSelectionModel: Selecting when no model has been set will result in a no-op.
...
```

This PR resolves these warnings by reseting the QAbstractItemModel used in various editors's `dispose`.
Related to #431 

I have not fully understood the warnings. But it looks like the right to do: If we are disposing the widget, we should be invalidating any data or structure changes have been posted to the model. 

This also resolves some sporadic errors from the test suite I get when I add some new tests for the table editor (in another test branch of mine).

### DIFF
--- a/traitsui/qt4/file_editor.py
+++ b/traitsui/qt4/file_editor.py
@@ -227,6 +227,9 @@ class CustomEditor(SimpleTextEditor):
     def dispose(self):
         """ Disposes of the contents of an editor.
         """
+        self._model.beginResetModel()
+        self._model.endResetModel()
+
         if self.control is not None:
             self.control.doubleClicked.disconnect(self._on_dclick)
 
--- a/traitsui/qt4/list_str_editor.py
+++ b/traitsui/qt4/list_str_editor.py
@@ -196,6 +196,9 @@ class _ListStrEditor(Editor):
     def dispose(self):
         """ Disposes of the contents of an editor.
         """
+        self.model.beginResetModel()
+        self.model.endResetModel()
+
         self.context_object.on_trait_change(
             self.update_editor, self.extended_name + "_items", remove=True
         )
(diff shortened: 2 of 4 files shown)


traitsui/qt4/list_str_editor.py
```diff
@@ -196,6 +196,9 @@ def init(self, parent):
     def dispose(self):
         """ Disposes of the contents of an editor.
         """
+        self.model.beginResetModel()
+        self.model.endResetModel()
+
```
kitchoi, line 201: This one is a drive-by: After adding the reset in the other editors but without this change in `list_str_editor`, the test suite is already free of warnings. But this looks like the right to do. We probably just did not have enough coverage to trigger the same warning for `ListStrEditor`.

rahulporuri:
FTR, `beginResetModel` and `endResetModel` docs - https://doc.qt.io/qt-5/qabstractitemmodel.html#beginResetModel and https://doc.qt.io/qt-5/qabstractitemmodel.html#endResetModel

traitsui/qt4/file_editor.py
```diff
@@ -227,6 +227,9 @@ def init(self, parent):
     def dispose(self):
         """ Disposes of the contents of an editor.
         """
+        self._model.beginResetModel()
```
rahulporuri, line 230: this is tangential but i don't know why `model` is private here but public in other places

rahulporuri:
I also checked that the warnings/messages don't appear in the test logs for this PR anymore (and they do in older CI runs)

kitchoi:
For the record, I was tempted to also move the `endResetModel()` to the bottom of the `dispose` method, so the entire dispose body is wrapped inside the reset scope. But I have no data whatsoever to indicate that this is a necessary change, so I will defer this move until we have data to support it.

PR #1043, opened 2020-07-21
Remove desktop shortcut for etsdemo EAM application

kitchoi:
This PR removes the desktop shortcut defined in the EAM metadata.
Currently installing an application bundle via the EDM GUI or the command line does not have the option for skipping the shortcuts. For now, we will remove the desktop shortcut until the installer provides an option to skip it.

Closes #1043

### DIFF
--- a/ets-demo/etsdemo/info.py
+++ b/ets-demo/etsdemo/info.py
@@ -28,7 +28,6 @@ def info():
             {
                 "name": "ETS Demo",
                 "command": "etsdemo",
-                "shortcut": "desktop",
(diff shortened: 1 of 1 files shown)


rahulporuri:
Personally, I still prefer creating a desktop shortcut for the etsdemo app but I definitely don't want to create a shortcut which is broken - on any platform. I prefer the desktop shortcut because it's 1 step to launch instead of the 2-step launch EDM GUI and then launch the ets demo app. As I write it, I realize that it's probably an unnecessary optimisation.

PR #1046, opened 2020-07-21
Set an icon for the demo app main window

### LINKED ISSUE #1028: Change the window icon to something nicer in the demo application
Quote from https://github.com/enthought/traitsui/pull/1023#issuecomment-661063039:

> nit : application icon is this weird people (?) icon at the moment, even though we are shipping the Enthought E icon, which is advertised as the icon to EDM GUI

87946993-4206cf80-ca92-11ea-8699-bd0a05a815d8

kitchoi:
Closes #1028

This PR sets the application main window icon.
I decided not to use Python logo to avoid trademark issue, see https://github.com/enthought/traitsui/issues/1028#issuecomment-661980408

### DIFF
--- a/ets-demo/etsdemo/app.py
+++ b/ets-demo/etsdemo/app.py
@@ -1090,6 +1090,7 @@ class Demo(ModelView):
             resizable=True,
             width=1200,
             height=700,
+            icon=ImageResource("enthought-icon"),
         )
 
 


rahulporuri:
i tested this locally and it works as expected - the Enthought icon shows up as the app icon instead of the "people (?)" icon that we see on master

PR #1047, opened 2020-07-21
Separating Pypi long description and readme 

### LINKED ISSUE #1017: Add instructions for development
At the moment, there is no obvious way for a developer who isn't familiar with the codebase to understand/infer that the `etstool.py` file is what they need to look at to understand how to setup the development environment and run the test suite.

It would be very useful to have a small section in the README file or a longer DEVELOP.rst file which provide information for developers on setting up the development environment, running flake8 and the test suite etc.

aaronayres35:
Added an explicit separator before the Running the Test Suite section.
So now, what is rendered on Pypi will simply be the first part of the readme.  Anything found after the seperator will appear on GitHub but not Pypi (content relevant to contributors and less so users such as how to run the test suite)

fixes #1017

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -164,7 +164,7 @@ if __name__ == "__main__":
             Topic :: Software Development :: Libraries
             """.splitlines() if len(c.strip()) > 0],
         description='traitsui: traits-capable user interfaces',
-        long_description=open('README.rst').read(),
+        long_description=open("README.rst").read().split(".. end_of_long_description")[0],
         long_description_content_type="text/x-rst",
         url='http://docs.enthought.com/traitsui',
         download_url='https://github.com/enthought/traitsui',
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -164,7 +164,7 @@ def additional_commands():
             Topic :: Software Development :: Libraries
             """.splitlines() if len(c.strip()) > 0],
         description='traitsui: traits-capable user interfaces',
-        long_description=open('README.rst').read(),
+        long_description=open("README.rst").read().split(".. end_of_long_description")[0],
```
kitchoi, line 167: Same orthogonal comment as in https://github.com/enthought/pyface/pull/617#discussion_r458616860
(Commenting here too because I can't assume which one gets read first)
(Please feel free to ignore and merge away.)

aaronayres35: I will merge now and then take a look at changing this later today. Thank you!

PR #1048, opened 2020-07-21
copied license files from traitsui root directory to ets-demo

### LINKED ISSUE #1045: Some license files are missing for the etsdemo distribution
We need LICENSE.txt and the image_LICENSE_Nuvola.txt in the `etsdemo` distribution.
We can copy them from traitsui root directory.

aaronayres35:
fixes #1045 

@kitchoi can you confirm I copied these files to the right directory.  I wasn't sure if they should go to `ets-demo` or `ets-demo/estdemo`


kitchoi:
Thank you @aaronayres35! Yes this is the correct directory.

PR #1050, opened 2020-07-22
Add MANIFEST.in for ets-demo

kitchoi:
Now that we have some license files, they need to be included for the ets-demo distribution. This PR adds the MANIFEST file to include these files for the source distribution.


rahulporuri:
@aaronayres35 can you take a look at the documentation around `MANIFEST.in` and understand why it's needed?

PR #1053, opened 2020-07-22
MAINT: Use PrefixList instead of TraitPrefixList

larsoner:
Get rid of some warnings of the form:
```
/home/larsoner/python/traitsui/traitsui/ui_traits.py:60
  /home/larsoner/python/traitsui/traitsui/ui_traits.py:60: DeprecationWarning: 'TraitPrefixList' trait handler has been deprecated. Use PrefixList instead.
    "simple", TraitPrefixList("simple", "custom", "text", "readonly"), cols=4
```


larsoner:
Actually I'm not sure this is a drop-in replacement, when doing this I get errors about number of arguments

rahulporuri:
looks like the signature changed. the trait type now only accepts a list or a tuple. See PR https://github.com/enthought/traits/pull/1142

larsoner:
Indeed, pushed a commit that might work

traitsui/ui_traits.py
```diff
@@ -43,27 +43,28 @@
...
-    TraitPrefixList,
     TraitType,
 )
-
+try:
+    from traits.api import PrefixList
+except ImportError:
+    def PrefixList(args, **kwargs):
+        from traits.api import TraitPrefixList
+        return Trait(args[0], TraitPrefixList(*args), **kwargs)
```
rahulporuri: traitsui master is targeting traitsui version 7.1.0 and i think we will be able to drop support of traits 6 and set minimum version of traits to 6.1.0. Given that, i think we can remove the try/except and simply use `PrefixList`

kitchoi: I am not sure about dropping support for traits 6. We should get a confirmation before ditching this compatibility layer.

kitchoi: Let's discuss in https://github.com/enthought/traitsui/issues/1054
I think it is a great PR to eliminate the warnings and we should let this in. Maybe we want to keep this compatibility layer for now to decouple this PR with #1054?

larsoner:
Found another `TraitPrefixList` in `traitsui/view`, will push a commit for it

kitchoi:
Thank you very much @larsoner. This looks good.
One very small request... There is already a `traitsui.helper` (singular) module. This PR is going to add a `traitsui.helpers` (plural) module. Instead of having two very similarly named modules, would it make sense to either rename the new module or move the compatibility layer to the existing `traitsui.helper` module?

kitchoi:
(I updated the title to make creating changelog later a bit easier. In retrospect, I did not actually have to that - I can just modify the main message in the merge commit. Sorry about this.)

larsoner:
> There is already a traitsui.helper (singular) module.

Yikes, don't know how I missed that! Moved in latest commit

> I updated the title to make creating changelog later a bit easier

No problem!

kitchoi:
The last commit looks right to me but CI isn't happy, I am looking into this.

kitchoi:
The CI is failing at the point of loading the toolkit from `traitsui.toolkit`.

To diagnose, I had to do this:
```
import logging
logging.basicConfig(level=logging.DEBUG)
from traitsui.toolkit import toolkit
toolkit()
```
Then I can see the traceback logged:
```
...
  File "/Users/kchoi/ETS/traitsui/traitsui/helper.py", line 28, in <module>
    from .ui_traits import SequenceTypes
  File "/Users/kchoi/ETS/traitsui/traitsui/ui_traits.py", line 48, in <module>
    from .helper import PrefixList
```

By moving the `PrefixList` to `helper`, a cycle import between `ui_traits` and `helper` is created, because `helper` imports from `ui_traits` too.
https://github.com/enthought/traitsui/blob/796bdb973cc3f2c77eb3097ff9f2e8b8f86ff11c/traitsui/helper.py#L28

 We can solve this by moving the definition of `SequenceTypes` across to `helper` and then in `ui_traits`, import both `PrefixList` and `SequenceTypes`.

PR #1057, opened 2020-07-23
TST: Add tests for ui_traits module

kitchoi:
This PR adds unit tests for the `ui_traits` module. 

Trait types and objects defined in this module are probably indirectly tested in places with more dependencies (e.g. integration tests). This PR adds unit tests that have very minimal dependency, making it easier to maintain.

### DIFF
--- a/traitsui/tests/test_ui_traits.py
+++ b/traitsui/tests/test_ui_traits.py
@@ -1,0 +1,83 @@
+# (C) Copyright 2005-2020 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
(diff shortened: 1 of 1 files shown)


kitchoi:
I did not try to go for full coverage: This was motivated by the effort in trying to resolve deprecation warnings.

rahulporuri:
> I did not try to go for full coverage:

Do you think it makes sense to create an issue to add the rest of the tests?

kitchoi:
> Do you think it makes sense to create an issue to add the rest of the tests?

Up to you. For me, I add tests based on needs, e.g. when I need to make a change to something that does not have good tests. In this case, I added the tests because of a recent PR.

kitchoi:
> In this case, I added the tests because of a recent PR.

To clarify, that is the PR(s) for resolving the deprecation warnings. Therefore I added enough tests to verify for myself that the deprecation warnings are resolved.

PR #1059, opened 2020-07-23
TST: Switch on default warning flag for CI test command

kitchoi:
Deprecation warnings triggered during test runs are visible, but deprecation warnings triggered prior to when a unittest test case is loaded are by default hidden from the console.

With that, deprecation warnings from trait types definition do not show up from running the current test command, because the warnings occur at import time.

This PR adds the `-W default` flag to make these warnings visible so that we can fix them more easily.


rahulporuri:
ref : https://docs.python.org/3/using/cmdline.html#cmdoption-w

PR #1064, opened 2020-07-24
Convert extracting of long_description from readme to a context manager

aaronayres35:
This implements a suggestion made on PR #1047

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -130,6 +130,8 @@ if __name__ == "__main__":
     data = read_module('__init__')
     __requires__ = data['__requires__']
     __extras_require__ = data['__extras_require__']
+    with open("README.rst", "r", encoding="utf-8") as readme:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -130,6 +130,8 @@ def write_version_py(filename=None):
     data = read_module('__init__')
     __requires__ = data['__requires__']
     __extras_require__ = data['__extras_require__']
+    with open("README.rst", "r", encoding="utf-8") as readme:
+        LONG_DESCRIPTION=readme.read().split(".. end_of_long_description")[0]
```
kitchoi, line 134: 🤖  TraitsUI and Pyface don't have flake8 in CI... Flake8 would have complained about the missing spaces around the equal sign...  Not your fault - we should have the machine nagging us about this sort of thing.

Issue #1066, opened 2020-07-24
Problem with Scrubber_editor_demo.py


aaronayres35:
Running on wx gives 

<details>

```
$ python Scrubber_editor_demo.py 
01:03:59 PM: Debug: Adding duplicate image handler for 'Windows bitmap file'
01:03:59 PM: Debug: Adding duplicate animation handler for '1' type
01:03:59 PM: Debug: Adding duplicate animation handler for '2' type
01:03:59 PM: Debug: Adding duplicate image handler for 'Windows bitmap file'
01:03:59 PM: Debug: Adding duplicate animation handler for '1' type
01:03:59 PM: Debug: Adding duplicate animation handler for '2' type
01:03:59 PM: Debug: Adding duplicate image handler for 'Windows bitmap file'
01:03:59 PM: Debug: Adding duplicate animation handler for '1' type
01:03:59 PM: Debug: Adding duplicate animation handler for '2' type
Traceback (most recent call last):
  File "/Users/aayres/.edm/envs/wx-env/lib/python3.6/site-packages/traitsui/wx/view_application.py", line 152, in OnInit
    args=self.args,
  File "/Users/aayres/.edm/envs/wx-env/lib/python3.6/site-packages/traitsui/view.py", line 462, in ui
    ui.ui(parent, kind)
  File "/Users/aayres/.edm/envs/wx-env/lib/python3.6/site-packages/traitsui/ui.py", line 246, in ui
    self.rebuild(self, parent)
  File "/Users/aayres/.edm/envs/wx-env/lib/python3.6/site-packages/traitsui/wx/toolkit.py", line 125, in ui_live
    ui_live.ui_live(ui, parent)
  File "/Users/aayres/.edm/envs/wx-env/lib/python3.6/site-packages/traitsui/wx/ui_live.py", line 52, in ui_live
    _ui_dialog(ui, parent, BaseDialog.NONMODAL)
  File "/Users/aayres/.edm/envs/wx-env/lib/python3.6/site-packages/traitsui/wx/ui_live.py", line 88, in _ui_dialog
    BaseDialog.display_ui(ui, parent, style)
  File "/Users/aayres/.edm/envs/wx-env/lib/python3.6/site-packages/traitsui/wx/ui_base.py", line 71, in display_ui
    ui.owner.init(ui, parent, style)
  File "/Users/aayres/.edm/envs/wx-env/lib/python3.6/site-packages/traitsui/wx/ui_live.py", line 203, in init
[...]

Issue #828, opened 2020-05-18
Illegal flag in wx/ui_panel.py. FIX INCLUDED


jmoraleda:
Recent versions of wxWidgets are more strict about not allowing positioning flags that have no effect so one gets the error:

```
  File "/usr/lib/python3/dist-packages/traits/has_traits.py", line 2019, in edit_traits
    return view.ui(
  File "/usr/lib/python3/dist-packages/traitsui/view.py", line 446, in ui
    ui.ui(parent, kind)
  File "/usr/lib/python3/dist-packages/traitsui/ui.py", line 244, in ui
    self.rebuild(self, parent)
  File "/usr/lib/python3/dist-packages/traitsui/wx/toolkit.py", line 115, in ui_subpanel
    ui_panel.ui_subpanel(ui, parent)
  File "/usr/lib/python3/dist-packages/traitsui/wx/ui_panel.py", line 101, in ui_subpanel
    ui_panel_for(ui, parent, False)
  File "/usr/lib/python3/dist-packages/traitsui/wx/ui_panel.py", line 115, in ui_panel_for
    ui.control = control = Panel(ui, parent, buttons).control
  File "/usr/lib/python3/dist-packages/traitsui/wx/ui_panel.py", line 202, in __init__
    sw = panel(ui, cpanel)
  File "/usr/lib/python3/dist-packages/traitsui/wx/ui_panel.py", line 306, in panel
    sg_sizer, resizable, contents = fill_panel_for_group(
  File "/usr/lib/python3/dist-packages/traitsui/wx/ui_panel.py", line 467, in fill_panel_for_group
    fp = FillPanel(panel, group, ui, suppress_label, is_dock_window,
  File "/usr/lib/python3/dist-packages/traitsui/wx/ui_panel.py", line 613, in __init__
    self.add_items(content, panel, self.sizer)
  File "/usr/lib/python3/dist-packages/traitsui/wx/ui_panel.py", line 1041, in add_items
    item_sizer.Add(control, growable,
wx._core.wxAssertionError: C++ assertion "!(flags & wxALIGN_CENTRE_VERTICAL)" failed at /home/jorge/vc/Phoenix/ext/wxWidgets/src/common/sizer.cpp(2109) in DoInsert(): Vertical alignment flags are ignored in vertical sizers
```

### FIX ###

To fix it, it suffices to remove `| wx.ALIGN_CENTER_VERTICAL` from line 1042 of `wx/ui_panel.ui`

Issue #434, opened 2018-03-14
ScrubberEditor segfault on Wx


corranwebster:
Get segfault on initial mouseover when running `Scrubber_editor_demo.py`

PR #1067, opened 2020-07-24
remove the ALIGN_CENTRE_VERTICAL FLAG to avoid error

aaronayres35:
This gets rid of the 
```
wx._core.wxAssertionError: C++ assertion "!(flags & (wxALIGN_BOTTOM | wxALIGN_CENTRE_VERTICAL))" failed at /Users/robind/projects/bb2/dist-osx-py36/build/ext/wxWidgets/src/common/sizer.cpp(2114) in DoInsert(): Vertical alignment flags are ignored with wxEXPAND
```
error mentioned in issue #1066  (another example of #828).  However, after this fix is completed, the issue of a segfault on mouseover as mentioned in #434 still occurs.

### DIFF
--- a/traitsui/wx/scrubber_editor.py
+++ b/traitsui/wx/scrubber_editor.py
@@ -333,8 +333,6 @@ class _ScrubberEditor(Editor):
         drop_down.Bind(wx.EVT_KILL_FOCUS, self._enum_completed)
 
     def _pop_up_text(self):
-        print("Im popping up text")
-        print(self.control)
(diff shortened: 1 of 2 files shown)


kitchoi:
Thanks for the fix @aaronayres35! Nice catch.
I opened #1089 to add a test that reproduces the bug, this does two things: (1) confirm the change here fixes the issue (2) prevent the issue from happening again.
Could you please review and see if it can be merged to this PR?

PR #1083, opened 2020-07-29
Fix resizable dialog not being shown on top of parent dialog in Linux

### LINKED ISSUE #1078: Resizeable windows opened with edit_traits() are forced to the background
Hello everyone,

Since the latest update (TraitsUI 7.0.1-1), I have noticed what I think is unwanted behaviour.

Whenever I open a second window with the `edit_traits()` function, and the resizable property of this View is set to True, the window is hidden behind the first window (which was called through `configure_traits()`).

Below is a demo case. Try opening a second window through the menu bar of the 'Main' window, this should trigger the bug. I am using PyQt5:

```
from traitsui.api import Handler, View, Item
from traits.api import HasTraits, String, Int
from traitsui.menu import Menu, Action, CloseAction, MenuBar

the_action = Action(
    name = "Open a window",
    action = "_do_something"
)

class Sub(HasTraits):
    a_trait = Int
    
        return View(
            Item("a_trait"),
            resizable = True
        )
    
class DemoHandler(Handler):
    def _do_something(self, uiinfo):
        sub_window = Sub()
        sub_window.edit_traits()

class Main(HasTraits):
    name = String

    traits_view = View(
        Item("name"),
        menubar = MenuBar(
            Menu(
                the_action,
                name = "A menu"
            )
        ),
        handler = DemoHandler()
    )

main = Main()
main.configure_traits()
```

Kind regards,

Matthias

kitchoi:
Fixes #1078

If the view's `resizable` flag is set to `True`, `_StickyDialog` (the widget used for the UI's dialog) sets the window flags to be `Qt.Window`. If `resizable` is `False`, then the flag includes `Qt.Dialog`. 
`Qt.Window` is a narrower flag compared to `Qt.Dialog`. In the case of #1078, since the second dialog does not have the parent set (`edit_traits` is not given a `parent` argument), the second "dialog" is treated as a second top-level window rather than a child dialog from the first window. Then it is up to the OS window manager to decide which window should be raised.

This PR fixes the issue by making sure the window flags include `Qt.Dialog`.
Note that `Qt.Dialog` equals to `Qt.WindowMaximized | Qt.Window`


kitchoi:
>  In the case of #1078, since the second dialog does not have the parent set (edit_traits is not given a parent argument).

I have also tried setting the `parent` to `uiinfo.ui.control` in the `edit_traits` call mentioned in #1078. While it would be a correct thing to do (I think), that does not fix the behaviour on Linux.

traitsui/qt4/tests/test_ui_base.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) 2019, Enthought Inc.
```
rahulporuri, line 1: nit - copyright header year.

traitsui/qt4/ui_base.py
```diff
@@ -128,16 +128,10 @@ def __init__(self, ui, parent):
...
         # Set the dialog's window flags and properties.
-        if ui.view.resizable:
-            flags = QtCore.Qt.Window
-        else:
-            flags = QtCore.Qt.Dialog | QtCore.Qt.WindowSystemMenuHint
+        flags = QtCore.Qt.Dialog
+        if not ui.view.resizable:
+            flags |= QtCore.Qt.WindowSystemMenuHint
             layout.setSizeConstraint(QtGui.QLayout.SetFixedSize)
-            if ui.view.resizable:
```
kitchoi, line 136: Note: This removal is actually a drive-by. This branch is never reached because it is in the else branch where the condition is false.

rahulporuri: 🤣  yeah i was wondering why that branch existed

rahulporuri: i just dug through history to find a reason for why this exists and i can see the commit where this was added but i don't see a PR associated with the commit/change - https://github.com/enthought/traitsui/commit/b61ab16c153375103923ecf0d72f15f15419f51e

kitchoi: 🤷

PR #1084, opened 2020-07-29
fixing broken link

### LINKED ISSUE #1077: [on master] Broken links in demos page in docs
The "demos" page in the traitsui docs points to the code on github for the various examples on github. BUT. the code got moved recently to now live inside the traitsui package instead of at the top-level of the github repository.

https://github.com/enthought/traitsui/blob/f3a757638e085b7b175cb1d2bcafdc2898f6dbba/docs/source/conf.py#L192-L196

An alternative - to ensure that the links are never broken but possibly outdated - is to permalink to a specific commit when the page was last updated/verified.

aaronayres35:
fixes #1077

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -189,9 +189,7 @@ except ImportError as exc:
 # #html_style = 'default.css'
 # #html_theme = 'classic'
 
-# Useful aliases to avoid repeating long URLs.
-# Note this is a permalink to a specific commit to ensure the link 
-# is never broken, but it may become outdated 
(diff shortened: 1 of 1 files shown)


rahulporuri:
@kitchoi what do you think? Is this the right approach or would it be better to let links break to ensure that we keep checking and fixing them.

kitchoi:
While we would all prefer an up-to-date content and working links, if I had to choose between broken links and valid link but outdated content, I'd choose broken links. But that might just be me.

PR #1090, opened 2020-07-30
Configure logging for etsdemo main function

### LINKED ISSUE #1026: Configure logging for the demo application main function
When the ETS demo application is launched in the production setting, we should configure logging. Currently the main function does not do that:
https://github.com/enthought/traitsui/blob/6fccc3966e89a8082aace9623b84a80a74c46f3e/ets-demo/etsdemo/main.py#L22-L27

As a first step, it should log to console.
(In the future it could be extended to logging to files or allow the log level to be modified)

As usual, the log configuration should be implemented as a context manager so that the log configuration is reset when the function exit.

kitchoi:
Closes #1026

This PR adds a basic context manager for configuring logger in the main function such that the configuration is cleaned up upon exiting the context. Currently we only log to console.
Command line options are added for setting the log level.

### DIFF
--- a/ets-demo/etsdemo/main.py
+++ b/ets-demo/etsdemo/main.py
@@ -10,13 +10,69 @@
 
 """ The main function for launching the demo application.
 """
+import argparse
+import contextlib
+import logging
 
 from etsdemo.app import Demo, DemoVirtualDirectory
 from etsdemo.loader import get_responses, response_to_node
 
 
+#: Logging constants
+_LOG_FORMAT = "%(asctime)s %(levelname)-8.8s [%(name)s:%(lineno)s] %(message)s"
+logger = logging.getLogger(__name__)
+
 #: Default application title
 _TITLE = "Enthought Tool Suite"
+
+_THIS_PACKAGE, _ = __name__.split(".", 1)
+
+
+@contextlib.contextmanager
+def _set_logger(logger, level):
+    """ Context manager for setting logging.
+    """
+    formatter = logging.Formatter(_LOG_FORMAT)
+    handler = logging.StreamHandler()
+    handler.setLevel(level)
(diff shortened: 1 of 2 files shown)


ets-demo/etsdemo/tests/test_main.py
```diff
@@ -45,8 +46,18 @@ class TestMain(unittest.TestCase):
...
+        with mock_iter_entry_points({}), \
+                mock_demo_launch(), \
+                mocked_io, \
+                mock.patch("sys.stdout", mocked_io), \
+                mock.patch("sys.stderr", mocked_io), \
+                mock.patch("sys.argv", argv):
             main_module.main()
+            log_content = mocked_io.getvalue()
+
+        self.assertIn("Found 0 resource(s).", log_content)
```
kitchoi, line 60: I think I need to add a test instead of modifying an existing one. One test for the normal case where we have no logging. The other one with some logging info.

kitchoi: Done.

ets-demo/etsdemo/main.py
```diff
@@ -10,14 +10,76 @@
...
+        Parsed command line arguments.
+    """
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--debug',
+        help="Log debugging information to console.",
+        action="store_const",
+        dest="log_level",
+        const=logging.DEBUG,
+        default=logging.WARNING,
```
rahulporuri, line 72: I think i finally understand what's happening here. Not passing `--debug` sets `log_level` to `logging.WARNING` and passing `--debug` sets `log_level` to `logging.DEBUG`

kitchoi: That's right.

PR #1092, opened 2020-07-31
TST: Replace Qt4 with Qt in tests and clarify test requirements

### LINKED ISSUE #978: Clarify which version of Qt is skipped/required in tests
Currently the `is_current_backend_qt4` test utility in `traitsui.tests._tools` refer to both Qt4 and Qt5. This is very confusing, we should differentiate the two versions and possibly create another convenient wrapper for where we only care about whether Qt is used regardless of its version.

kitchoi:
Closes #978

It has been a constant source of confusion that `is_current_backend_qt4` actually includes both Qt4 and Qt5. In fact, it is intended for Qt, regardless of its version.

This inevitably has to change a lot of lines of codes in trivial ways. While I am at this, I took the opportunity to flip "skip_if_toolkit" and "skip_if_not_toolkit" with positive assertions. Skips are confusing to read, and `skip_if_not_*` can only exclude one toolkit. Sometimes we use `skip_if_null` to actually mean "this test targets Qt and Wx". For we have three toolkits, skipping "null" gives us the other two. If in the future we should introduce a fourth toolkit support incrementally, this `skip_if_` logic won't work.

In summary, this PR clarifies test requirements with the following changes:
(1) Replace `is_current_backend_qt4` with `is_qt`. This will just mean Qt, regardless of Qt version. A different naming pattern is used to avoid confusion. Likewise, `is_current_backend_wx` is replaced with `is_wx`.
(2) Introduce `@requires_toolkit` to replace `@skip_if_toolkit` and `@skip_if_not_toolkit` decorator. This decorator specifies the toolkit requirements for a test. An enum `ToolkitName` is introduced for supplying toolkit names to `requires_toolkit`.

Note:
I did not actually remove `is_current_backend_*` and `skip_if_*` in this PR. This is to minimize disruptions to other PRs in the meantime while this is reviewed. These functions should be removed when other PRs have settled with the new utility function. This is an internal deprecation.

Discussion:
- Naming: `@requires_toolkit` or `@requires_one_of_toolkits` or `@requires_toolkit_in` or `@support_toolkits` or...?

### DIFF
--- a/traitsui/tests/_tools.py
+++ b/traitsui/tests/_tools.py
@@ -121,7 +121,7 @@ def _deprecated(func):
     def wrapped(*args, **kwargs):
         warnings.warn(
             "{!r} will be removed. "
-            "Use is_wx, is_qt or is_null or requires_toolkit instead.".format(
+            "Use is_wx, is_qt, is_null or requires_toolkit instead.".format(
                 func
             ),
             DeprecationWarning,
--- a/traitsui/tests/_tools.py
+++ b/traitsui/tests/_tools.py
@@ -13,13 +13,14 @@
 #
 # ------------------------------------------------------------------------------
 
-
+import enum
 import re
 import sys
 import traceback
 from functools import partial
 from contextlib import contextmanager
 from unittest import skipIf, TestSuite
+import warnings
 
 from pyface.api import GUI
 from pyface.toolkit import toolkit_object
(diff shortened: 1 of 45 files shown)


traitsui/tests/editors/test_liststr_editor_selection.py
```diff
@@ -193,16 +192,16 @@ def right_click_item(control, index):
...
         raise unittest.SkipTest("Test not implemented for this toolkit")
 
     else:
         raise unittest.SkipTest("Test not implemented for this toolkit")
 
 
-@unittest.skipIf(is_current_backend_wx(), "Issue enthought/traitsui#752")
-@skip_if_null
+@unittest.skipIf(is_wx(), "Issue enthought/traitsui#752")
+@requires_toolkit([ToolkitName.qt, ToolkitName.wx])
```
kitchoi, line 204: The second `ToolkitName.wx` looks redundant now, but I am going to leave it...

traitsui/tests/_tools.py
```diff
@@ -67,33 +68,102 @@ def handle_exception(object, trait_name, old, new):
...
+        warnings.warn(
+            "{!r} will be removed. "
+            "Use is_wx, is_qt, is_null or requires_toolkit instead.".format(
+                func
+            ),
+            DeprecationWarning,
+            stacklevel=3,
+        )
+        return func(*args, **kwargs)
+    return wrapped
```
aaronayres35, line 131: Could you explain how this function works after the standup today?

traitsui/tests/_tools.py
```diff
@@ -67,33 +68,102 @@ def handle_exception(object, trait_name, old, new):
...
+
+
 def _is_current_backend(backend_name=""):
     return ETSConfig.toolkit == backend_name
 
 
 #: Return True if current backend is 'wx'
-is_current_backend_wx = partial(_is_current_backend, backend_name="wx")
+is_current_backend_wx = _deprecated(
```
aaronayres35, line 139: for example, here _deprecated is called on the old version of the code

traitsui/tests/_tools.py
```diff
@@ -67,33 +68,102 @@ def handle_exception(object, trait_name, old, new):
...
-is_current_backend_null = partial(_is_current_backend, backend_name="null")
-
+is_current_backend_null = _deprecated(
+    partial(_is_current_backend, backend_name="null")
+)
 
-#: Test decorator: Skip test if backend is not 'wx'
-skip_if_not_wx = skipIf(not is_current_backend_wx(), "Test only for wx")
+#: Test decorator: Skip test if backend is not wx
+skip_if_not_wx = _deprecated(requires_toolkit([ToolkitName.wx]))
```
aaronayres35, line 154: but here its called on the new code

kitchoi: I can also change this to `_deprecated(skipIf(not is_wx(), "Test only for wx"))` actually.
The purpose of the change is to avoid calling `is_current_backend_wx()` here, which will trigger the deprecation warning.

traitsui/tests/_tools.py
kitchoi: (Discussed offline)

PR #1100, opened 2020-08-03
Maint: Specify wx being required by a test

kitchoi:
This PR switches `skip_if_not_wx` to `requires_toolkit` with wx.

The test was added in a PR when the PR introducing `requires_toolkit` was pending.

I believe this is the last occurrence of `skip_if_not_`. After confirming no other PRs are using it, we can remove the placeholder.

### DIFF
--- a/traitsui/tests/editors/test_animatedGIF_editor.py
+++ b/traitsui/tests/editors/test_animatedGIF_editor.py
@@ -19,7 +19,8 @@ from traits.api import Bool, File, HasTraits
 from traitsui.api import Item, View
 from traitsui.tests._tools import (
     create_ui,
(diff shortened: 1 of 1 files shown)


traitsui/tests/editors/test_animatedGIF_editor.py
```diff
@@ -39,7 +40,7 @@ def test_animated_gif_editor(self):
         view = View(
             Item('gif_file', editor=AnimatedGIFEditor(playing='playing')),
         )
-        
+
```
kitchoi, line 43: This is my IDE's doing, not me.
(The change complies with flake8 so I am keeping it.)

PR #1103, opened 2020-08-03
using collections.abc instead of collections

### LINKED ISSUE #1098: Deprecation warnings with Python 3.8 on collections ABC
Running the test suite with Python 3.8 reveals the following warning:
```
test_undo (traitsui.tests.test_undo.TestEditorUndo) ... /Users/kchoi/Work/ETS/traitsui/traitsui/undo.py:154: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
  elif isinstance(v1, collections.Sequence):
```

aaronayres35:
fixes #1098 
Since we no longer support python 2 I believe this shouldn't be a problem

### DIFF
--- a/traitsui/undo.py
+++ b/traitsui/undo.py
@@ -21,7 +21,7 @@
 
 
 
-import collections
+import collections.abc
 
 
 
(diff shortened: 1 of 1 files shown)


rahulporuri:
I see a couple of try/excepts for this import in pyface. Do you have time to fix those as well?

PR #1104, opened 2020-08-03
fixing some simple flake8 errors

aaronayres35:
I was stuck on most of my other issues, and wanted to try to make some form of a tangible contribution.  Rahul had mentioned flake8 isn't currently being used on traitsui and pyface, but that ideally it would be.  I decided to try and do a style clean up and fix a bunch of flake8 errors.  I quickly realized my efforts were futile, at least for now, and stopped here.  Regardless I thought I would open a PR as even some style fixes can't hurt (although I apologize for the tedious PR review).

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -122,7 +122,7 @@ test_dependencies = {
 extra_dependencies = {
     # XXX once pyside2 is available in EDM, we will want it here
     'pyside2': set(),
-    'pyqt': {'pyqt<4.12'}, # FIXME: build 1 of 4.12.1 appears to be bad
+    'pyqt': {'pyqt<4.12'},  # FIXME: build 1 of 4.12.1 appears to be bad
     'pyqt5': {'pyqt5'},
     # XXX once wxPython 4 is available in EDM, we will want it here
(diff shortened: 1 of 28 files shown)


kitchoi:
Thank you very much!  I think doing this incrementally is not a bad idea. The alternative is to use tools like autopep8 and black to lint the entire code base, but it often causes cross-PR disruptions.
As we do so, it would be worth opting-in files (or opting-out depending on the state of affairs) to be linted by flake8 so we prevent them from deteriorated (see [configuration options](https://flake8.pycqa.org/en/latest/user/configuration.html)).

rahulporuri:
> Thank you very much! I think doing this incrementally is not a bad idea. The alternative is to use tools like autopep8 and black to lint the entire code base, but it often causes cross-PR disruptions.
As we do so, it would be worth opting-in files (or opting-out depending on the state of affairs) to be linted by flake8 so we prevent them from deteriorating (see configuration options).

I fully agree.

kitchoi:
Opened https://github.com/enthought/traitsui/issues/1105

PR #1108, opened 2020-08-04
Add a CI job for testing against traits 6.0

### LINKED ISSUE #1054: Continuous integration test for support of Traits 6.0
Currently CI tests against Traits 6.1.0 and does not test against Traits 6.0.0, and it would be easy to accidentally use new features of Traits 6.1.0 without realizing that would be in effect dropping support against Traits 6.0.

Actions:
- Decide if the next release can drop support for Traits 6.0
- If not, add a CI job to test against Traits 6.0

kitchoi:
Closes #1054

This PR adds a job for the Travis build to test against Traits 6.0.
I did not multiply the entire matrix with Traits 6.0 because I did not find this necessary. Given most of the GUI tests are skipped on Null or Wx backend, choosing one of the Qt backend exercise most logic that would depend on traits 6.1. Then the choice of PyQt5 over PySide2 is a bit arbitrary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ matrix:
   include:
     - env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5 pyside2"
     - env: RUNTIME=3.6 TOOLKITS="wx"
+    - env: RUNTIME=3.6 TOOLKITS="pyqt5" TRAITS_REQUIRES="^=6.0"
     - os: osx
       env: RUNTIME=3.6 TOOLKITS="null pyqt pyqt5 pyside2"
     - os: osx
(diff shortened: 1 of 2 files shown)


kitchoi:
Confirmed the new job is running with Traits 6.0
In particular, we do have a new test that depends on Traits 6.1 and is skipped:
```
test_format_func_used (traitsui.tests.editors.test_text_editor.TestTextEditor) ... skipped 'This test requires traits >= 6.1.0'
```

kitchoi:
And currently the CI job for Wx is allowed to fail, so I also don't see much value in including wx for this new CI job. In fact, it is better to exclude wx because the job can be required to pass.

PR #1112, opened 2020-08-06
Relax test assertion for etsdemo main function

kitchoi:
The test for running `main` in the normal condition is too strict: It could fail if there were warnings ~printed to the console through code outside of our control~ (Edited:  but the test is about testing logging settings.)
For purpose of the test, we can just assert the log message (asserted in the next test) is not emitted for the normal run condition.

### DIFF
--- a/ets-demo/etsdemo/tests/test_main.py
+++ b/ets-demo/etsdemo/tests/test_main.py
@@ -47,7 +47,7 @@ class TestMain(unittest.TestCase):
     @require_gui
     def test_main(self):
         # Main function must be launchable even if there are no data available.
-        # In normal running situation, the console should be clean
+        # In normal running situation, no loggings should be emitted
         argv = ["etsdemo"]
         mocked_io = io.StringIO()
         with mock_iter_entry_points({}), \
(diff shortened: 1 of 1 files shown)


kitchoi:
When we are in Python 3.10, we can use `unittest.TestCase.assertNoLogs`. Checking for a log message that is asserted in the next test seems good enough for now.

ets-demo/etsdemo/tests/test_main.py
```diff
@@ -59,7 +59,7 @@ def test_main(self):
             main_module.main()
             console_output = mocked_io.getvalue()
 
-        self.assertEqual(console_output, "")
+        self.assertNotIn("Found 0 resource(s).", console_output)
```
aaronayres35, line 62: Where exactly does "Found 0 resource(s)." come from?  The idea is that no loggings should be emitted, does this mean that if any log is emitted, this will always be in the console_output?

kitchoi: Offline discussion:
The message comes from https://github.com/enthought/traitsui/blob/9ea583c7d69a32dc4a6d7896529274161c0cab70/ets-demo/etsdemo/main.py#L109

The description of this PR is a bit misleading: Not only that we don't want the test to fail because of warnings from code outside of our control, this particular test is targeting the logging settings, so we don't want it to fail even if there are warnings from our code. Even though such warnings are undesirable, they are not the objective of this test.
I fixed the description.

Thanks @aaronayres35!

PR #1113, opened 2020-08-06
TST: Remove --debug from the eam commmand which isn't available on eam 1.1.0

kitchoi:
The current latest version of eam on PyPI is 1.1.0, which does not have this `--debug` option.
This means the test would fail if one installs `etsdemo` and its dependencies all via PyPI.
This PR fixes this by removing the option. The option is not required for the test objective: It only serves to make debugging test failure a bit easier.

This issue has not appeared on CI because `eam` is installed via EDM. And the latest version available has this `--debug` option.


rahulporuri:
can you open an issue in the eam repository about uploading the new eggs to PyPI?

kitchoi:
> can you open an issue in the eam repository about uploading the new eggs to PyPI please?

Yep. That's done.

Issue #868, opened 2020-06-04
ListStrEditor behaviour discrepancies between qt and wx versions


ievacerny:
`ListStrEditor` behaviour differs in quite a few places between qt and wx implementations:

1. Single select editor initial selection: 
    - QT: No initial selection
    - WX: First item is selected
2. Single select editor, setting `selected_index=-1`:
   - QT: Selection is cleared
   - WX: All items are selected visually, but `selected` isn't updated
3. Multi select editor, invalid value handling when assigning to `multi_selected`:
   - QT: Any invalid values that are assigned to `multi_selected` are ignored during the rest of the update
   - WX: No other updated happens
4. Updating editor, both single select and multi select:
   - QT: Uses values of `selected` or `multi_selected` as the source of truth for the update
   - WX: Uses values of `selected_index` or `multi_selected_indices` as the source of truth for the update
5. Context menu, right click handling
   - QT: Supports the addition of a context menu, but doesn't register the right clicks via `right_clicked` and `right_clicked_index`
    - WX: Doesn't support the addition of a context menu, but does register the right clicks via `right_clicked` and `right_clicked_index`
6. `setx` and `callx` methods only exist in QT editor

It's quite likely that there are even more differences between the implementations hiding in the code.

Issue #175, opened 2014-08-05
ListStrEditor `context menu` behaviour probably broken 


itziakos:
It looks that the behaviour regarding right_clicking and context_menus is probably broken. 
- The are two  methods named `_on_context_menu` defined in the file. As a result the `right_click` functionality is discarded (see https://github.com/enthought/traitsui/blob/master/traitsui/qt4/list_str_editor.py#L350 and https://github.com/enthought/traitsui/blob/master/traitsui/qt4/list_str_editor.py#L388).
- It also looks that contrary to the docstring in ListStrEditor about using the default context menu when `menu=None` the `setContextMenuPolicy` is always set to `QtCore.Qt.CustomContextMenu` (see https://github.com/enthought/traitsui/blob/master/traitsui/qt4/list_str_editor.py#L474)

PR #1115, opened 2020-08-06
Fix updating right_clicked and right_clicked_item for Qt ListStrEditor

kitchoi:
This issue fixes this part of #868:

> Supports the addition of a context menu, but doesn't register the right clicks via right_clicked and right_clicked_index

This was because of this half of #175:

> The are two methods named _on_context_menu defined in the file. As a result the right_click functionality is discarded

This PR fixes both by (1) renaming the overridden `_on_context_menu` to `_on_mouse_right_click` and (2) hook it up with `QListView.mouseReleaseEvent`.
Using context menu signal for right click events is an abuse: It makes testing hard.  By hooking this handler up with `mouseReleaseEvent`, I can test the functionality with `QTest.mouseClick`.
Furthermore, the `menu` option on the editor can be switched off by setting the trait to `False`. The `right_clicked` feature should be independent of the context menu feature.

(Note: It would also have been possible to test the combined `_on_context_menu` and `_on_mouse_right_click` had it not because of https://bugreports.qt.io/browse/QTBUG-52552).
(Note: It was a yak-shaving exercise. I initially planned to fix the trivial deprecation warning for `collections.Callable`, realizing that that line of code is not covered by test, then I went to see if I could get it under test... so on and on. The `_on_context_menu` is still untested in this PR, but at least we fix another behaviourial bug.)
(Note: I noticed that the wx ListStrEditor does not handle right clicking an area where there is no item. The new test `test_list_str_editor_right_click_out_of_bound ` will fail on wx with an IndexError. But the test is also skipped due to #752. Since it is an entirely orthogonal issue, I decided not to mix the fix in and I will open a separate issue.)


traitsui/tests/editors/test_liststr_editor_selection.py
```diff
@@ -621,12 +657,35 @@ class ListStrModelRightClick(HasTraits):
...
             self.assertEqual(model.right_clicked_index, 0)
 
-            right_click_item(editor.control, 1)
+            right_click_item(editor, 1)
             process_cascade_events()
 
             self.assertEqual(model.right_clicked, "two")
             self.assertEqual(model.right_clicked_index, 1)
 
+    def test_list_str_editor_right_click_out_of_bound(self):
```
kitchoi, line 666: This tests the `if mi.isValid():` logic in `_on_mouse_right_click`.

kitchoi:
> (Note: I noticed that the wx ListStrEditor does not handle right clicking an area where there is no item. The new test test_list_str_editor_right_click_out_of_bound will fail on wx with an IndexError. But the test is also skipped due to #752. Since it is an entirely orthogonal issue, I decided not to mix the fix in and I will open a separate issue.)

Opened #1116

traitsui/qt4/list_str_editor.py
```diff
@@ -388,8 +388,8 @@ def _on_activate(self, mi):
         self.activated_index = index = mi.row()
         self.activated = self.adapter.get_item(self.object, self.name, index)
 
-    def _on_context_menu(self, point):
-        """ Handle a context menu request.
+    def _on_mouse_right_click(self, point):
```
rahulporuri, line 391: Umm. I'm not sure why there are two `_on_context_menu` methods defined on `_ListStrEditor`.

If this is the right approach, do we also want to not hook up the non-existent `_on_context_menu` method in `_ListView.__init__`?

kitchoi: > do we also want to not hook up the non-existent _on_context_menu method in _ListView.__init__?

I might have misunderstood what you meant... let me try to clarify what is happening here.

The second `_on_context_menu` (now the only `_on_context_menu`) is hooked up in the `_ListView.__init__` already.

This `_on_mouse_right_click` (previously called `_on_context_menu` as well) was as if it was never there. Now it is hooked up with the `_ListView.mouseReleaseEvent`. If we don't hook it up, then we won't be fixing that part of #868 I quoted.

What happened here is that this `_on_mouse_right_click` used to be the only `_on_context_menu`. It was hooked up with the context menu event because right clicking the widget fires that event too (among many other events). Later on we wanted additional control on the context menu behaviour, and added another `_on_context_menu`. And because there was no flake8 telling us the method is defined twice, the first one is effectively removed.

rahulporuri (approved):
LGTM

kitchoi:
I will merge master to this branch to trigger CI since the last build was quite a while ago.

PR #1129, opened 2020-08-10
Fix deprecation warnings regarding collections.abc in ListStrEditor and TreeEditor

kitchoi:
Fixes two other occurrences of https://github.com/enthought/traitsui/issues/1098

### DIFF
--- a/traitsui/qt4/list_str_editor.py
+++ b/traitsui/qt4/list_str_editor.py
@@ -18,11 +18,11 @@
 """ Traits UI editor for editing lists of strings.
 """
 
-
-from pyface.qt import QtCore, QtGui, is_qt5
-import collections
+import collections.abc
 
 from pyface.image_resource import ImageResource
+from pyface.qt import QtCore, QtGui, is_qt5
+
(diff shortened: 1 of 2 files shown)


traitsui/qt4/tree_editor.py
```diff
@@ -203,7 +203,7 @@ def _selection_changed(self, selection):
         try:
             tree = self._tree
             if not isinstance(selection, str) and isinstance(
-                selection, collections.Iterable
+                selection, collections.abc.Iterable
```
kitchoi, line 206: There is a (terrible) try-except around this block of code... taking a leap of faith.

traitsui/qt4/list_str_editor.py
```diff
@@ -18,11 +18,11 @@
 """ Traits UI editor for editing lists of strings.
 """
 
+import collections.abc
 
+from pyface.image_resource import ImageResource
```
kitchoi, line 23: Drive-by pep8, collections should be in the standard library block

PR #1138, opened 2020-08-13
Fix Qt date editor problem with None value for date

### LINKED ISSUE #967: Qt simple DateEditor view shows a date despite the underlying value being none
Initially mentioned in #962, to reproduce:

```
from traits.api import Date, HasTraits
from traitsui.api import UItem, View, DateEditor

class Model(HasTraits):

    date = Date()

        return View(
            UItem('date', editor=DateEditor())
        )

model = Model(date=None)

if __name__ == '__main__':
    model.configure_traits()
```

On Mac OSX, the view shows this:
Screenshot 2020-07-06 at 11 08 14

This is misleading as the underlying value is `None`.

aaronayres35:
fixes #967 

See the comments on #967 for more details.

### DIFF
--- a/traitsui/qt4/date_editor.py
+++ b/traitsui/qt4/date_editor.py
@@ -22,6 +22,7 @@
 
 import datetime
 
+from pyface.api import GUI
 from pyface.qt import QtCore, QtGui
 from traits.api import Bool, Date, Dict, Instance, Set
 
@@ -76,6 +77,8 @@ class SimpleEditor(Editor):
             self.control.setMaximumDate(max_date)
 
         self.control.dateChanged.connect(self.update_object)
+        GUI.invoke_later(self.update_editor)
+ 
 
     def dispose(self):
         """ Disposes of the contents of an editor."""
@@ -91,6 +94,8 @@ class SimpleEditor(Editor):
         if value:
             q_date = QtCore.QDate(value.year, value.month, value.day)
             self.control.setDate(q_date)
+        else:  # value is None
+            self.control.lineEdit().setText("")
 
     def update_object(self, q_date):
         """ Handles the user entering input data in the edit control.
--- a/traitsui/tests/editors/test_date_editor.py
+++ b/traitsui/tests/editors/test_date_editor.py
@@ -74,6 +74,27 @@ class TestDateEditorSimpleQt(unittest.TestCase):
             displayed_text = editor.control.lineEdit().displayText()
             self.assertEqual(displayed_text, "")
 
+    def test_simple_editor_date_changed_to_none(self):
+        # Test view when the date is changed from a valid date to None
+        view = View(
+            Item(
+                name="single_date",
+                style="simple",
+            )
+        )
+        model = Foo()
+        model.single_date = datetime.date(2019, 2, 1)
+        with reraise_exceptions(), \
+                create_ui(model, dict(view=view)) as ui:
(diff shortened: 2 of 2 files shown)


kitchoi:
I can reproduce the test error in the PySide2 environment with OSX installed from `etstool.py`.
But if I create another environment and install `PySide2` from PyPI, I get PySide2 5.15.0 instead of PySide2 5.11.0, and the test passes there.

This is how I install another environment with a new PySide2
```
$ python3.8 -m venv ~/my_favourite_folder_for_venvs/pyside2-env
$ source ~/my_favourite_folder_for_venvs/pyside2-env/bin/activate
$ pip install .[pyside2]   # in traitsui source root folder
```

Do you mind confirming if you observe the same behaviour please?

aaronayres35:
> I can reproduce the test error in the PySide2 environment with OSX installed from `etstool.py`.
> But if I create another environment and install `PySide2` from PyPI, I get PySide2 5.15.0 instead of PySide2 5.11.0, and the test passes there.
> 
> This is how I install another environment with a new PySide2
> 
> ```
> $ python3.8 -m venv ~/my_favourite_folder_for_venvs/pyside2-env
> $ source ~/my_favourite_folder_for_venvs/pyside2-env/bin/activate
> $ pip install .[pyside2]   # in traitsui source root folder
> ```
> 
> Do you mind confirming if you observe the same behaviour please?

When I followed your instructions I got that the test passed as well

kitchoi:
Currently the `etstool.py` (used by CI) is pinning PySide2 5.11:
https://github.com/enthought/traitsui/blob/88f4cd8887c36cf5916b8f551594156c705bc420/etstool.py#L204

kitchoi:
Despite the test passes with a newer version of PySide2, I can't verify that this fixes the issue in #967 when I ran the script provided there. @aaronayres35 Have you tried?

kitchoi:
Actually, now I realize that my idea of using `GUI.invoke_later` is a worse workaround than the one in https://forum.qt.io/topic/3068/solved-enhancing-qdateedit-to-allow-for-null-empty-dates/17
The reason is that we could not make sure when the callable provided in `GUI.invoke_later` is called by the event loop, the `QDateEdit` has not been destroyed. For example, the `init` could be called and then the editor be disposed of all without having started the GUI event loop. Then the GUI event loop may start for other reasons, and the callable will raise because it would try to access `self.control` which is no longer valid. This is probably rare (or not a use case) but can happen.

Now I think perhaps we should close this and try something else. @aaronayres35, would that be okay? I am sorry about this.

aaronayres35:
> Actually, now I realize that my idea of using `GUI.invoke_later` is a worse workaround than the one in https://forum.qt.io/topic/3068/solved-enhancing-qdateedit-to-allow-for-null-empty-dates/17
> The reason is that we could not make sure when the callable provided in `GUI.invoke_later` is called by the event loop, the `QDateEdit` has not been destroyed. For example, the `init` could be called and then the editor be disposed of all without having started the GUI event loop. Then the GUI event loop may start for other reasons, and the callable will raise because it would try to access `self.control` which is no longer valid. This is probably rare (or not a use case) but can happen.
> 
> Now I think perhaps we should close this and try something else. @aaronayres35, would that be okay? I am sorry about this.

Yeah that is okay with me, I will close this now.

kitchoi:
Thanks.

Issue #1125, opened 2020-08-10
Better information about missing dependencies in demo application


rahulporuri:
At the moment, there are a number of demo scripts that can be "clicked on" from the ETS Demo application which won't display anything in the "Demo" area. This is because the demo scripts need dependencies which aren't available in a basic ets demo application. At the moment, missing dependencies aren't reported in the application.

For example

- "Adapted tree editor demo" requires apptools,
- "HDF5 tree demo" requires pytables,
- "HDF5 tree demo2" requires h5py etc.

Some of the demo scripts explicitly mention the need for additional packages in the description e.g. "To run this demonstration successfully, you must have the AppTools egg installed." in "Adapted tree editor demo" but the other two in the aforementioned list don't explicitly state this requirement.

1. It would be nice if all of the demo scripts that depend on additional packages explicitly document that dependence in the description.
2. It would be nicer if we can update `demo` section of traitsui ``__extras_requires__`` to list **all** demo dependencies. That way, users can simply do `pip install traitsui[demo]` and won't have to worry about any missing dependencies.

https://github.com/enthought/traitsui/blob/3e3a05a52215e33d88fab882e0f7f27638aa8eb4/traitsui/__init__.py#L30

PR #1140, opened 2020-08-13
Update extras_require to include requirements for examples

kitchoi:
This provides for point 2 of #1125
 This PR updates the `extras_requires` to include dependencies for the demo examples.

We should mention this in the README, but then we should also mention installation options for PyQt5 etc... they aren't there at the moment, and I decided not to mix them in this PR.

### DIFF
--- a/traitsui/__init__.py
+++ b/traitsui/__init__.py
@@ -37,6 +37,7 @@ __extras_require__ = {
         "chaco",   # for a very simple example, see enthought/traitsui#1139
         "h5py",
         "numpy",
+        "pandas",
         "tables",
     ],
     "test": ["packaging"],
--- a/traitsui/__init__.py
+++ b/traitsui/__init__.py
@@ -27,7 +27,14 @@ __extras_require__ = {
     "pyqt": ["pyqt>=4.10", "pygments"],
     "pyqt5": ["pyqt5", "pygments"],
     "pyside2": ["pyside2", "shiboken2", "pygments"],
-    "demo": ["configobj", "docutils"],
+    "examples": [
+        # Dependencies for examples
+        "apptools",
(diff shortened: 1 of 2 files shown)


traitsui/__init__.py
```diff
@@ -27,7 +27,14 @@
     "pyqt": ["pyqt>=4.10", "pygments"],
     "pyqt5": ["pyqt5", "pygments"],
     "pyside2": ["pyside2", "shiboken2", "pygments"],
-    "demo": ["configobj", "docutils"],
+    "examples": [
```
kitchoi, line 30: It was called `demo` because it was the dependencies for the *demo application*.
To be fair, the `traitsui.extras.demo` is still here. But I think we can get rid of that pretty soon now that `etsdemo` is released.

kitchoi: Oops, I thought it was added recently, but it isn't: https://github.com/enthought/enable/commit/c6e7664d8e91180a17eaaf9ce60dfd5218712eb5
I have to add the `demo` back...

traitsui/__init__.py
```diff
@@ -27,7 +27,18 @@
...
+        # to be deprecated, see enthought/traitsui#950
+        "configobj", "docutils",
+    ],
+    "examples": [
+        # Dependencies for examples
+        "apptools",
+        "chaco",   # for a very simple example, see enthought/traitsui#1139
+        "h5py",
+        "numpy",
+        "tables",
```
rahulporuri, line 40: pandas too.

kitchoi: Thanks!

kitchoi:
For the record, `pip install chaco` on my Python3.8 environment is currently failing with some errors with compiling C++ code, but I think it is orthogonal to this: We still need to declare what TraitsUI's examples depend on.

PR #1142, opened 2020-08-13
adding to docstring to specify Popup_Dialog_demo not implemented on qt

### LINKED ISSUE #1068: NotImplementedError when running Popup_Dialog_demo.py on qt
https://github.com/enthought/traitsui/blob/master/traitsui/examples/demo/Advanced/Popup_Dialog_demo.py

When changing the gender of the person from 'Male to 'Female' the following error is output to the command line 

<details>

```
$ python Popup_Dialog_demo.py 
Exception occurred in traits notification handler for object: <__main__.Person object at 0x7ff71d7208e0>, trait: gender, old value: Male, new value: Female
  File "/Users/aayres/.edm/envs/qt-env/lib/python3.6/site-packages/traits/trait_notifiers.py", line 522, in _dispatch_change_event
    self.dispatch(handler, *args)
  File "/Users/aayres/.edm/envs/qt-env/lib/python3.6/site-packages/traits/trait_notifiers.py", line 616, in dispatch
    handler(*args)
  File "/Users/aayres/Desktop/traitsui/traitsui/ui.py", line 909, in dispatch
    self.method(self.info)
  File "Popup_Dialog_demo.py", line 62, in object_gender_changed
    self._ui = self.edit_traits(parent=info.gender.control)
  File "/Users/aayres/Desktop/traitsui/traitsui/handler.py", line 454, in edit_traits
    args,
  File "/Users/aayres/Desktop/traitsui/traitsui/view.py", line 462, in ui
    ui.ui(parent, kind)
  File "/Users/aayres/Desktop/traitsui/traitsui/ui.py", line 246, in ui
    self.rebuild(self, parent)
  File "/Users/aayres/Desktop/traitsui/traitsui/toolkit.py", line 174, in ui_popup
    not_implemented_message.format(ETSConfig.toolkit)
NotImplementedError: the 'qt4' toolkit does not implement this method
```

</details>

It seems as this may be the expected behavior but if so the doctoring should mention that it currently only works with wx

aaronayres35:
Fixes #1068 

There is no ui_popup function in https://github.com/enthought/traitsui/blob/master/traitsui/qt4/toolkit.py, but there is in https://github.com/enthought/traitsui/blob/master/traitsui/wx/toolkit.py.

If I am mistaken and this is actually supposed to work on Qt, then the issue can be directly addressed.

### DIFF
--- a/traitsui/examples/demo/Advanced/Popup_Dialog_demo.py
+++ b/traitsui/examples/demo/Advanced/Popup_Dialog_demo.py
@@ -2,8 +2,6 @@
 #  License: BSD Style.
 
 """
-Note: This demo only works on the wx backend.
-
 Demonstrates using a popup view within another view.
 
 Try changing the gender of the person from 'Male' to 'Female' using the
@@ -24,7 +22,7 @@ The main items of interest in this demo are:
   value specifies the control that the popup dialog should be positioned near.
 
 Notes:
-
+- This demo only works on the wx backend.
 - Traits UI will automatically position the popup dialog near the specified
   control in such a way that the pop-up dialog will not overlay the control
   and will be entirely on the screen (as long as these two conditions do not
--- a/traitsui/examples/demo/Advanced/Popup_Dialog_demo.py
+++ b/traitsui/examples/demo/Advanced/Popup_Dialog_demo.py
@@ -2,6 +2,8 @@
 #  License: BSD Style.
 
 """
+Note: This demo only works on the wx backend.
+
 Demonstrates using a popup view within another view.
 
 Try changing the gender of the person from 'Male' to 'Female' using the


traitsui/examples/demo/Advanced/Popup_Dialog_demo.py
```diff
@@ -2,6 +2,8 @@
 #  License: BSD Style.
 
 """
+Note: This demo only works on the wx backend.
```
kitchoi, line 5: There is a `Notes` section at the bottom. Should this go there or do you think this is important enough and therefore needs to go at the top?

aaronayres35: I think its fine in the bottom section, I will move it.

PR #1143, opened 2020-08-13
adding a workaround that prevents a possible segfault in Settable_cached_property.py

### LINKED ISSUE #1069: Settable_cached_property.py Seg fault
the docstring instructs a user to "Now try typing a value into the 'd' trait and you will see that an error
results (indicated by the text entry field turning red), because this is a
normal 'cached' trait that has no 'setter' method defined."

When I do this on the lower text field for D and try to change the text field becomes red as expected.  However, on the text field on the right hand side if I change D and hit enter, 2 pop up windows appear saying "The 'd' trait of a SettableCachedProperty instance is 'read only'."  Then, when I close those windows the GUi closes and I see 

```
$ python Settable_cached_property.py 
Segmentation fault: 11
```
In the command line.

aaronayres35:
fixes #1069 

This changes the view of the d attribute to not show the slider.  However, the d attribute is intended to be read only and a user can still follow the instructions given in the doctoring to see the text field turn red.  

This may not be the desired solution though, because we are explicitly making it a readonly view, but we want to show that because it is a cached trait with no setter method, it can not be modified.  I just did not know of another way to remove the text box on the right side, which gave the opportunity for a segfault.

It is possible the original behavior isn't even really a big problem and that this change is unnecessary in which case I will just scrap this PR

### DIFF
--- a/traitsui/examples/demo/Advanced/Settable_cached_property.py
+++ b/traitsui/examples/demo/Advanced/Settable_cached_property.py
@@ -84,7 +84,7 @@ class SettableCachedProperty(HasTraits):
         Item('c'),
         '_',
         Item('d',
-             editor=RangeEditor(low=1, high=400, mode='slider')),
+             editor=RangeEditor(low=1, high=400, mode='slider'), style='readonly'),
         Item('d'),
         width=0.3
     )


kitchoi:
Neither do I like this workaround.
Whatever we do with this example, I think we should have an issue for the RangeEditor. Fixing that would not only fix the example but also benefits others who want to use the pattern in production applications.

aaronayres35:
Closing as this is not the right solution.  See comment above.

Issue #509, opened 2019-03-01
Pyside2 5.12 CI failures


corranwebster:
Using Pyside 2 we the test suite will not run, erroring out almost immediately.  I suspect some sort of issue with the PyPI wheel compared to the Python version we are using.

kitchoi:
I don't know if this was referring to the error since in the first commit of #1146:
```
[EXECUTING] edm run -e traitsui-test-3.6-pyside2 -- python -W default -m coverage run -p -m unittest discover -v traitsui
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl, xcb.
Command '['edm', 'run', '-e', 'traitsui-test-3.6-pyside2', '--', 'python', '-W', 'default', '-m', 'coverage', 'run', '-p', '-m', 'unittest', 'discover', '-v', 'traitsui']' returned non-zero exit status 250.
```

If so, then this is a build issue with CI setup, as opposed to TraitsUI not supporting PySide2 5.12 or tests fail with PySide2 5.12 (which I can't find the evidence of), and #1146 will close this issue.

PR #1146, opened 2020-08-14
Relax constrain on PySide2 version

kitchoi:
Looks like PySide2 version was pinned to workaround #509
This was a while ago. Now we have PySide2 5.15.0 from PyPI, let's see if it is good now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,18 @@ services:
 addons:
   apt:
     packages:
-    - libglu1-mesa-dev
+    # Qt dependencies
+    - libxkbcommon-x11-0
+    - libxcb-icccm4
+    - libxcb-image0
+    - libxcb-keysyms1
+    - libxcb-randr0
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -35,8 +47,6 @@ cache:
 before_install:
   - mkdir -p "${HOME}/.cache/download"
   - if [[ ${TRAVIS_OS_NAME} == 'linux' ]]; then ./install-edm-linux.sh; export PATH="${HOME}/edm/bin:${PATH}"; fi
-  # libdbus and libxkb for Pyside2, libsdl for wxpython
-  - if [[ ${TRAVIS_OS_NAME} == 'linux' ]]; then sudo apt-get install -y libdbus-1-3 libxkbcommon-x11-0 libsdl1.2debian; fi
```
rahulporuri, line 39: I guess we don't need `libdbus-1-3` because CI is happy without it.

PR #1147, opened 2020-08-14
Mention additional dependencies in examples

### LINKED ISSUE #1125: Better information about missing dependencies in demo application

kitchoi:
Closes #1125

This PR highlights some additional dependencies in the examples.

Note that for examples that are primarily for editing NumPy array (e.g. `NumPy_array_view_editor_demo.py`), I did not add the mention of "this example requires NumPy" because this feels dumb :)

### DIFF
--- a/traitsui/examples/demo/Standard_Editors/DataFrameEditor_demo.py
+++ b/traitsui/examples/demo/Standard_Editors/DataFrameEditor_demo.py
@@ -11,6 +11,8 @@
   missing.
 
 -------------------------------------------------------------------------------
+
+To run this demonstration successfully, you must have **pandas** installed.
 """
 # Issue related to the demo warning: enthought/traitsui#944
 
--- a/traitsui/examples/demo/Advanced/Adapted_tree_editor_demo.py
+++ b/traitsui/examples/demo/Advanced/Adapted_tree_editor_demo.py
@@ -5,8 +5,8 @@
 Demonstrates an alternative method of defining a **TreeEditor** by creating
 **ITreeNodeAdapter** subclasses.
 
-To run this demonstration successfully, you must have the **AppTools** egg
-installed.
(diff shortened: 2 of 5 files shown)


traitsui/examples/demo/Standard_Editors/DataFrameEditor_demo.py
```diff
@@ -11,6 +11,8 @@
   missing.
 
 -------------------------------------------------------------------------------
+
+To run this demonstration successfully, you must have **pandas** installed.
```
kitchoi, line 15: I almost did not add this sentence, but then DataFrameEditor is not necessarily about "pandas", so many things can be called DataFrame.  This example is missing a better description but that's orthogonal.

Issue #1149, opened 2020-08-18
Extend UITester API


kitchoi:
This issue laid out the few initial enhancements we need for the UITester proposed in #1107.
(At the time of writing, the PR #1107 still needs a bit of refactoring / renaming of things.)

- [x] Define the methods in UITester and UIWrapper with basic implementation. (That is #1107)
- [x] Extend UITester to set a default registry containing implementations for TraitsUI editors.

<details>
    <summary>Details</summary>
Since toolkit specific components may not be importable in an environment that does not have the toolkit, the content of the registry should be different depending on the toolkit being used. Which toolkit is used should be determined at runtime. At this step, the registry object can contain no or just a few implementations. The main goal here is to get the toolkit resolution setup.

</details>

- [x] Add the delay parameter to the UITester and allow it to be propagated to the subsequent UIWrapper.

<details>
    <summary>Details</summary>
This feature allows developers to slow down the test run (down to every single keystroke or mouse click) as they write the test code so that they can verify the test code does what they want. In a typical usage, the delay will be removed after the developer is satisfied with their test code.

The delay parameter should have a default value of no delay. We don't have to use this delay parameter at this step, the parameter will be used in the next step when we add actual implementations performing user interactions.

</details>

- [x] Implement MouseClick for ButtonEditor
<details>
    <summary>Details</summary>

This will be used for testing mouse clicking a button. This editor is chosen because it is very simple but also widely used.

One should then be able to write test code like this:

```
tester = UITester()
with tester.create_ui(obj) as ui:
    button = tester.find_by_name(ui, "some_name")
    with self.assertTraitsChanges(obj, "some_name", count=1):
        button.perform(command.MouseClick())
```

This should use the `MouseClick` command object being introduced in #1107 (at the time of writing, it is still there.)

It should be possible to slow down the test by setting the delay parameter in UITester (added in the previous step).
And the test should be compatible with both Qt and Wx. Convert TraitsUI's own test for ButtonEditor.

</details>

- [x] Implement KeySequence, KeyClick and DisplayedText for TextEditor

<details>
    <summary>Details</summary>
      Similar to MouseClick, key clicking is a very basic operation for testing. By "key clicking", this means pressing and then immediately releasing a key.

Note that "key clicking" is different from just "key press". Key press does not include the release, aka it could be holding a key down.

In addition to the users causing changes via mouse clicks and key clicks, tests also need to inspect content being displayed on the GUI. This step also introduces a query: DisplayedText.

[...]

PR #1150, opened 2020-08-18
Draft implementation for various UITester features

kitchoi:
This is a draft PR which I will close immediately. This contains proof-of-concept code which I used to experiment with ways to implement the UITester. I also tested the API with a number of editors: ButtonEditor, TextEditor, ListEditor, EnumEditor, TableEditor, TreeEditor, RangeEditor as well as the UI panel specific TabbedFoldGroupEditor.

Qt/Wx toolkit compatibility was tested for ButtonEditor and RangeEditor only.

It is open for the reference of #1149.

Caution:
- Various classes and variables may be renamed in the future. 
- Since I was writing this mainly for testing a concept / demo, there are quite a few undocumented functions... please feel free to ask about them. And the structures may not be the most optimal.
- Where there are documentation, they might also have been outdated.
- It does contain more than what we need for #1149.


traitsui/testing/ui_tester.py
```diff
@@ -0,0 +1,422 @@
...
+    A ``UIWrapper`` finds toolkit specific implementations for simulating
+    user interactions from one or many ``InteractionRegistry`` objects. These
+    registries are typically provided via another public interface, e.g.
+    ``UITester``.
+
+    Attributes
+    ----------
+    editor : Editor
+        An instance of Editor. It is assumed to be in a state after the UI
+        has been initialized but before it is disposed of.
```
kitchoi, line 244: The `editor` attribute will be renamed, and it may not be an instance of traitsui Editor.
#1107 will clarify that.

traitsui/testing/interactor_registry.py
```diff
@@ -0,0 +1,89 @@
...
+from traitsui.testing.exceptions import (
+    ActionNotSupported, LocationNotSupported,
+)
+
+
+class InteractionRegistry:
+
+    def __init__(self):
+        self.editor_to_action_to_handler = {}
+        self.editor_to_location_solver = {}
```
kitchoi, line 12: These two are independent, so we might as well have a registry for interaction, and a separate registry for locations.

Issue #768, opened 2020-03-30
Python 2 clean-up


corranwebster:
Remove some remaining code for Python 2 support:

- [x] `Editor._str` is no longer needed.
- [ ] `capitalize = lambda s: s.capitalize()` removal
- [x] remove `u'...'` strings

Others will be added as discovered.

PR #1151, opened 2020-08-18
Cleanup some Python 2 compat code

rahulporuri:
addresses part of #768


traitsui/editor.py
```diff
@@ -691,26 +691,6 @@ def __set_value(self, value):
                 self.error(excp)
                 raise
 
-    def _str(self, value):
```
kitchoi, line 694: It does look like no one is calling this method.

kitchoi (approved):
LGTM

PR #1152, opened 2020-08-19
first round of updates for UI Tester (default_registry toolkit setup and added delay)

aaronayres35:
Addresses first two bullets on issue #1149

### DIFF
--- a/traitsui/testing/tester/default_registry.py
+++ b/traitsui/testing/tester/default_registry.py
@@ -1,0 +1,22 @@
+#  Copyright (c) 2005-2020, Enthought, Inc.
+#  All rights reserved.
+#
+#  This software is provided without warranty under the terms of the BSD
+#  license included in LICENSE.txt and may be redistributed only
+#  under the conditions described in the aforementioned license.  The license
+#  is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+#  Thanks for using Enthought open source!
+#
+
+import importlib
+
+from traits.etsconfig.api import ETSConfig
+
+from traitsui.testing.tester.registry import TargetRegistry
+
+
+def get_default_registry():
+    # side-effect to determine current toolkit
+    module = importlib.import_module(".default_registry", "traitsui.testing.tester." + ETSConfig.toolkit)
+    return module.get_default_registry()
--- a/traitsui/testing/tester/tests/test_default_registry.py
+++ b/traitsui/testing/tester/tests/test_default_registry.py
@@ -1,0 +1,23 @@
+#  Copyright (c) 2005-2020, Enthought, Inc.
+#  All rights reserved.
+#
+#  This software is provided without warranty under the terms of the BSD
+#  license included in LICENSE.txt and may be redistributed only
+#  under the conditions described in the aforementioned license.  The license
+#  is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+#  Thanks for using Enthought open source!
+#
+
+import unittest
+
+from traitsui.testing.tester.default_registry import get_default_registry
+from traitsui.testing.tester.registry import TargetRegistry
+
+
+class TestDefaultRegistry(unittest.TestCase):
+
+    def test_load_default_registries(self):
+        registry = get_default_registry()
+        self.assertIsInstance(registry, TargetRegistry)
+        
--- a/traitsui/testing/tester/ui_tester.py
+++ b/traitsui/testing/tester/ui_tester.py
@@ -11,6 +11,7 @@
 
 from traitsui.ui import UI
 from traitsui.testing.tester import locator
+from traitsui.testing.tester.default_registry import get_default_registry
 from traitsui.testing.tester.registry import TargetRegistry
 from traitsui.testing.tester.ui_wrapper import UIWrapper
 from traitsui.tests._tools import (
@@ -131,7 +132,7 @@ class UITester:
         of decreasing priority. A shallow copy will be made.
     """
 
-    def __init__(self, registries=None):
+    def __init__(self, registries=None, delay=0):
         """ Instantiate the UI tester.
         """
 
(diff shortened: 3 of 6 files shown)


traitsui/testing/tester/default_registry.py
```diff
@@ -0,0 +1,26 @@
...
+from traitsui.tests._tools import is_null
+
+
+def get_default_registry():
+    # side-effect to determine current toolkit
+    if is_null():
+        return TargetRegistry()
+    else:
+        module = importlib.import_module(".default_registry", "traitsui.testing.tester." + ETSConfig.toolkit)
+        return module.get_default_registry()
```
aaronayres35, line 26: I wasn't sure if the Null toolkit warranted having a separate file.  It seemed to me that it wouldn't and an empty default registry would serve the needed purpose.

traitsui/testing/tester/default_registry.py
```diff
@@ -0,0 +1,26 @@
...
+
+import importlib
+
+from traits.etsconfig.api import ETSConfig
+
+from traitsui.testing.tester.registry import TargetRegistry
+from traitsui.tests._tools import is_null
+
+
+def get_default_registry():
```
kitchoi, line 20: Could we have a docstring describing what it is please? (same for the other functions)
(I did not have very many docstring for the draft branch either but I have an excuse: It was draft 😂 )

traitsui/testing/tester/ui_wrapper.py
```diff
@@ -80,6 +81,7 @@ def locate(self, location):
         return UIWrapper(
             target=self._get_next_target(location),
             registries=self._registries,
+            delay=self.delay,
```
kitchoi, line 84: I think we need a test to check that the `delay` parameter is persisted throughout after calling `UITester.find_by_name`.

traitsui/testing/tester/tests/test_default_registry.py
```diff
@@ -0,0 +1,23 @@
...
+from traitsui.testing.tester.default_registry import get_default_registry
+from traitsui.testing.tester.registry import TargetRegistry
+
+
+class TestDefaultRegistry(unittest.TestCase):
+
+    def test_load_default_registries(self):
+        registry = get_default_registry()
+        self.assertIsInstance(registry, TargetRegistry)
+
```
kitchoi, line 23: Nit: I don't know why GitHub is highlighting this lack of newline at the end of file.

What do you think about setting up flake8 on CI just to lint this subpackage, would it help? (Separate PR, of course)

kitchoi (commented):
Thank you!

traitsui/testing/tester/tests/test_default_registry.py
aaronayres35: I agree that would be a good idea.  I can try to do that soon

traitsui/testing/tester/default_registry.py
```diff
@@ -0,0 +1,34 @@
...
+        that is toolkit specific.  
+    """
+    # side-effect to determine current toolkit
+    from pyface.toolkit import toolkit_object
+    from traits.etsconfig.api import ETSConfig
+
+    if ETSConfig.toolkit == "null":
+        return TargetRegistry()
+    else:
+        toolkit = {'wx': 'wx', 'qt4': 'qt4', 'qt': 'qt4'}[ETSConfig.toolkit]
```
aaronayres35, line 32: There may be a better way to map qt to qt4

traitsui/testing/tester/ui_tester.py
```diff
@@ -131,7 +132,7 @@ class App(HasTraits):
         of decreasing priority. A shallow copy will be made.
     """
 
-    def __init__(self, registries=None):
+    def __init__(self, registries=None, delay=0):
```
kitchoi, line 135: Oops, I think I failed to push a comment :(
Could we expand the description in the class docstring what this `delay` is about please?
In particular, what unit did you choose?

I realized that the "Attributes" section should really be named "Parameters" and there needs to be a separate "Attributes" section to conform to NumPy docstring convention. The `UIWrapper`'s `Parameters` needs to be moved up into the class docstring too. I can do the moves in a separate PR though. If you could just add the description for the `delay` next to `registries`, that'd be great.

traitsui/testing/tester/default_registry.py
kitchoi: Looks fine to me. This is logically the same as defining an entry point to map "qt" to the "qt4" package like how it is done in traitsui and pyface setup.py.

traitsui/testing/tester/default_registry.py
```diff
@@ -0,0 +1,34 @@
...
+
+    Returns
+    -------
+    registry : TargetRegistry
+        The default registry containing implementations for TraitsUI editors
+        that is toolkit specific.  
+    """
+    # side-effect to determine current toolkit
+    from pyface.toolkit import toolkit_object
+    from traits.etsconfig.api import ETSConfig
```
kitchoi, line 27: Nit: The `ETSConfig` import does not have the toolkit resolution side-effect, though. So the second import can be taken out. It would be clearer that the side-effect comment is only referring to the pyface import.

traitsui/testing/tester/default_registry.py
```diff
@@ -0,0 +1,34 @@
...
+#  under the conditions described in the aforementioned license.  The license
+#  is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+#  Thanks for using Enthought open source!
+#
+
+import importlib
+
+from traitsui.testing.tester.registry import TargetRegistry
+
```
kitchoi, line 15: 🤖 Flake8: We need two blank lines before a new function definition.

kitchoi:
For reference, the CI job was rebuilt on the last commit after #1155 was merged, hence receiving the workaround.

PR #1157, opened 2020-08-21
Additional UITester updates - adding MouseClick for ButtonEditor and tests

aaronayres35:
Implements the fourth bullet on issue #1149 
(commit history got wrapped up with the last PR, but the files changed seem to only be the new changes).

In this PR, a few extra things needed to be added.  For example, the Default Target locator class and the corresponding logic in UIWrapper method _perform_or_inspect.  Additionally, in order to replace the old TraitsUI test for the button editor, I also needed to add the query.DisplayedText handlers to the default registries.


traitsui/testing/tester/wx/default_registry.py
```diff
@@ -21,5 +28,31 @@ def get_default_registry():
         The default registry containing implementations for TraitsUI editors
         that is wx specific.  
     """
+    registry = get_wobject_registry()
+
+    button_editor.register(registry)
+    
+    return registry
+
+def get_wobject_registry():
```
kitchoi, line 37: Same question as above, did you want to move this function to helpers?

BTW, unlike [`QObject`](https://doc.qt.io/qt-5/qobject.html),  `wobject` is not a well-known name. Perhaps we should name this `wx_object`?

traitsui/tests/editors/test_button_editor.py
```diff
@@ -46,36 +49,25 @@ class ButtonTextEdit(HasTraits):
...
+        tester = UITester()
+        with tester.create_ui(button_text_edit, dict(view=view)) as ui:
+            button = tester.find_by_name(ui, "play_button")
+            actual = button.inspect(query.DisplayedText())
+            self.assertEqual(actual, "I'm a play button")
 
             button_text_edit.play_button_label = "New Label"
-            self.assertEqual(get_button_text(button), "New Label")
+            actual = button.inspect(query.DisplayedText())
+            self.assertEqual(actual, "New Label")
```
kitchoi, line 65: The following tests are now failing on wx: 
```
ERROR: test_custom_button_editor (traitsui.tests.editors.test_button_editor.TestButtonEditor)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_button_editor.py", line 77, in test_custom_button_editor
    self.check_button_text_update(custom_view)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_button_editor.py", line 60, in check_button_text_update
    actual = button.inspect(query.DisplayedText())
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\testing\tester\ui_wrapper.py", line 129, in inspect
    return self._perform_or_inspect(interaction)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\testing\tester\ui_wrapper.py", line 185, in _perform_or_inspect
    supported=supported,
traitsui.testing.tester.exceptions.InteractionNotSupported: No handler is found for target <class 'traitsui.wx.button_editor.CustomEditor'> with interaction <class 'traitsui.testing.tester.query.DisplayedText'>. Supported these: []
ERROR: test_custom_button_editor_clicked (traitsui.tests.editors.test_button_editor.TestButtonEditor)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_button_editor.py", line 94, in test_custom_button_editor_clicked
    self.check_button_fired_event(custom_view)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_button_editor.py", line 88, in check_button_fired_event
    button.perform(command.MouseClick())
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\testing\tester\ui_wrapper.py", line 116, in perform
    self._perform_or_inspect(interaction)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\testing\tester\ui_wrapper.py", line 185, in _perform_or_inspect
    supported=supported,
traitsui.testing.tester.exceptions.InteractionNotSupported: No handler is found for target <class 'traitsui.wx.button_editor.CustomEditor'> with interaction <class 'traitsui.testing.tester.command.MouseClick'>. Supported these: []
ERROR: test_simple_button_editor (traitsui.tests.editors.test_button_editor.TestButtonEditor)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_button_editor.py", line 74, in test_simple_button_editor
    self.check_button_text_update(simple_view)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_button_editor.py", line 60, in check_button_text_update
    actual = button.inspect(query.DisplayedText())
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\testing\tester\ui_wrapper.py", line 129, in inspect
    return self._perform_or_inspect(interaction)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\testing\tester\ui_wrapper.py", line 185, in _perform_or_inspect
    supported=supported,
traitsui.testing.tester.exceptions.InteractionNotSupported: No handler is found for target <class 'traitsui.wx.button_editor.SimpleEditor'> with interaction <class 'traitsui.testing.tester.query.DisplayedText'>. Supported these: [<class 'traitsui.testing.tester.command.MouseClick'>]
```
I think they are passing before. I can reproduce this locally.

They are failing on Appveyor and Travis too. However the wx build is allowed to fail, because there are still existing issues with wx we have not finished fixing. So I am afraid we cannot just rely on the green tick on the PR, we will need to check the CI job log for new failures and maybe run the test suite locally too.

traitsui/testing/tester/wx/implementation/button_editor.py
```diff
@@ -0,0 +1,28 @@
...
+from traitsui.wx.button_editor import SimpleEditor, CustomEditor
+from traitsui.testing.tester import locator, query
+from traitsui.testing.tester.wx import helpers
+
+
+def register(registry):
+    """ Register solvers/handlers specific to wx Button Editors
+    for the given registry.
+
+    If there are any conflicts, an error will occur.
```
kitchoi, line 20: Nit: Could we expand the docstring to document the type of `registry` please?

traitsui/testing/tester/qt4/tests/test_default_registry.py
```diff
@@ -0,0 +1,67 @@
...
+        self.assertEqual(click_slot.call_count, 1)
+
+    def test_mouse_click_disabled(self):
+        button = QtGui.QPushButton()
+        button.setEnabled(False)
+
+        click_slot = mock.Mock()
+        button.clicked.connect(click_slot)
+
+        wrapper = wrapper = UIWrapper(
```
kitchoi: `wrapper` is defined twice, we can remove one of them. Same for the test above.

traitsui/testing/tester/ui_wrapper.py
```diff
@@ -161,6 +161,24 @@ def _perform_or_inspect(self, interaction):
                 with _event_processed():
                     return handler(self, interaction)
 
+        try:
+            # there may not be a handler registered for this interaction on
+            # the original target.  Instead, it may be registered for DefaultTarget
```
kitchoi, line 166: Thank you for adding these explanations. However having so many comments beg a different question: Is the code block difficult to understand? If so, we might be better off extracting this nested-try-block into a separate (private) method so that we can better document it.

traitsui/testing/tester/locator.py
```diff
@@ -35,3 +35,17 @@ class TargetByName:
     """
     def __init__(self, name):
         self.name = name
+
+class DefaultTarget:
+    """ A locator for locating a default target.  In many cases, handlers may
```
kitchoi, line 40: Nit:  "many" maybe an overstatement depending on future additions to the registry... maybe "some"?

traitsui/testing/tester/locator.py
```diff
@@ -35,3 +35,17 @@ class TargetByName:
...
+    not be registered for certain interactions on a given target.  Instead
+    there may be a solver registered a locator_class of DefaultTarget and a
+    target_class of the original target.  The registered solver will then
+    return some generic (toolkit specific) object, for which frequently used
+    handlers will be registered.  If user tries to perform the intetractions
+    of these handlers on the original object, machinery is in place for the
+    UIWrapper to try and locate a DefaultTarget and then look for the needed
+    handler automatically, if it can not find a handler directly for the
+    original target and interaction.
```
kitchoi, line 49: The above is good for developers whose tasks are implementing solvers / handlers. Thank you!  For users of the tester, this might be a bit heavy going. Most of the time, the users would not be using this object directly (it is used under the hood). Maybe we should mention this upfront. This object is going to be part of the public facing API for people _using_ the tester.

traitsui/testing/tester/qt4/default_registry.py
```diff
@@ -21,5 +27,44 @@ def get_default_registry():
...
     """
+    registry = get_qobject_registry()
+
+    button_editor.register(registry)
+
+
+    return registry
+
+
+def get_qobject_registry():
```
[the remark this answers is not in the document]
aaronayres35: I ended up leaving it in the default_registry file, because I thought it might be useful if someone was trying to look into what handlers/solvers are included in the default registry to be able to see them all in one file (at least for the most part aside from specific editor implementations).  The specific handlers (e.g. `mouse_slick_qwidget` are kept in helpers.py).  After thinking again though, maybe it should be in handlers, because realistically a user should never need to think about `QPushButtons` etc., they would only want to know which editors are supported and then they could assume they could do whatever is needed with those editors.  

So, I think I will move it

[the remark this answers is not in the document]
aaronayres35: Actually, I may move it after my next PR to avoid conflicts (I have been working on text editor implementation where this function is in the default registry file and don't want to deal with transferring it over and then having to merge).  But I think I will move it eventually.

traitsui/testing/tester/wx/helpers.py
```diff
@@ -0,0 +1,62 @@
...
+    wx.MilliSleep(wrapper.delay)
+
+    left_down_event = wx.MouseEvent(
+        wx.wxEVT_LEFT_DOWN
+    )
+    left_up_event = wx.MouseEvent(
+        wx.wxEVT_LEFT_UP
+    )
+    control.ProcessEvent(left_down_event)
+    control.ProcessEvent(left_up_event)
```
aaronayres35, line 62: This works as intended, but Im not sure if this is the best way to emulate a Custom Style Button being clicked.  When using Custom Editor, the CustomEditor creates an `_control` attribute that is an ImageButton instance, and then the control of that is assigned as the editors control.  That control is a wx.Window (see the added handlers in wx Default Registry).  Since it is a window, normally, checks are performed to ensure the location of the click is correct 
```wdx, wdy = control.GetClientSize().Get()
        x, y = event.GetX(), event.GetY()
        control.Refresh()
        if (0 <= x < wdx) and (0 <= y < wdy):
            if self.style == "radio":
                self.selected = True
            elif self.style == "checkbox":
                self.selected = not self.selected
            else:
                self.clicked = True
```
(from https://github.com/enthought/pyface/blob/master/pyface/ui/wx/image_button.py)

@kitchoi should I try to update this to specify a location to click?  

Sldo the name specifies ImageButton as that was its intended use, but in reality the control is just a wx.Window, and no specifics to ImageButton are used.  Unsure if I should change the name and leave it only depending on wx.Window, or if I should rewrite the function to be more ImageButton specific.

traitsui/testing/tester/wx/default_registry.py
```diff
@@ -19,7 +26,48 @@ def get_default_registry():
...
+        target_class=wx.Window,
+        interaction_class=command.MouseClick,
+        handler=helpers.mouse_click_ImageButton
+    )
+
+    registry.register_handler(
+        target_class=wx.Window,
+        interaction_class=query.DisplayedText,
+        handler=lambda wrapper, _: wrapper.target.GetLabel()
+    )
```
kitchoi, line 71: `wx.Window` [is extremely broad in its functionality](https://wxpython.org/Phoenix/docs/html/wx.Window.html), so I am not sure it belongs in this registry for reuse purposes, despite it being a wx object.

We discussed the cost and benefit of the approach of DefaultTarget + generic registry:
The cost is a more obscure code path that is difficult to do static code analysis on. The benefit is the ability to reuse interactions already defined on common and simple widgets.

The widgets in this registry should be so simple that it cannot be broken down into smaller pieces, there should be only one place else a mouse click or key click could occur, and expected outcome of these actions should be unambiguous. Simple things like button and textbox are good candidates for that, but a container that can be used for many different use cases is not. If we are good with deciding what interactions should be elevated to be shared status, this would be okay, but if we are not, tracking these mistakes will be hard.

That's the risk factor I feared: It is so tempting to add more complicated components to this registry where they don't belong. I think we prove that this trap is real.

The elaborate comments are initial indications to the cost, this `wx.Window` is the last straw. The combined cost and risk make me think that this DefaultTarget and the `get_wx_object_registry` / `get_qobject_registry` is not worth the benefit of reuse, so let's revert this. Instead of using a DefaultTarget for delegation, we should register the MouseClick and DisplayedText handlers to the (simple/custom) ButtonEditor directly. This will keep the logic and pattern simple to follow - developers will not need to think very hard about this.

It is really good that this happens in such an early stage, thank you very much!  Now we can make this conscious choice with evidence of what could go wrong had we kept this pattern.

Most of the tests in this PR can be kept. The new tests we added for the nested-try-block can go. The tests in `test_default_registry` will probably need to be modified to call the functions from `helpers` instead (and maybe moved to `test_helpers`).

traitsui/testing/tester/default_registry.py
```diff
@@ -23,14 +23,13 @@ def get_default_registry():
     -------
     registry : TargetRegistry
         The default registry containing implementations for TraitsUI editors
-        that is toolkit specific.  
+        that is toolkit specific.
     """
-    # side-effect to determine current toolkit
-    from pyface.toolkit import toolkit_object
```
kitchoi, line 29: I think we still need this here. Let's bring this back and tell flake8 not to complain about this line, see https://flake8.pycqa.org/en/3.1.1/user/ignoring-errors.html#in-line-ignoring-errors

traitsui/testing/tester/ui_wrapper.py
```diff
@@ -161,12 +161,63 @@ def _perform_or_inspect(self, interaction):
...
+            pass
+        else:
+            # if we can locate a DefaultTarget try to perform the
+            # interactation on it instead
+            try:
+                return default_target._perform_or_inspect(interaction)
+            # if we can't add to supported
+            except InteractionNotSupported as e:
+                supported.extend(e.supported)
+        return -1
```
kitchoi, line 218: Returning `-1` for reporting an error is a pattern used in other languages. It has an immediate downside: It means nothing should return `-1` in a normal run condition, so an `inspect(interaction)` that returns `-1` will need to be avoided or worked around somehow. In Python we can raise an exception for it to be captured to avoid this problem.

If this method is to be kept, instead of returning `-1`, we should make sure it raises InteractionNotSupported.
See comment above about reverting the pattern, which means this method will go.

traitsui/testing/tester/wx/helpers.py
kitchoi: > This works as intended, but Im not sure if this is the best way to emulate a Custom Style Button being clicked.

This looks like a decent solution to me.

> should I try to update this to specify a location to click?

I did not notice that wx custom button editor has these many options.
Let's keep this simple and limit ourselves to testing the `style="button"` behaviour for now.

> Sldo the name specifies ImageButton as that was its intended use, but in reality the control is just a wx.Window, and no specifics to ImageButton are used. Unsure if I should change the name and leave it only depending on wx.Window, or if I should rewrite the function to be more ImageButton specific.

I am with you on this dilemma. I would guess that this function will evolve to become more ImageButton specific. Perhaps we should move this function to the module specific to wx ButtonEditor implementations and keep the name to be specific to ImageButton. Then we can generalize later when it becomes clearer this function can be generalized. Given this function will be called directly by name, it will be easy to track this down and move this out to a more shareable module later.

traitsui/testing/tester/wx/default_registry.py
aaronayres35: I realize that for the current use case, I believe these handlers can easily be modified to use ImageButton as the target class instead of wx.Window.  After I do that I will move them into the Button Editor implementation rather than the get_wx_object_registry.

aaronayres35: I can still undo the Default Target delegation as well.

traitsui/testing/tester/ui_wrapper.py
aaronayres35: Yeah I was worried about this.  The problem I was having is that some of the handlers were returning None (like performing an inspection) and I wanted to check, did this method return a call to a handler, or did it reach completion in which case no handler or solver was found so it adds to supported.  I was struggling to figure out how to separate one form the other.  Just reading within the method makes more sense, but regardless it seems this is likely to get scrapped anyway.

traitsui/testing/tester/wx/default_registry.py
kitchoi: Thank you. Let's undo the Default Target delegation if that's okay for you.

aaronayres35:
CI will fail, I still need to update test_default_registry files.

traitsui/testing/tester/qt4/helpers.py
```diff
@@ -0,0 +1,33 @@
...
+    """ Performs a mouce click on a Qt widget.
+
+    Parameters
+    ----------
+    wrapper : UIWrapper
+        The wrapper object wrapping the Qt widget.
+    interaction : instance of traitsui.testing.tester.command.MouseClick
+        interaction is unused here, but it is included so that the function
+        matches the expected format for a handler.  Note this handler is
+        intended to be used with an interaction_class of a MouseClick.
```
aaronayres35, line 27: I need to update docstrings

kitchoi (approved):
LGTM. Thank you!

Issue #1073, opened 2020-07-24
CompoundEditor_demo.py segfault on qt


aaronayres35:
It is possible this is the expected behavior but I was confused.

When using Qt, if you click for example 'C' in the custom section, and then in the text section try to input an integer error messages pop up and when you close them it seg faults.  
On wx however, you simply get the pop ups giving a value error, but the GUI doesn't close and there is no segfault. Once you manual close it, there is an error in the command line namely 
```
$ python CompoundEditor_demo.py 
03:35:06 PM: Debug: Adding duplicate image handler for 'Windows bitmap file'
03:35:06 PM: Debug: Adding duplicate animation handler for '1' type
03:35:06 PM: Debug: Adding duplicate animation handler for '2' type
03:35:06 PM: Debug: Adding duplicate image handler for 'Windows bitmap file'
03:35:06 PM: Debug: Adding duplicate animation handler for '1' type
03:35:06 PM: Debug: Adding duplicate animation handler for '2' type
03:35:06 PM: Debug: Adding duplicate image handler for 'Windows bitmap file'
03:35:06 PM: Debug: Adding duplicate animation handler for '1' type
03:35:06 PM: Debug: Adding duplicate animation handler for '2' type
Traceback (most recent call last):
  File "/Users/aayres/Desktop/traitsui/traitsui/wx/editor_factory.py", line 125, in update_object
    self.value = self.control.GetValue()
AttributeError: 'NoneType' object has no attribute 'GetValue'
```

kitchoi:
> On wx however, you simply get the pop ups giving a value error, but the GUI doesn't close and there is no segfault. Once you manual close it, there is an error in the command line namely

If I added some print functions in `traitsui.wx.editor_factory.TextEditor.update_object`:
```

    def update_object(self, event):
        """ Handles the user changing the contents of the edit control.
        """
        print("entering", event)
        if isinstance(event, wx.FocusEvent):
            event.Skip()
        try:
            self.value = self.control.GetValue()
        except TraitError as excp:
            pass
        print("exiting", event)
```
When I type an invalid value in the textbox, then I see this in the console:
```
entering <wx._core.CommandEvent object at 0x7fee9cf41948>
entering <wx._core.FocusEvent object at 0x7fee9cf4faf8>
```
When I then press OK on the dialog showing the validation error, a second dialog popups, and at that time, this turns up in the console:
```
exiting <wx._core.FocusEvent object at 0x7fee9cf4faf8>
```
[...]

PR #1161, opened 2020-08-25
Implement dispose for wx simple TextEditor (base class)

### LINKED ISSUE #942: wx text style FontEditor needs a dispose method
After editing a text style `FontEditor` and closing the window (using demo `Standard_Editors/FontEditor_demo.py`), an error is thrown:
```
  File "/Users/icernyte/Documents/repos/traitsui/traitsui/wx/font_editor.py", line 291, in update_object
    self.value = self.control.GetValue()
AttributeError: 'NoneType' object has no attribute 'GetValue'
```

This happens because of a `FocusEvent` that is processed after `control` is set to `None`.

The event handler is bound in the general `TextEditor` that `FontEditor` subclasses, so the `dispose` method would actually have to be added there. I opened the issue for `FontEditor` because it's the only editor that throws an error like this. Not sure why others aren't running into the same issue.

### LINKED ISSUE #1145: UI disposal may cause AttributeError on None

>

kitchoi:
Closes #942 
Also resolve the second item in https://github.com/enthought/traitsui/issues/1073
Also part of #752 

The cause of the issue is as explained in #942.  This base class is used by both the text style FontEditor and the text style CompoundEditor.

I came across this while I was trying to _resolve_ #1145.
(The solution I am going to propose for #1145 also "resolves the error" in the context when the UI wrapping this editor has its `dispose` method called. But I believe removing the handler hooks with Unbind is still the right thing to do.)

As a drive-by improvement to the test coverage, implementations are added for testing the Qt implementation of the same editor.

### DIFF
--- a/traitsui/tests/editors/test_font_editor.py
+++ b/traitsui/tests/editors/test_font_editor.py
@@ -15,29 +15,15 @@ import unittest
 from traits.api import HasTraits
 from traitsui.api import Font, Item, View
 from traitsui.tests._tools import (
-    is_wx,
     requires_toolkit,
     ToolkitName,
 )
 from traitsui.testing.tester import command
 from traitsui.testing.tester.ui_tester import UITester
-from traitsui.testing.tester.registry import TargetRegistry
 
 
 class ObjectWithFont(HasTraits):
     font_trait = Font()
-
-
-# A local registry of interaction handlers for testing purposes.
-LOCAL_TARGET_REGISTRY = TargetRegistry()
-
-if is_wx():
-    from traitsui.wx.font_editor import TextEditor
-    LOCAL_TARGET_REGISTRY.register_handler(
-        target_class=TextEditor,
-        interaction_class=command.MouseClick,
-        handler=lambda wrapper, _: wrapper.target.control.SetFocus()
-    )
 
 
 @requires_toolkit([ToolkitName.wx])
@@ -47,7 +33,7 @@ class TestFontEditor(unittest.TestCase):
         # Setting focus on the widget and then disposing the widget
         # should not cause errors.
         view = View(Item("font_trait", style="text"))
-        tester = UITester([LOCAL_TARGET_REGISTRY])
+        tester = UITester()
         with tester.create_ui(ObjectWithFont(), dict(view=view)) as ui:
             wrapper = tester.find_by_name(ui, "font_trait")
             wrapper.perform(command.MouseClick())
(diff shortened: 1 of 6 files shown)


traitsui/tests/editors/test_font_editor.py
```diff
@@ -0,0 +1,53 @@
...
+class TestFontEditor(unittest.TestCase):
+
+    def test_create_and_dispose_text_style(self):
+        # Setting focus on the widget and then disposing the widget
+        # should not cause errors.
+        view = View(Item("font_trait", style="text"))
+        tester = UITester([LOCAL_TARGET_REGISTRY])
+        with tester.create_ui(ObjectWithFont(), dict(view=view)) as ui:
+            wrapper = tester.find_by_name(ui, "font_trait")
+            wrapper.perform(command.MouseClick())
```
kitchoi, line 53: For the record, I was able to get the attribute error shown in #942, but for some reasons the event is processed so late that the error occurrs after the context manager exits. Hence it is not captured for causing the test to fail. The error was printed to the console however. I did not try to dive any further into understanding why the exception could not be captured. The fix resolves issue such that the error is not printed to the console.

traitsui/tests/editors/test_font_editor.py
```diff
@@ -0,0 +1,53 @@
...
+# A local registry of interaction handlers for testing purposes.
+LOCAL_TARGET_REGISTRY = TargetRegistry()
+
+if is_wx():
+    from traitsui.wx.font_editor import TextEditor
+    LOCAL_TARGET_REGISTRY.register_handler(
+        target_class=TextEditor,
+        interaction_class=command.MouseClick,
+        handler=lambda wrapper, _: wrapper.target.control.SetFocus()
+    )
```
kitchoi, line 40: This demonstrates how we can write tests for corner cases without necessarily exposing those test logic to the tester public API.
However this particular interaction might be worth bumping to the public API anyway.
After #1190, I will be able to extend support for the text-style font editor more trivially.

kitchoi:
Updated to test the font editor for Qt as well. It is strictly speaking orthogonal to fixing #942, but is a good opportunity to expand test coverage and UITester support.

traitsui/tests/editors/test_font_editor.py
aaronayres35: FWIW I observed the same

traitsui/wx/editor_factory.py
```diff
@@ -116,6 +116,17 @@ def init(self, parent):
         parent.Bind(wx.EVT_TEXT_ENTER, self.update_object, id=self.control.GetId())
         self.set_tooltip()
 
+    def dispose(self):
```
aaronayres35, line 119: could we add a docstring here?

aaronayres35 (approved):
LGTM! just one minor comment for adding a docstring

PR #1162, opened 2020-08-25
Ui tester api updates3 - Implement KeySequence, KeyClick and DisplayedText for TextEditor

aaronayres35:
Implements the fifth bullet on issue #1149


traitsui/testing/tester/wx/helpers.py
```diff
@@ -11,6 +11,63 @@
...
+                key_event.SetUnicodeKey(KEY)
+                widget.EmulateKeyPress(key_event)
+
+    else:
+        wx.MilliSleep(delay)
+        KEY = mapping[key]
+        key_event = wx.KeyEvent(wx.wxEVT_CHAR)
+        apply_modifiers(key_event, modifiers)
+        key_event.SetKeyCode(mapping[key])
+        widget.EmulateKeyPress(key_event)
```
aaronayres35, line 69: I attempted to add functionality for handling modifiers.  However, this code is currently not working.  It may be due to drawbacks with the EmulateKeyPress method (https://discuss.wxpython.org/t/wx-textctrl-and-emulatekeypress/28300/3), it may not...
Currently, the code allows for simulated key clicks/sequences for standard characters, and pyface key names (e.g. "Backspace").
However, trying to perform key clicks with things like "Shift-A", or "Ctrl-C" currently do not work.

traitsui/testing/tester/wx/helpers.py
```diff
@@ -29,3 +91,73 @@ def mouse_click_button(control, delay):
...
+        The wx Object to be acted on. Should be wx.TextCtrl instance. If it is
+        a wx.StaticText (i.e. from a ReadonlyEditor) nothing occurs.
+    interaction : instance of command.KeyClick
+        The interaction object holding the key input
+        to be simulated being typed
+    delay : int
+        Time delay (in ms) in which the key click will be performed.
+    """
+    if isinstance(control, wx.StaticText):
+        return
```
kitchoi, line 132: See another comment: It is questionable to support key click and key sequence for StaticText.  I think we can just remove this for now. Same for key_sequence_text_ctrl.

traitsui/testing/tester/wx/helpers.py
```diff
@@ -29,3 +91,73 @@ def mouse_click_button(control, delay):
...
+    control.ProcessEvent(click_event)
+
+
+def key_click_text_ctrl(control, interaction, delay):
+    """ Performs simulated typing of a key on the given wxObject
+    after a delay.
+
+    Parameters
+    ----------
+    control : wxObject
```
kitchoi, line 122: I think we only support `wx.TextCtrl` here, the `EmulateKeyPress` is not generally available I think. So let's be specific about it.

traitsui/testing/tester/wx/helpers.py
```diff
@@ -11,16 +11,78 @@
...
+    key_event.SetShiftDown(modifier_map["Shift"])
+
+
+def key_click(widget, key, delay=0):
+    """ Performs a key click of the given key on the given widget after
+    a delay.
+
+    Parameters
+    ----------
+    widget : wxObject
```
kitchoi, line 33: This function requires a `wx.TextCtrl` to use `EmulateKeyPress`. Let's be specific about this type requirement.

traitsui/testing/tester/exceptions.py
```diff
@@ -15,6 +15,17 @@ class TesterError(Exception):
     pass
 
 
+class SimulationError(Exception):
```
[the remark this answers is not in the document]
aaronayres35: I agree, I think we can add SimulationError when a need arises, and just have Disabled subclass TesterError for now

traitsui/testing/tester/qt4/implementation/text_editor.py
```diff
@@ -0,0 +1,77 @@
...
+    Notes
+    -----
+    Qt SimpleEditors occassionally use QtGui.QTextEdit as their control, and
+    other times use QtGui.QLineEdit
+    '''
+    if isinstance(wrapper.target.control, QtGui.QLineEdit):
+        return wrapper.target.control.displayText()
+    elif isinstance(wrapper.target.control, QtGui.QTextEdit):
+        return wrapper.target.control.toPlainText()
+
```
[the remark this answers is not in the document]
kitchoi: > Generally use of isinstance is a code smell

I need to be more specific actually: If the `isinstance` check is against an interface or an ABC, that would not be a code smell.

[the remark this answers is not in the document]
aaronayres35: I made this change and moved the function to helpers.  It now takes the argument widget which should be a Widget and should be one of 1) QtGui.QLineEdit 2) QtGui.QTextEdit or 3) QtGui.QLabel.  
Maybe I should make the current else for the QtGui.QLabel case an elif, and then have an else where I raise some form of exception? what do you think?

traitsui/testing/tester/wx/implementation/text_editor.py
```diff
@@ -0,0 +1,86 @@
...
+    other times use wx.StaticText.
+    '''
+    if isinstance(wrapper.target.control, wx.TextCtrl):
+        return wrapper.target.control.GetValue()
+    elif isinstance(wrapper.target.control, wx.StaticText):
+        return wrapper.target.control.GetLabel()
+    raise TypeError("readonly_displayed_text_handler expected a UIWrapper with"
+                    " a ReadonlyEditor as a target. ReadonlyEditor control"
+                    " should always be either wx.TextCtrl, or wx.StaticText." 
+                    " {} was found".format(wrapper.target.control))
```
aaronayres35, line 43: This raise was a bit awkward because if someone passes a wrapper that does not have a target with a control, three would already be an error.  Maybe this should be rewritten to not take a wrapper, and instead take just a wx object, more like what was done for the qt similar function.  However, unlike with qt this wouldn't really be reusable, and in this way it can be used directly as the handler which was nice.  Any suggestions on how to raise a better error or if this is okay?

traitsui/testing/tester/qt4/implementation/text_editor.py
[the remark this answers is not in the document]
kitchoi: > Maybe I should make the current else for the QtGui.QLabel case an elif, and then have an else where I raise some form of exception? what do you think?

Sounds good to me!

traitsui/testing/tester/wx/implementation/text_editor.py
kitchoi: Thank you. I think this error is okay in this context: This exception is for capturing programming error on our part. It is better to fail early than doing something unexpected silently (returning None).

traitsui/testing/tester/wx/helpers.py
```diff
@@ -11,16 +11,57 @@
...
+                    key, sorted(mapping)
+                ))
+        else:
+            wx.MilliSleep(delay)
+            key_event = wx.KeyEvent(wx.wxEVT_CHAR)
+            key_event.SetUnicodeKey(KEY)
+            widget.EmulateKeyPress(key_event)
+    else:
+        wx.MilliSleep(delay)
+        KEY = mapping[key]
```
kitchoi, line 50: This `KEY` is unused.

traitsui/tests/editors/test_text_editor.py
```diff
@@ -196,77 +165,174 @@ def test_custom_editor_init_and_dispose_no_auto_set(self):
...
+            name_field.perform(command.KeyClick("Enter"))
+            display_name = name_field.inspect(query.DisplayedText())
+            self.assertEqual(foo.name, "NEW")
+            self.assertEqual(display_name, foo.name)
 
-            key_press_return(editor)
-            process_cascade_events()
-
+    # If auto_set is false, the value can be set by killing the focus. This
+    # can be done by simply moving to another textbox.
```
kitchoi, line 201: 👍

kitchoi (approved):
LGTM. Thank you.

Issue #1051, opened 2020-07-22
Errors in doc examples


aaronayres35:
This issue relates to the example code found in
tutorials - https://github.com/enthought/traitsui/tree/master/examples
(a sub-issue of #1027 )

When I run the tree_editor.py file everything runs fine, but if I click on an employees name, and then click in any textbook I get the error:
```
$ python tree_editor.py 
Traceback (most recent call last):
  File "/Users/aayres/.edm/envs/general-use/lib/python3.6/site-packages/traitsui/qt4/ui_panel.py", line 338, in sizeHint
    if ui.view.width > 0:
AttributeError: 'NoneType' object has no attribute 'width'
Abort trap: 6
```

Also, initially when I ran view_standalone.py, view_multi_object.py, and wizard.py the window closed immediately.  Quickly realized this was because I needed to specify the toolkit with 
```
from traits.etsconfig.api import ETSConfig
ETSConfig.toolkit = 'wx'
```
Should these lines be included in the wx files because traitsui defaults to trying PyQt4 first?  It is probably unnecessary, but just a thought.  Also, when these files are run, I get strange output (this has been noticed before - see Pr #1044 )
```
$ python view_multi_object.py 
09:22:04 AM: Debug: Adding duplicate image handler for 'PNG file'
09:22:04 AM: Debug: Adding duplicate image handler for 'JPEG file'
09:22:04 AM: Debug: Adding duplicate image handler for 'TIFF file'
09:22:04 AM: Debug: Adding duplicate image handler for 'GIF file'
09:22:04 AM: Debug: Adding duplicate image handler for 'PNM file'
09:22:04 AM: Debug: Adding duplicate image handler for 'PCX file'
09:22:04 AM: Debug: Adding duplicate image handler for 'IFF file'
09:22:04 AM: Debug: Adding duplicate image handler for 'Windows icon file'
09:22:04 AM: Debug: Adding duplicate image handler for 'Windows cursor file'
09:22:04 AM: Debug: Adding duplicate image handler for 'Windows animated cursor file'
09:22:04 AM: Debug: Adding duplicate image handler for 'TGA file'
09:22:04 AM: Debug: Adding duplicate image handler for 'XPM file'
09:22:04 AM: Debug: Adding duplicate image handler for 'Windows bitmap file'
09:22:04 AM: Debug: Adding duplicate animation handler for '1' type
09:22:04 AM: Debug: Adding duplicate animation handler for '2' type
[...]

PR #1164, opened 2020-08-26
[WIP] Fix AttributeError accessing removed objects during UI.dispose

### LINKED ISSUE #907: Crash when using UI.dispose() with a TableEditor nested in a ListEditor

### LINKED ISSUE #1145: UI disposal may cause AttributeError on None

kitchoi:
Closes #1145
Closes #907
Replaces the workaround in the second item of #897
Also resolve the second item in #1073 (note that #1161 also fixes it)

Understanding the issue
---
In the `UI.dispose` code path, it first asks the editors to clean up themselves. This involves resetting some of the attributes (e.g. `control`, `factory`) to `None`. I do not know why they need to be unset, presumably to avoid holding references to GUI toolkit objects, but this is the root cause of the AttributeError. I assume we are keeping this attribute removal. The goal is therefore to figure out why these attributes are still being accessed after they are removed.

After all the editors have reset its attribute, the UI then asks the GUI to delete widgets. In Qt, that request is asynchronous via the use of `deleteLater`. However, the attributes on the editors are unset synchronously.

A number of scenarios can cause the AttributeError to occur.

For example, an application typically has many nested widgets. While the application is running and when a nested widget is disposed, the neighbouring widgets need to respond by adjusting their sizes, possibly involving repainting themselves. An example of a widget being disposed of would be an item being removed from a list while a custom ListEditor is used. 

In the context of an entire UI disposal, all the nested widgets are recursively disposed of. Consequently a nested widget will have an event for resizing / repainting prior to its delete instruction being delivered. Note that a UI is not necessary the main window, it could also be nested (e.g. custom ListEditor and TreeEditor both have nested UI objects). This can happen while the GUI is still being active in a production setting.

This scenario is relevant for both Qt and Wx: In Qt, Qt often drives these resizing / repainting events automatically. In Wx, we typically have event handlers for monitoring losing focus on a widget (Qt typically provides this sort of low-level handling already, e.g. `QLineEdit.editingFinished` signal). 

This issue not only affects tests, but also production environment, as we have seen in #1073 and we have evidence of it from external reports (see discussion in #897).

Trying to unset the attributes asynchronously by queuing that action in the event loop resolves the issue in some situations (see https://github.com/enthought/traitsui/pull/910), but it does not resolve the issue entirely because it still relies on event ordering. I verified this not being able to fix a test I added in this PR, neither could it fix the second item in #1073.

Fix
---
The fix is to hide the GUI toolkit control recursively early in `UI.dispose`. This disables the widgets, preventing the cascade of repaint / resizing events when the UI is being disposed of and is entering an inconsistent state.

This fix is not dissimilar in principle compared to a previous attempt fixing the same AttributeError: #46
However, the fix was effectively reverted in https://github.com/enthought/traitsui/pull/283

The fix here is different in the sense that instead of asking for the nested controls to be _destroyed_, we are hiding them, i.e. disabling them. Hiding is a safe and idempotent operation - we can call it many times without an issue, whereas destroy isn't.

Hiding the control early is consistent with what happens when a user closes a GUI window. If we inspect the events being sent while we close a GUI window, we can see a HideEvent being emitted.

Note that initially I thought hiding a parent Qt widget would hide all of its children: That is not true. Removing the `hide_children` line in `UI` in this PR will cause the test I added to fail. It is necessary to hide children widgets explicitly and recursively.

This PR
---
- Adds tests that independently reproduce the AttributeError.
- Reverted the workaround in TabularEditor. I updated the test that would reproduce the AttributeError without the workaround, and then verified that the fix in `UI.dispose` fixes the issue.

I think this makes a lot of `if self.control is not None` or `if self.factory is not None` guards unnecessary.

### DIFF
--- a/traitsui/qt4/tabular_editor.py
+++ b/traitsui/qt4/tabular_editor.py
@@ -929,7 +929,8 @@ class _TableView(QtGui.QTableView):
             if self._user_widths is None:
                 self._user_widths = [None] * len(self._editor.adapter.columns)
             self._user_widths[index] = new
-            if not self._editor.factory.auto_resize:
+            if (self._editor.factory is not None
+                    and not self._editor.factory.auto_resize):
                 self.resizeColumnsToContents()
 
     @contextmanager
--- a/traitsui/qt4/tabular_editor.py
+++ b/traitsui/qt4/tabular_editor.py
@@ -929,8 +929,7 @@ class _TableView(QtGui.QTableView):
             if self._user_widths is None:
                 self._user_widths = [None] * len(self._editor.adapter.columns)
             self._user_widths[index] = new
-            if (self._editor.factory is not None
-                    and not self._editor.factory.auto_resize):
+            if not self._editor.factory.auto_resize:
                 self.resizeColumnsToContents()
 
     @contextmanager
--- a/traitsui/tests/editors/test_tabular_editor.py
+++ b/traitsui/tests/editors/test_tabular_editor.py
@@ -417,8 +417,10 @@ class TestTabularEditor(UnittestTools, unittest.TestCase):
             editor, = ui.get_editors("people")
             editor.adapter.columns = [("Name", "name")]
             # Avoid calling process_cascade_events here.
-            ui.dispose()
-            process_cascade_events()
+            try:
+                ui.dispose()
+            finally:
+                process_cascade_events()
 
     @contextlib.contextmanager
     def report_and_editor(self, view):
--- a/traitsui/tests/editors/test_tabular_editor.py
+++ b/traitsui/tests/editors/test_tabular_editor.py
@@ -399,14 +399,26 @@ class TestTabularEditor(UnittestTools, unittest.TestCase):
             editor.adapter.columns = [("Name", "name")]
             process_cascade_events()
 
-    def test_view_column_resized_attribute_error_workaround(self):
-        # This tests the workaround which checks if `factory` is None before
-        # using it while resizing the columns.
-        # The resize event is processed after UI.dispose is called.
-        # Maybe related to enthought/traits#431
-        with reraise_exceptions(), \
-                self.report_and_editor(get_view()) as (_, editor):
+    def test_view_column_resized_not_called_on_disposed(self):
+        # A workaround / guard was added to check if `factory` is None
+        # in _TableView.columnResized. (see enthought/traitsui##897)
+        # This was caused by the factory attribute being removed as part of
+        # UI.dispose when there are resize events left unprocessed.
+        report = Report(
+            people=[
+                Person(name="Theresa", age=60),
+                Person(name="Arlene", age=46),
+                Person(name="Karen", age=40),
+            ]
+        )
+        view = get_view()
+        with reraise_exceptions():
+            ui = report.edit_traits(view=view)
+            editor, = ui.get_editors("people")
             editor.adapter.columns = [("Name", "name")]
+            # Avoid calling process_cascade_events here.
+            ui.dispose()
+            process_cascade_events()
 
     @contextlib.contextmanager
     def report_and_editor(self, view):
--- a/traitsui/tests/editors/test_tabular_editor.py
+++ b/traitsui/tests/editors/test_tabular_editor.py
@@ -399,29 +399,6 @@ class TestTabularEditor(UnittestTools, unittest.TestCase):
             editor.adapter.columns = [("Name", "name")]
             process_cascade_events()
 
-    def test_view_column_resized_not_called_on_disposed(self):
-        # A workaround / guard was added to check if `factory` is None
-        # in _TableView.columnResized. (see enthought/traitsui##897)
-        # This was caused by the factory attribute being removed as part of
-        # UI.dispose when there are resize events left unprocessed.
-        report = Report(
-            people=[
-                Person(name="Theresa", age=60),
-                Person(name="Arlene", age=46),
-                Person(name="Karen", age=40),
-            ]
-        )
-        view = get_view()
-        with reraise_exceptions():
-            ui = report.edit_traits(view=view)
-            editor, = ui.get_editors("people")
-            editor.adapter.columns = [("Name", "name")]
-            # Avoid calling process_cascade_events here.
-            try:
-                ui.dispose()
-            finally:
-                process_cascade_events()
-
     @contextlib.contextmanager
     def report_and_editor(self, view):
         """
--- a/integrationtests/test_all_examples.py
+++ b/integrationtests/test_all_examples.py
@@ -302,11 +302,6 @@ class TestExample(unittest.TestCase):
                     # before running the next one.
                     process_cascade_events()
 
-                    # XXX temporary change to see how Appveyor + OSX + Qt4
-                    # reacts
-                    for _ in range(10):
-                        process_cascade_events()
-
         # Report skipped files
         for file_path, reason in skipped_files:
             with self.subTest(file_path=file_path):
--- a/integrationtests/test_all_examples.py
+++ b/integrationtests/test_all_examples.py
@@ -229,8 +229,10 @@ def replaced_configure_traits(
     )
     with reraise_exceptions():
         process_cascade_events()
-        ui.dispose()
-        process_cascade_events()
+        try:
+            ui.dispose()
+        finally:
+            process_cascade_events()
 
 
 @contextlib.contextmanager
--- a/integrationtests/test_all_examples.py
+++ b/integrationtests/test_all_examples.py
@@ -302,6 +302,11 @@ class TestExample(unittest.TestCase):
                     # before running the next one.
                     process_cascade_events()
 
+                    # XXX temporary change to see how Appveyor + OSX + Qt4
+                    # reacts
+                    for _ in range(10):
+                        process_cascade_events()
+
         # Report skipped files
         for file_path, reason in skipped_files:
             with self.subTest(file_path=file_path):
--- a/integrationtests/test_all_examples.py
+++ b/integrationtests/test_all_examples.py
@@ -229,13 +229,6 @@ def replaced_configure_traits(
     )
     with reraise_exceptions():
         process_cascade_events()
-
-        # Temporary fix for enthought/traitsui#907
-        if is_qt():
-            ui.control.hide()
-        if is_wx():
-            ui.control.Hide()
-
         ui.dispose()
         process_cascade_events()
 
--- a/traitsui/qt4/toolkit.py
+++ b/traitsui/qt4/toolkit.py
@@ -378,9 +378,15 @@ class GUIToolkit(Toolkit):
     def hide_control(self, control):
         """ Hide a GUI toolkit control."""
         control.hide()
+
+    def hide_children(self, control):
+        """ Recursively hide all of the children controls of a specified GUI
+        toolkit control.
+        """
         for w in control.children():
             if isinstance(w, QtGui.QWidget):
                 self.hide_control(w)
+                self.hide_children(w)
 
     def destroy_control(self, control):
         """ Destroys a specified GUI toolkit control.
--- a/traitsui/qt4/toolkit.py
+++ b/traitsui/qt4/toolkit.py
@@ -378,6 +378,9 @@ class GUIToolkit(Toolkit):
     def hide_control(self, control):
         """ Hide a GUI toolkit control."""
         control.hide()
+        for w in control.children():
+            if isinstance(w, QtGui.QWidget):
+                self.hide_control(w)
 
     def destroy_control(self, control):
         """ Destroys a specified GUI toolkit control.
--- a/traitsui/qt4/toolkit.py
+++ b/traitsui/qt4/toolkit.py
@@ -375,6 +375,10 @@ class GUIToolkit(Toolkit):
         """
         event.ignore()
 
+    def hide_control(self, control):
+        """ Hide a GUI toolkit control."""
+        control.hide()
+
     def destroy_control(self, control):
         """ Destroys a specified GUI toolkit control.
         """
@@ -384,7 +388,7 @@ class GUIToolkit(Toolkit):
 
         # This may be called from within the finished() signal handler so we
         # need to do the delete after the handler has returned.
-        control.hide()
+        self.hide_control(control)
         control.deleteLater()
 
         # PyQt v4.3.1 and earlier deleteLater() didn't transfer ownership to
--- a/traitsui/tests/test_ui.py
+++ b/traitsui/tests/test_ui.py
@@ -303,19 +303,53 @@ if is_qt():
 
 if is_wx():
 
-    # Similar AttributeError can occur if one does not explicitly remove a
-    # bound event handler.
+    # Mimic causing a cascade of events where a handler to be called
+    # will require control/factory/... attributes that will be set
+    # to None in dispose. e.g. In enthought/traitsui#1073, the user closes the
+    # UI and that action causes a cascade of new events.
     import wx
+    from traitsui.wx.helper import TraitsUIPanel
+
 
     class EditorWithCustomWidget(ToolkitSpecificEditor):  # noqa: F811
 
         def init(self, parent):
-            self.control = wx.TextCtrl(parent, -1, "dummy")
-            self.control.Bind(wx.EVT_KILL_FOCUS, self._access_attribute)
-            self.control.SetFocus()
+            self.control = TraitsUIPanel(parent, -1)
 
-        def _access_attribute(self, event):
-            assert self.factory is not None
+            sizer = wx.BoxSizer(wx.HORIZONTAL)
+            self.control.text1 = wx.TextCtrl(parent, -1, "dummy")
+            sizer.Add(self.control.text1, 0, wx.LEFT | wx.EXPAND, 4)
+            self.control.text2 = wx.TextCtrl(parent, -1, "dummy")
+            sizer.Add(self.control.text2, 0, wx.LEFT | wx.EXPAND, 4)
+
+            self.control.text1.Bind(
+                wx.EVT_KILL_FOCUS, self._update_with_attr_access,
+            )
+            self.control.text2.Bind(
+                wx.EVT_KILL_FOCUS, self._update_with_attr_access,
+            )
+            self.control.SetSizerAndFit(sizer)
+            self._count = 0
+
+        def _update_with_attr_access(self, event):
+            if self._count > 1:
+                # prevent infinite loop
+                return
+
+            # Test objective: We should not run into AttributeError here
+            assert self.control is not None
+
+            if event.GetEventObject() is self.control.text1:
+                self.control.text2.SetFocus()
+                self._count += 1
+            if event.GetEventObject() is self.control.text2:
+                self.control.text1.SetFocus()
+                self._count += 1
+
+        def update_editor(self):
+            # Trigger cascade of events at initialization.
+            self.control.text1.SetFocus()
+            self.control.text2.SetFocus()
 
         def dispose(self):
             super().dispose()
@@ -333,6 +367,10 @@ class TestUIDispose(unittest.TestCase):
         obj = DummyObject()
         view = View(
             Group(
+                Item(
+                    "number",
+                    editor=BasicEditorFactory(klass=EditorWithCustomWidget),
+                ),
                 Item(
                     "number",
                     editor=BasicEditorFactory(klass=EditorWithCustomWidget),
--- a/traitsui/tests/test_ui.py
+++ b/traitsui/tests/test_ui.py
@@ -18,20 +18,29 @@ Test cases for the UI object.
 """
 
 import unittest
+from unittest import mock
 
-from traits.api import Property
+from pyface.api import GUI, Window
+from traits.api import Any, Bool, Instance, Property
 from traits.has_traits import HasTraits, HasStrictTraits
 from traits.trait_types import Str, Int
 import traitsui
+from traitsui.api import Group, View
 from traitsui.item import Item, spring
-from traitsui.view import View
 
+from traitsui.basic_editor_factory import BasicEditorFactory
+from traitsui.editor import Editor
 from traitsui.tests._tools import (
     count_calls,
     create_ui,
+    is_qt,
+    is_wx,
+    process_cascade_events,
     requires_toolkit,
+    reraise_exceptions,
     ToolkitName,
 )
+from traitsui.toolkit import toolkit, toolkit_object
 
 
 class FooDialog(HasTraits):
@@ -240,3 +249,85 @@ class TestUI(unittest.TestCase):
 
             obj.name = "too short"
             self.assertTrue(editor.invalid)
+
+
+# Regression test on an AttributeError commonly seen (enthought/traitsui#1145)
+# Code in ui_panel makes use toolkit specific attributes on the toolkit specific
+# Editor
+ToolkitSpecificEditor = toolkit_object("editor:Editor")
+
+
+if is_qt():
+
+    from pyface.qt import QtGui, QtCore
+
+    class CustomWidget(QtGui.QLabel):
+
+        def __init__(self, editor, parent=None):
+            super(CustomWidget, self).__init__()
+            self._some_editor = editor
+
+        def sizeHint(self):
+            assert self._some_editor.factory is not None
+            return super().sizeHint()
+
+
+    class EditorWithCustomWidget(ToolkitSpecificEditor):
+
+        def init(self, parent):
+            from pyface.qt import QtCore, QtGui
+
+            # To reproduce an assertion from sizeHint, it is necessary to have
+            # two widgets, where one of them is created with a nested UI.
+            # When the nested UI is disposed, the original AttributeError is
+            # caused by the neighboring widgets trying to resize / repaint /
+            # ... adjust to fit the layout. These do not happen if the widgets
+            # are made to be hidden first before dispose is called.
+            self.control = QtGui.QSplitter(QtCore.Qt.Horizontal)
+            widget = CustomWidget(editor=self)
+            self.control.addWidget(widget)
+            self.control.setStretchFactor(0, 2)
+            self._ui = self.edit_traits(
+                parent=self.control,
+                kind="subpanel",
+                view=View(Item("_", label="DUMMY"), width=100, height=100),
+            )
+            self.control.addWidget(self._ui.control)
+
+        def dispose(self):
+            self._ui.dispose()
+            super().dispose()
+
+        def update_editor(self):
+            pass
+
+
+class DummyObject(HasTraits):
+
+    number = Int()
+
+
+class TestUIDispose(unittest.TestCase):
+
+    @requires_toolkit([ToolkitName.qt])
+    def test_absence_of_attribute_error_from_dispose(self):
+        obj = DummyObject()
+        view = View(
+            Group(
+                Item(
+                    "number",
+                    editor=BasicEditorFactory(klass=EditorWithCustomWidget),
+                ),
+            ),
+        )
+        with reraise_exceptions():
+
+            # create_ui is useful for other tests but not this one: it
+            # flushes the event loop prior to calling dispose. Here we want
+            # to test even if the event loop is not flushed before calling
+            # dispose, it would still be okay at the end.
+            ui = obj.edit_traits(view=view)
+            try:
+                ui.dispose()
+            finally:
+                process_cascade_events()
--- a/traitsui/tests/test_ui.py
+++ b/traitsui/tests/test_ui.py
@@ -18,10 +18,8 @@ Test cases for the UI object.
 """
 
 import unittest
-from unittest import mock
 
-from pyface.api import GUI, Window
-from traits.api import Any, Bool, Instance, Property
+from traits.api import Property
 from traits.has_traits import HasTraits, HasStrictTraits
 from traits.trait_types import Str, Int
 import traitsui
@@ -29,7 +27,6 @@ from traitsui.api import Group, View
 from traitsui.item import Item, spring
 
 from traitsui.basic_editor_factory import BasicEditorFactory
-from traitsui.editor import Editor
 from traitsui.tests._tools import (
     count_calls,
     create_ui,
@@ -40,7 +37,7 @@ from traitsui.tests._tools import (
     reraise_exceptions,
     ToolkitName,
 )
-from traitsui.toolkit import toolkit, toolkit_object
+from traitsui.toolkit import toolkit_object
 
 
 class FooDialog(HasTraits):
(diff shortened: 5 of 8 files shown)


traitsui/tests/editors/test_tabular_editor.py
```diff
@@ -399,14 +399,28 @@ def test_adapter_columns_changes(self):
...
+        view = get_view()
+        with reraise_exceptions():
+            ui = report.edit_traits(view=view)
+            editor, = ui.get_editors("people")
             editor.adapter.columns = [("Name", "name")]
+            # Avoid calling process_cascade_events here.
+            try:
+                ui.dispose()
+            finally:
+                process_cascade_events()
```
kitchoi, line 423: I updated the test because later changes to use `create_ui` masks the issue by calling `process_cascade_events` prior to UI.dispose.

kitchoi: Actually I think we can remove this test now that the workaround is removed too.

traitsui/qt4/tabular_editor.py
```diff
@@ -929,8 +929,7 @@ def columnResized(self, index, old, new):
             if self._user_widths is None:
                 self._user_widths = [None] * len(self._editor.adapter.columns)
             self._user_widths[index] = new
-            if (self._editor.factory is not None
```
kitchoi, line 932: I _could_ keep this but I'd prefer not. Other places in this class uses `factory` without checking if the value is None.

corranwebster:
I think with hide "not working recursively" I think if you hide the a widget, then all it's _children_ should also be hidden.  I suspect that there may be an issue here where:

- some complex widgets use sibling-level widgets together with a layout (rather than say having a panel which contains them all and is a parent to them), which is valid, and so you will need to hide them all individually
- potentially errors where things are _not_ being parented correctly; it may be worth investigating this.

But you shouldn't need to blindly recurse to hide all widgets.

Additionally, hiding may not be enough if some of the issues are coming from requests for size information.  Under some circumstances a hidden widget may still take up space in the layout (or Sizer, under Wx), and so may still be asked for its size, even though it is hidden.  I don't know if this can happen in TraitsUI code, but it may be prudent to, in addition to hiding the widget, remove it from any layout that it is part of.  This may be awkward, because I don't think TraitsUI keeps direct references to the layout objects, so you'd either need to start keep those references somewhere, or use toolkit methods to try and discover them.

traitsui/qt4/tabular_editor.py
kitchoi: Aha this causes the integration tests (which open and close a lot of GUI) to have this failure in a subtest that has nothing to do with tabular editor:
```
FAIL: test_run (test_all_examples.TestExample) (file_path='/Users/appveyor/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/traitsui/examples/demo/Advanced/Time_editor_demo.py')
  File "/Users/appveyor/projects/traitsui/integrationtests/test_all_examples.py", line 295, in test_run
    run_file(file_path)
RuntimeError: Uncaught exceptions found.
=== Exception (type: <class 'AttributeError'>, value: 'NoneType' object has no attribute 'auto_resize') ===
  File "/Users/appveyor/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/traitsui/qt4/tabular_editor.py", line 932, in columnResized
    if not self._editor.factory.auto_resize:
AttributeError: 'NoneType' object has no attribute 'auto_resize'
```
This is in Mac OSX and Qt4. Note that the `Time_editor_demo` has nothing to do with TabularEditor. This is clearly caused by events being left unprocessed at the end of a test.  We have already identified the approach to process cascaded events not to work in OSX and Qt4 (see #951) despite Qt says it should (really, Qt5 says it should).

kitchoi:
> I think with hide "not working recursively" I think if you hide the a widget, then all it's children should also be hidden. I suspect that there may be an issue here where:
>
> some complex widgets use sibling-level widgets together with a layout (rather than say having a panel which contains them all and is a parent to them), which is valid, and so you will need to hide them all individually
potentially errors where things are not being parented correctly; it may be worth investigating this.
But you shouldn't need to blindly recurse to hide all widgets.

Thank you. 
I think I am still a bit confused here...  I am trying to hide the control using `control.hide()` and then call `.hide()` on the widgets returned by `control.children()`.  If there are widgets that are not parented correctly, they would not be returned by `.children()` and I would not be hiding them. Is there something I missed?

On the other hand, as far as I understand, hiding is safe and idempotent. If a widget is already hidden, hiding it again is not a problem.

> Additionally, hiding may not be enough if some of the issues are coming from requests for size information. Under some circumstances a hidden widget may still take up space in the layout (or Sizer, under Wx), and so may still be asked for its size, even though it is hidden. I don't know if this can happen in TraitsUI code, but it may be prudent to, in addition to hiding the widget, remove it from any layout that it is part of. This may be awkward, because I don't think TraitsUI keeps direct references to the layout objects, so you'd either need to start keep those references somewhere, or use toolkit methods to try and discover them.

While I have not been able to reproduce the error in a different scenario where the hiding action does not fix the issue, I would not claim this would necessarily fix all the issues - I think we will just have to wait for these issues to appear so that we have a mean to investigate and fix. I am finding it difficult to understand the circumstances you describe. If there are some actual code, ideally a test, that would help me to understand. Otherwise I don't know what I can do about that.

corranwebster:
Yes, if you call `control.hide()` then all children of that widget are effectively hidden (even if they have not had `hide()` called explicitly), so you shouldn't _have_ to (but yes, it is idempotent, and in this particular circumstance, safe to do; but in general recursing the children to hide them is not a pattern that you should need).  In particular this comment isn't true:

> Note that initially I thought hiding a parent Qt widget would hide all of its children: That is not true.

Relevant documentation of visibility is here: https://doc.qt.io/qt-5/qwidget.html#visible-prop but the key takeaway is that when a widget is hidden, all its children are no longer visible; to be visible the widget and all its ancestors cannot be hidden.  (I am reasonably confident that wx works similarly at least in terms of effective behaviour if not exact implementation details).

I heartily approve of hiding the top-level widgets, and think it's likely part of the solution.

The fact that when you do an extra thing that you shouldn't have to do and it makes the problem go away, it makes me think that there is in fact something else going on; my suspicion is that it is something to do with layout participation, and so I'd encourage you to look more closely at the cases where hiding the top level widget but not the children fails.

My intuition would suggest things like:

- does manually removing the parent widget from layouts also fix the problem in these cases?
- does manually removing the children from layouts fix the problem?
- what happens if you manually set the `sizePolicy` to have the [retainSizeWhenHidden](https://doc.qt.io/qt-5/qsizepolicy.html#setRetainSizeWhenHidden) flag set?

kitchoi:
Thank you. I shared the uneasiness with recursively explicitly hiding the children widgets because I too believe hiding a widget should hide the children widgets. After hiding a parent widget, when I inspect the `isHidden` on a child widget, I still see it returns false (Qt documentation is somewhat vague about whether it is equivalent to not isVisible). I will try to come up with a Qt-only example to verify that.

> does manually removing the parent widget from layouts also fix the problem in these cases?
> does manually removing the children from layouts fix the problem?

By removing, do you mean also mean calling deleteLater on the children?

Qt's doc on `removeItem`:
> Removes the layout item item from the layout. It is the caller's responsibility to delete the item.

This would be similar to #46 (reverted in #283 to the segfaults it causes). I could still try that but I am really not comfortable with manually deleting nested child widgets.

corranwebster:
> Thank you. I shared the uneasiness with recursively explicitly hiding the children widgets because I too believe hiding a widget should hide the children widgets. After hiding a parent widget, when I inspect the `isHidden` on a child widget, I still see it returns false (Qt documentation is somewhat vague about whether it is equivalent to not isVisible). I will try to come up with a Qt-only example to verify that.

`isHidden()` != `isVisible()`, and it's fairly clear when you work out how things have to work.  You can have something which is not hidden but also not visible.  As soon as you hide a widget, it and all its children are no longer _visible_, but the children are not hidden.  If you then hide one of the children, and show the parent, all the non-hidden children will become visible, but the hidden one will not be shown.

So "hidden" is about intent, "visible" is about current state (at least in Qt).

But just because something is hidden doesn't mean it can't affect layout, depending on the size policy in operation.

> > does manually removing the parent widget from layouts also fix the problem in these cases?
> > does manually removing the children from layouts fix the problem?
> 
> By removing, do you mean also calling deleteLater on the children?

No.  Not `deleteLater` (and certainly not recursive `deleteLater`, which also shouldn't be needed), I mean calling `removeWidget` (or `removeItem`) on the appropriate layout object.  Removal from layout != deletion.

The widget will be removed from the layout automatically on actual deletion, but suspicion is that manually removing the widget from its layout might fix these problems because it seems like they are because something is asking for a size at a time before deletion when we don't have enough info to answer that question.

The hard part is that as far as I know there is no way to ask a widget "which layout do you belong to?" which is really what we want.  It's possible that re-parenting the widget to `None` might work:
```
self.control.hide()
self.control.setParent(None)
self.control.deleteLater()
```
since it has the side-effect of removing the control from layout (effectively making it an invisible top-level window).  This trick may not work on wx, though (there is a `Reparent()` call, I don't know if `Reparent(None)` is valid, and what it does to the `Sizer`).

The alternative would be getting the widget's parent's layout and searching through it and it's sub-layouts until you find the widget, and then removing it from the appropriate layout.

You should only need to do this with top-level widgets.  It is possible that for some particular, complex widgets you may need some additional special handling.

kitchoi:
I think I know why when I hide the parent widget, the children widgets are not hidden: Because, calling `hide` only post the HideEvent, the event is not delivered or processed yet.  If I add `GUI.process_events()` after `toolkit().hide_control(self.control)` but remove `toolkit().hide_children(self.control)`, then the children widgets are hidden too.

However I think it would be bad to call process events in UI.dispose. 

Does this relieve your concern?

kitchoi:
> No. Not deleteLater (and certainly not recursive deleteLater, which also shouldn't be needed), I mean calling removeWidget (or removeItem) on the appropriate layout object. Removal from layout != deletion.
>
> The widget will be removed from the layout automatically on actual deletion, but suspicion is that manually removing the widget from its layout might fix these problems because it seems like they are because something is asking for a size at a time before deletion when we don't have enough info to answer that question.

If I try to remove items from layout but not call `deleteLater` on them, should I worry about causing Qt object leaks?

kitchoi:
> I think I know why when I hide the parent widget, the children widgets are not hidden: Because, calling hide only post the HideEvent, the event is not delivered or processed yet. If I add GUI.process_events() after toolkit().hide_control(self.control) but remove toolkit().hide_children(self.control), then the children widgets are hidden too.

Scrap that. I should not be trying to do that in the test with TraitsUI, a lot of other things happen in process_events. Here is the pure Qt code.
`hide` is called on parent, but its children are not hidden. The children are not visible. That's consistent with "hide" being a request / intent. The children are not visible because its parent is hidden, but the children themselves are not hidden.
```
>>> app = QtGui.QApplication([])
>>> widget = QtGui.QWidget()
>>> layout = QtGui.QVBoxLayout(widget)
>>> w2 = QtGui.QWidget()
>>> w3 = QtGui.QWidget()
>>> widget.isHidden()
True
>>> w2.isHidden()
True
>>> layout.addWidget(w2)
>>> w2.isHidden()
False
>>> w2.parent() is widget
True
>>> layout.addWidget(w3)
>>> w3.parent() is widget
True
>>> w3.isHidden()
False
>>> widget.show()
>>> w2.isHidden()
False
>>> w2.isVisible()
True
>>> widget.hide()
>>> widget.isHidden()
True
>>> w2.isHidden()
False
>>> from pyface.api import GUI
>>> GUI.process_events()
>>> w2.isHidden()
False
>>> w2.isVisible()
False
```

corranwebster:
Yes, that would make sense as to why you don't see the visibility changing if visibility change is done via event dispatch.  Even so, once the parent widget is hidden, no children should ever get a paint event, so this should solve the problem of errors during paint events without needing to recursively hide the children.

I agree that it would be a bad idea to process events in a dispose method (in particular, this would likely trigger the actual widget destruction right then).

I am still concerned that hiding is "solving" other errors because the children are not being asked for their size or getting resize events when hidden.  However hiding does not _guarantee_ that a widget will not be asked for its size or get a resize event; the only sure way to do that is to remove the widget from the layout.

I am leaning more and more towards (for Qt) the right thing being: hide, unparent, deleteLater for the top-level widget(s).  Try that before doing anything recursive.

corranwebster:
Removing from layout and destruction are distinct things.  Destroying a widget _does_ remove it from layout as a side-effect, but a widget can safely be removed from a layout and it will still be destroyed when its parent is destroyed, because removing from a layout doesn't affect the parent-child relationship.

If you look at the base `QObject` class, you will see that parent/child relationships and destruction handling is there, but layout is not part of the base class.

So in a `destroy` method, if it makes sense, you can safely remove things from layout and they will still get deleted.

kitchoi:
> Even so, once the parent widget is hidden, no children should ever get a paint event, so this should solve the problem of errors during paint events without needing to recursively hide the children.

When I saw in the tests is that the child's sizeHint is still requested when its sibling is removed, and that happens when the children are not hidden (Qt's terminology of "hidden") but the parent is hidden.

For example, this is the hierarchy of objects in my Qt test:
```
 <traitsui.qt4.ui_base._StickyDialog object at 0x7fd9937751f8> layout:  <PyQt5.QtWidgets.QVBoxLayout object at 0x7fd99761cf78> parent:  None
     <PyQt5.QtWidgets.QMainWindow object at 0x7fd9937750d8> layout:  <PyQt5.QtWidgets.QLayout object at 0x7fd99762eaf8> parent:  <traitsui.qt4.ui_base._StickyDialog object at 0x7fd9937751f8>
         <PyQt5.QtWidgets.QWidget object at 0x7fd99762e9d8> layout:  <PyQt5.QtWidgets.QGridLayout object at 0x7fd9976245e8> parent:  <PyQt5.QtWidgets.QMainWindow object at 0x7fd9937750d8>
             <PyQt5.QtWidgets.QLabel object at 0x7fd99762e948> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fd99762e9d8>
             <PyQt5.QtWidgets.QSplitter object at 0x7fd997624f78> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fd99762e9d8>
                 <PyQt5.QtWidgets.QWidget object at 0x7fd99762e798> layout:  <PyQt5.QtWidgets.QVBoxLayout object at 0x7fd99762e828> parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fd997624f78>
                     <PyQt5.QtWidgets.QFrame object at 0x7fd99762e318> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fd99762e798>
                     <PyQt5.QtWidgets.QFrame object at 0x7fd99762e438> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fd99762e798>
                 <traitsui.tests.test_ui.CustomWidget object at 0x7fd99762e558> layout:  None parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fd997624f78>
                 <PyQt5.QtWidgets.QSplitterHandle object at 0x7fd99762eca8> layout:  None parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fd997624f78>
                 <PyQt5.QtWidgets.QSplitterHandle object at 0x7fd99762ed38> layout:  None parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fd997624f78>
```

The main UI control is `_StickyDialog`. The `QSplitter` is the one I created in the test, and it has a stretch factor set and it contains a`CustomWidget` whose sizeHint may raise the AttributeError (the objective of my test). This setup is mimicking the TreeEditor: The node is represented in a nested UI inside a splitter, and the nested UI's sizeHint is overridden to make use of traitsui's View sizing information. 

Suppose I set `_StickyDialog` to be hidden. As shown above, this won't set "isHidden" to true for the great-grand child `QSplitter`. The following is my guess: When one of the widgets inside `QSplitter` is deleted, I suppose that will emit a signal and `QSplitter` will react to that. Because `QSplitter.isHidden()` is false, QSplitter may then try to update its geometry (or its children's geometry) and recalculate the layout. Then QSplitter may request itself and its content to be repainted via another signal, that signal is propagated...and then be rejected by one of its ancestors: No repaint occurs, but layout recalculation still occurs.

Even if we know how and if a hidden parent will prevent all its descendants to resize, unless it is documented (I could not find it yet), this is getting close to relying on implementation details and so I think I want to avoid relying on that.

However I think the "hidden" being a per-object property is not an implementation detail: It is a [documented](https://doc.qt.io/qt-5/qwidget.html#isHidden) design. Given it is idempotent, I think it is acceptable to request all objects in the hierarchy to be hidden when we are about to dispose the UI.

> I am leaning more and more towards (for Qt) the right thing being: hide, unparent, deleteLater for the top-level widget(s). Try that before doing anything recursive.

OK.
I will focus on Qt first, not sure how this translate to Wx but I will worry about that later.
Say I implemented this on `GUIToolkit`.
```
    def hide_unparent_delete_later(self, control):
        control.hide()
        control.setParent(None)
        control.deleteLater()
```

Then in `UI.finish` (most of `UI.dispose` code path), instead of calling `toolkit().hide(self.control); toolkit().hide_children(self.control)`, I call this method:
```
    def finish(self):
        """ Finishes disposing of a user interface.
        """
        print("hiding, unparenting, and deleting {!r} later".format(self.control))
        toolkit().hide_unparent_delete_later(self.control)
        # the rest of the original method
```
Note that in my test I have a nested UI, so this `finish` is called twice, once for the main window UI first, then once for the nested UI.
I also print the object hierarchy before and after main UI dispose. Then I ran the test, this is the output:

<details>
    <summary> Test failure </summary>

```
----Before dispose----
 <traitsui.qt4.ui_base._StickyDialog object at 0x7fdf9d76a1f8> layout:  <PyQt5.QtWidgets.QVBoxLayout object at 0x7fdfa1724f78> parent:  None
     <PyQt5.QtWidgets.QMainWindow object at 0x7fdf9d76a0d8> layout:  <PyQt5.QtWidgets.QLayout object at 0x7fdfa1736af8> parent:  <traitsui.qt4.ui_base._StickyDialog object at 0x7fdf9d76a1f8>
         <PyQt5.QtWidgets.QWidget object at 0x7fdfa17369d8> layout:  <PyQt5.QtWidgets.QGridLayout object at 0x7fdfa172d5e8> parent:  <PyQt5.QtWidgets.QMainWindow object at 0x7fdf9d76a0d8>
             <PyQt5.QtWidgets.QLabel object at 0x7fdfa1736948> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fdfa17369d8>
             <PyQt5.QtWidgets.QSplitter object at 0x7fdfa172df78> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fdfa17369d8>
                 <PyQt5.QtWidgets.QWidget object at 0x7fdfa1736798> layout:  <PyQt5.QtWidgets.QVBoxLayout object at 0x7fdfa1736828> parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fdfa172df78>
                     <PyQt5.QtWidgets.QFrame object at 0x7fdfa1736318> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fdfa1736798>
                     <PyQt5.QtWidgets.QFrame object at 0x7fdfa1736438> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fdfa1736798>
                 <traitsui.tests.test_ui.CustomWidget object at 0x7fdfa1736558> layout:  None parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fdfa172df78>
                 <PyQt5.QtWidgets.QSplitterHandle object at 0x7fdfa1736ca8> layout:  None parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fdfa172df78>
                 <PyQt5.QtWidgets.QSplitterHandle object at 0x7fdfa1736d38> layout:  None parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fdfa172df78>
hiding, unparenting, and deleting <traitsui.qt4.ui_base._StickyDialog object at 0x7fdf9d76a1f8> later
hiding, unparenting, and deleting <PyQt5.QtWidgets.QWidget object at 0x7fdfa1736798> later
----After dispose----
 <traitsui.qt4.ui_base._StickyDialog object at 0x7fdf9d76a1f8> layout:  <PyQt5.QtWidgets.QVBoxLayout object at 0x7fdfa1724f78> parent:  None
     <PyQt5.QtWidgets.QMainWindow object at 0x7fdf9d76a0d8> layout:  <PyQt5.QtWidgets.QLayout object at 0x7fdfa1736af8> parent:  <traitsui.qt4.ui_base._StickyDialog object at 0x7fdf9d76a1f8>
         <PyQt5.QtWidgets.QWidget object at 0x7fdfa17369d8> layout:  <PyQt5.QtWidgets.QGridLayout object at 0x7fdfa172d5e8> parent:  <PyQt5.QtWidgets.QMainWindow object at 0x7fdf9d76a0d8>
             <PyQt5.QtWidgets.QLabel object at 0x7fdfa1736948> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fdfa17369d8>
             <PyQt5.QtWidgets.QSplitter object at 0x7fdfa172df78> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fdfa17369d8>
                 <traitsui.tests.test_ui.CustomWidget object at 0x7fdfa1736558> layout:  None parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fdfa172df78>
                 <PyQt5.QtWidgets.QSplitterHandle object at 0x7fdfa1736d38> layout:  None parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fdfa172df78>
Unexpected error captured in sys excepthook. 
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/tests/test_ui.py", line 272, in sizeHint
    assert self._some_editor.factory is not None
AssertionError
```
</details>

That does not resolve the AttributeError. Is there something else I should be doing here?

corranwebster:
I think we were talking at cross purposes somewhat - I was assuming this was going in the `Editor` `destroy`, not in the UI `finish` method.

I assume that the `finish` method is calling `destroy` for every editor recursively?  Need to check.

I will look into this a bit more, but this is still making me thing the issue is with layout, not visibility.  I have to keep on emphasising that just because you hide things it doesn't mean that they won't be asked for their size, so recursive hide might look like it works, but it is a side-effect and may not work in corner cases.

kitchoi:
I see. So you mean we should be hiding, unparenting and deleting (later) children in every single Editor implementation?
Like I will write that logic in Qt TreeEditor, TableEditor, EnumEditor, TabbedFoldGroupEditor, so on and on?

corranwebster:
The `dispose` method should usually call `super()` (I think) so you should only need to do it in the base editor class for the toolkit (and maybe one or two other places).  Basically anywhere you call `deleteLater(self.control)`.

kitchoi:
> I have to keep on emphasising that just because you hide things it doesn't mean that they won't be asked for their size, so recursive hide might look like it works, but it is a side-effect and may not work in corner cases.

And suppose it does not work in corner cases... Just checking, hiding does not hurt and is consistent with the behaviour when we close the window so I think we can still let it happen, is that right?

corranwebster:
And might also want to do it in `toolkit`'s `destroy_control`.

kitchoi:
> The dispose method should usually call super() (I think) so you should only need to do it in the base editor class for the toolkit (and maybe one or two other places). Basically anywhere you call deleteLater(self.control).

I see. Thank you. Yeah there are two methods called `dispose` so I assumed you meant the one on UI.

I restored the `UI.dispose` to how it is on master, and now I add this to the `Editor` base class `dispose` method:
```
toolkit().hide_unparent_delete_later(self.control)
```
The `hide_unparent_delete_later` is what I showed above.

After the main UI dispose, this is the new object hierarchy so we can confirm the objects are unparented.
The AttributeError is still there.
<details>
    <summary> Test failure </summary>

```
----Before dispose----
 <traitsui.qt4.ui_base._StickyDialog object at 0x7fe425f751f8> layout:  <PyQt5.QtWidgets.QVBoxLayout object at 0x7fe42a94e0d8> parent:  None
     <PyQt5.QtWidgets.QMainWindow object at 0x7fe425f750d8> layout:  <PyQt5.QtWidgets.QLayout object at 0x7fe42a955c18> parent:  <traitsui.qt4.ui_base._StickyDialog object at 0x7fe425f751f8>
         <PyQt5.QtWidgets.QWidget object at 0x7fe42a955af8> layout:  <PyQt5.QtWidgets.QGridLayout object at 0x7fe42a94e708> parent:  <PyQt5.QtWidgets.QMainWindow object at 0x7fe425f750d8>
             <PyQt5.QtWidgets.QLabel object at 0x7fe42a955a68> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fe42a955af8>
             <PyQt5.QtWidgets.QSplitter object at 0x7fe42a955678> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fe42a955af8>
                 <PyQt5.QtWidgets.QWidget object at 0x7fe42a9558b8> layout:  <PyQt5.QtWidgets.QVBoxLayout object at 0x7fe42a955948> parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fe42a955678>
                     <PyQt5.QtWidgets.QFrame object at 0x7fe42a955438> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fe42a9558b8>
                     <PyQt5.QtWidgets.QFrame object at 0x7fe42a955558> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fe42a9558b8>
                 <traitsui.tests.test_ui.CustomWidget object at 0x7fe42a9550d8> layout:  None parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fe42a955678>
                 <PyQt5.QtWidgets.QSplitterHandle object at 0x7fe42a955dc8> layout:  None parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fe42a955678>
                 <PyQt5.QtWidgets.QSplitterHandle object at 0x7fe42a955e58> layout:  None parent:  <PyQt5.QtWidgets.QSplitter object at 0x7fe42a955678>
----After dispose----
 <traitsui.qt4.ui_base._StickyDialog object at 0x7fe425f751f8> layout:  <PyQt5.QtWidgets.QVBoxLayout object at 0x7fe42a94e0d8> parent:  None
     <PyQt5.QtWidgets.QMainWindow object at 0x7fe425f750d8> layout:  <PyQt5.QtWidgets.QLayout object at 0x7fe42a955c18> parent:  <traitsui.qt4.ui_base._StickyDialog object at 0x7fe425f751f8>
         <PyQt5.QtWidgets.QWidget object at 0x7fe42a955af8> layout:  <PyQt5.QtWidgets.QGridLayout object at 0x7fe42a94e708> parent:  <PyQt5.QtWidgets.QMainWindow object at 0x7fe425f750d8>
             <PyQt5.QtWidgets.QLabel object at 0x7fe42a955a68> layout:  None parent:  <PyQt5.QtWidgets.QWidget object at 0x7fe42a955af8>
Unexpected error captured in sys excepthook. 
  File "/Users/kchoi/Work/ETS/traitsui/traitsui/tests/test_ui.py", line 272, in sizeHint
    assert self._some_editor.factory is not None
AssertionError
```

</details>

kitchoi:
> And might also want to do it in toolkit's destroy_control.

Yep, did that too. Same test failure.
I replaced these two lines in `destroy_control`:
```
        control.hide()
        control.deleteLater()
```
with 
```
        self.hide_unparent_delete_later(control)
```
(which is equivalent to adding the `control.setParent(None)` in between the two lines.)

traitsui/qt4/tabular_editor.py
kitchoi: Confirmed this is Qt4 refusing to process all cascaded events in tests. The tests are behaving fine with Qt5 and other platforms. I will put this back for now as there are downstream projects still testing with Qt4. This is the wrong place to do this sort of thing but it is cheap.

traitsui/tests/editors/test_tabular_editor.py
kitchoi: The workaround is to be put back but this test won't fail even if the workaround is removed because the changes in UI.finish introduced in this PR will fix this issue. In other words, the fix to `UI` makes this test confusing, so I will remove this test.

kitchoi:
For the record, if I changed the `_size_hint_wrapper` in `ui_panel` such that it does not hold a reference to the control (it is preventing garbage collection), this PR fixes the first item of https://github.com/enthought/traitsui/issues/1051 and #1166.

kitchoi:
> For the record, if I changed the _size_hint_wrapper in ui_panel such that it does not hold a reference to the control (it is preventing garbage collection), this PR fixes the first item of #1051 and #1166.

Actually, releasing the strong reference to the control in the wrapper fixes #1166 without this PR too. So that's an orthogonal issue then!

kitchoi:
Closing. There is an alternative, potential solution, which is to move the `toolkit().destroy_control(self.control)` in `UI.finish` up to go before calling `UI.reset` (which calls `Editor.dispose`). The key requirement is that for this alternative to work, `destroy_control` should be implemented such that the destroy action is deferred, which is true for Qt but false for Wx.  I will comment on #1145 to provide further details.

PR #1168, opened 2020-08-28
Check that the UI has a valid View before modifying size in _size_hint_wrapper

corranwebster:
It is possible that a `UI` instance has its `View` set to `None` during the lifetime of a widget or layout that is using the `_size_hint_wrapper` (which is itself a hack).  In some circumstances this causes a segfault at UI dispose time.  This PR puts a guard in place to prevent exceptions in this code.

No tests at the moment, because this is an implementation detail of a hack in a private implementation detail, and I'm not sure I want to be formalizing this piece of code that much, but if review thinks it's needed, I will add.

### DIFF
--- a/traitsui/qt4/ui_panel.py
+++ b/traitsui/qt4/ui_panel.py
@@ -335,10 +335,11 @@ def _size_hint_wrapper(f, ui):
 
     def sizeHint():
         size = f()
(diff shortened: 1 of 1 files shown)


kitchoi (approved):
LGTM

PR #1171, opened 2020-08-28
Ui tester api updates4 - Implement locating textbox in RangeEditor and performing KeySequence

aaronayres35:
Implements the sixth bullet on issue #1149 (Implement locating the textbox in RangeEditor and then perform KeySequence event on it.)


traitsui/qt4/range_editor.py
```diff
@@ -449,6 +449,9 @@ def update_object_on_scroll(self, pos):
     def update_object_on_enter(self):
         """ Handles the user pressing the Enter key in the text field.
         """
+        # it is possible we get the event after the control has gone away
+        if self.control is None:
+            return
```
aaronayres35, line 454: @kitchoi This isn't really directly a part of this PR.  However, this check was included in the update_object_on_enter method for the SimpleSliderEditor, but not on LargeRangeSliderEditor for Idk what reason.  When running the tests I was seeing a failure where this was called with self.control as None.  I can't see why this check would ever be a problem (seeing as it is used in the same way in the other editor style).  If needed I can undo this change, and remove the test which fails because of it (or I could make the change in a separate PR).  Let me know if this was reasonable to change or not.

kitchoi: This change looks reasonable to me... persumably this was motivated by a failure of a test you added?

aaronayres35: Note: the same thing was added for wx

traitsui/testing/tester/qt4/implementation/range_editor.py
```diff
@@ -0,0 +1,87 @@
...
+#  is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+#  Thanks for using Enthought open source!
+#
+
+from traitsui.qt4.range_editor import (
+    LargeRangeSliderEditor,
+    LogRangeSliderEditor,
+    RangeTextEditor,
+    SimpleSliderEditor,
```
aaronayres35, line 16: Eventually we will want to add SimpleSpinEditor, but this can be done in a later PR.
For later reference: The SimpleSpinEditor class in Qt has a control that is a QtGui.QSpinBox object, which subclasses QtGui.AbstractSpinBox which has a lineEdit() method which returns the QLineEdit for the spin box (likely we would just need to resolve to this and then the other handlers should work).  For wx however, the SimpleSpinEditor has a control of a wx.SpinCtrl, which is supposedly combines a wx.SpinCtrl and wx.TextCtrl in one.  However, it doesn't seem to subclass wx.TextCtrl and doesn't appear to support the same methods, so our handlers for KeyClick, KeySequence, etc. may not work correctly.

traitsui/tests/editors/test_range_editor_text.py
```diff
@@ -68,39 +70,20 @@ def test_wx_text_editing(self):
...
-            textctrl.SetValue("1")
+            text = tester.find_by_name(ui, "number").locate(locator.WidgetType.textbox)
+            text.perform(command.KeyClick("1"))
+            text.perform(command.KeyClick("Enter"))
 
         # the number traits should be between 3 and 8
         self.assertTrue(3 <= num.number <= 8)
 
-    @requires_toolkit([ToolkitName.qt])
-    def test_avoid_slider_feedback(self):
```
kitchoi, line 83: For future reference, I believe this is removed because the same scenario is now tested in test_range_editor.py

traitsui/tests/editors/test_range_editor.py
```diff
@@ -49,3 +49,70 @@ def test_simple_editor_format_func(self):
...
+
+    def test_log_range_slider_editor_set_with_text(self):
+        return self.check_set_with_text(mode='logslider')
+
+    def test_range_text_editor_set_with_text(self):
+        return self.check_set_with_text(mode='text')
+
+    # There is a problem with KeySequence on wx.  trying to include the key
+    # '\b' does not succesfully type a backspace.  Instead EmulateKeyPress
+    # seems to literally type "\x08" which leads to errors.
```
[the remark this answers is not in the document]
aaronayres35: That makes sense so me, but would we then need to hard code all other escape characters?  for example '\b', '\t', '\n', etc.  That may be just what we need to do and that is perfectly fine.  I had just left the tests separate for now. so that the issue was known and we could try to address it.  It seems very likely to me that this may be the only solution, but I didn't know if there may would be a way to get the escape characters to work directly without explicitly checking and swapping them for the pyface key names.

traitsui/testing/tester/qt4/located_object_handlers.py
```diff
@@ -0,0 +1,57 @@
...
+#  All rights reserved.
+#
+#  This software is provided without warranty under the terms of the BSD
+#  license included in LICENSE.txt and may be redistributed only
+#  under the conditions described in the aforementioned license.  The license
+#  is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+#  Thanks for using Enthought open source!
+#
+
```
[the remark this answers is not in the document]
aaronayres35: I think that makes sense to me, I like the rename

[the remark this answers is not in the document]
aaronayres35: I think the name should reflect that these aren't the only available targets though, and that they are meant to be general/common to come up.  Thoughts on the name general_ui_targets or common_ui_targets?

traitsui/tests/editors/test_range_editor.py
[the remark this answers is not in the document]
kitchoi: There are only a few of these characters so I think it is not a big deal: We can maintain this as a separate mapping. Alternatively we can forbid the use of `
\b` in KeySequence and force test code to use KeyClick("Backspace"), but we should choose one route, not both.

traitsui/testing/tester/qt4/located_object_handlers.py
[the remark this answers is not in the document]
kitchoi: Hmmm, common_ui_targets?

traitsui/tests/editors/test_range_editor.py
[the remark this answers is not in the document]
rahulporuri: `KeyClick("Backspace")` with an optional argument for the # of backspaces sounds a lot better than manually doing `\b`.

traitsui/testing/tester/qt4/implementation/range_editor.py
```diff
@@ -0,0 +1,87 @@
...
+    ----------
+    wrapper : UIWrapper
+        Wrapper containing the Range Editor target.
+    location : locator.WidgetType.textbox
+        The location we are looking to resolve.
+    """
+    if location == locator.WidgetType.textbox:
+        return LocatedTextbox(textbox=wrapper.target.control.text)
+
+    raise NotImplementedError()
```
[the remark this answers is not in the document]
aaronayres35: That makes sense, I think it doesn't hurt to include the `NotImplementedError`.  Only having the check for a slider seems quite specific though.  TBH I'm not entirely sure what all the widget types would be - but we could simply do

```if location in [locator.WidgetType.slider, locator.WidgetType.tabbar, etc.]:
    raise NotImplementedError(f"Logic for interacting with the {location} has not been implemented.")```

File: traitsui/qt4/range_editor.py [hunk 7]
```diff
@@ -449,6 +449,9 @@ def update_object_on_scroll(self, pos):
     def update_object_on_enter(self):
         """ Handles the user pressing the Enter key in the text field.
         """
+        # it is possible we get the event after the control has gone away
```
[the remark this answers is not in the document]
aaronayres35: Sure thing, I copy pasted this comment from line 180, shall I replace it there as well?

traitsui/tests/editors/test_range_editor.py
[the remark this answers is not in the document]
kitchoi: OK, then we would have to check for `\b` in the Qt implementation to disallow it. Otherwise test code trying to use `\b` would not be toolkit agnostic. In either case, these escape characters need to be handled one way or another.

traitsui/testing/tester/qt4/implementation/range_editor.py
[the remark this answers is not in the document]
aaronayres35: I currently just made it a singleton list `[locator.WidgetType.slider]` as I wasn't sure what all widgets would be possible.  This can easily be added to.

traitsui/tests/editors/test_range_editor.py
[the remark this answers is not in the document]
kitchoi: Just to clarify, I think either way is fine.

traitsui/testing/tester/locator.py
```diff
@@ -35,3 +37,15 @@ class TargetByName:
...
     def __init__(self, name):
         self.name = name
+
+
+class WidgetType(enum.Enum):
+    """ A locator for locating nested widgets within a UI. Many editors will
+    contain many sub-widgets (e.g. a textbox, slider, tabs, buttons, etc.).
+
+    For example when working with a range editor, one could call
+    tester.find_by_name(ui, "ranged_number").locate(locator.WidgetType.textbox)
```
kitchoi, line 47: I think we need double backticks here for it to be rendered as code in the API documentation.
```
  ``tester.find_by_name(ui, "ranged_number").locate(locator.WidgetType.textbox)``
```

traitsui/testing/tester/qt4/implementation/range_editor.py
```diff
@@ -0,0 +1,101 @@
...
+        The location we are looking to resolve.
+    """
+
+    if location == locator.WidgetType.textbox:
+        return LocatedTextbox(textbox=wrapper.target.control)
+    if location in [locator.WidgetType.slider]:
+        raise NotImplementedError(
+            f"Logic for interacting with the {location}"
+            " has not been implemented."
+        )
```
kitchoi, line 69: I think the RangeTextEditor only has a textbox, i.e. no slider. So this block can be removed.

traitsui/testing/tester/wx/implementation/range_editor.py
```diff
@@ -0,0 +1,99 @@
...
+        The location we are looking to resolve.
+    """
+
+    if location == locator.WidgetType.textbox:
+        return LocatedTextbox(textbox=wrapper.target.control)
+    if location in [locator.WidgetType.slider]:
+        raise NotImplementedError(
+            f"Logic for interacting with the {location}"
+            " has not been implemented."
+        )
```
kitchoi, line 68: Same here, I think the RangeTextEditor is meant to only have a textbox and no slider.

traitsui/tests/editors/test_range_editor.py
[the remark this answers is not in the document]
aaronayres35: > `KeyClick("Backspace")` with an optional argument for the # of backspaces sounds a lot better than manually doing `\b`.

It makes sense to me to support both.  My understanding is just that we want the code to be toolkit agnostic.  I added kits suggestion for now so that '\b' works on both wx and qt.  Maybe in a later pr we can add to the KeyClick functions to take an optional argument to repeat the specified character n times.

traitsui/testing/tester/wx/helpers.py
```diff
@@ -133,4 +140,6 @@ def key_sequence_text_ctrl(control, interaction, delay):
     if not control.HasFocus():
         control.SetFocus()
     for char in interaction.sequence:
+        if char == '\b':
+            char = "Backspace"
```
kitchoi, line 144: I did not investigate why previously `\b` did not work. Now I tried... if I changed the `SetUnicodeKey` to `SetKeyCode` in `key_click` then we don't need this special casing of `\b`. Could you confirm please?

kitchoi (commented):
Now I looked a bit more into the `"\b"` business, looks like wx is actually happy with it.

traitsui/testing/tester/wx/helpers.py
aaronayres35: Ah nice, yes this works for me as well.  Do you understand why?  I was under the impression '\b' would work as a unicode character.

traitsui/testing/tester/wx/helpers.py
```diff
@@ -43,7 +43,7 @@ def key_click(widget, key, delay=0):
         else:
             wx.MilliSleep(delay)
             key_event = wx.KeyEvent(wx.wxEVT_CHAR)
-            key_event.SetUnicodeKey(KEY)
+            key_event.SetKeyCode(KEY)
```
kitchoi, line 46: I think this results in another test to fail:
```
FAIL: test_key_click (traitsui.testing.tester.wx.tests.test_helpers.TestInteractions)
  File "/Users/kchoi/Work/ETS/traitsui-clone/traitsui/testing/tester/wx/tests/test_helpers.py", line 96, in test_key_click
    self.assertEqual(textbox.Value, "A")
AssertionError: 'a' != 'A'
- a
+ A
```

I think the test failure makes sense because to type "A", one needs to have the cap on or have used the Shift key. Apparently Qt disagrees and would happily insert a capitalized `"A"` in that case. So the two disagree and we should normalize it. 😂 
That's orthogonal though we should open a separate issue to solve.

kitchoi: Once #1176 is merged and if we re-run the CI for this PR we should see this test failure.

aaronayres35: Should I modify the test to pass and comment on the discrepancy?  I can open a separate issue about it

kitchoi: I realized that the tests with wx and key clicks are actually failing on Windows.... this is getting orthogonal. Let's revert this change to `SetKeyCode` and modify the tests for RangeEditor here so that it does not rely on backspaces, then open a separate issue to fix the discrepancies.

kitchoi (approved):
LGTM!

kitchoi:
Sigh, we did this again... It is hard to see new test failures with wx CI job being allowed to fail. Here are the new errors on Windows:
```
FAIL: test_large_range_slider_editor_set_with_text (traitsui.tests.editors.test_range_editor.TestRangeEditor)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_range_editor.py", line 77, in test_large_range_slider_editor_set_with_text
    return self.check_set_with_text(mode='xslider')
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_range_editor.py", line 70, in check_set_with_text
    self.assertEqual(model.value, 4)
AssertionError: 1 != 4
FAIL: test_log_range_slider_editor_set_with_text (traitsui.tests.editors.test_range_editor.TestRangeEditor)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_range_editor.py", line 80, in test_log_range_slider_editor_set_with_text
    return self.check_set_with_text(mode='logslider')
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_range_editor.py", line 70, in check_set_with_text
    self.assertEqual(model.value, 4)
AssertionError: 1 != 4
FAIL: test_range_text_editor_set_with_text (traitsui.tests.editors.test_range_editor.TestRangeEditor)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_range_editor.py", line 83, in test_range_text_editor_set_with_text
    return self.check_set_with_text(mode='text')
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_range_editor.py", line 70, in check_set_with_text
    self.assertEqual(model.value, 4)
AssertionError: 1 != 4
FAIL: test_simple_slider_editor_set_with_text (traitsui.tests.editors.test_range_editor.TestRangeEditor)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_range_editor.py", line 74, in test_simple_slider_editor_set_with_text
    return self.check_set_with_text(mode='slider')
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_range_editor.py", line 70, in check_set_with_text
    self.assertEqual(model.value, 4)
AssertionError: 1 != 4

```

aaronayres35:
> Sigh, we did this again... It is hard to see new test failures with wx CI job being allowed to fail. Here are the new errors on Windows:

:/ At least it seems to be the effectively same issue #1177.

PR #1174, opened 2020-09-01
Clean super usage in etsdemo

rahulporuri:
This PR updates `super` usage in the `etsdemo` package but not in `traitsui` because `etsdemo` is small and new. This was done using manual search and replace.


kitchoi (approved):
LGTM

PR #1172, opened 2020-08-31
Ui tester api updates5 - Implement locating a nested UI for a custom style ListEditor

aaronayres35:
Implements the seventh bullet on issue #1149  (Implement locating a nested UI for a custom style ListEditor. Consequently, one should be able to navigate into nested editors inside a nested UI.)

Note that PR #1171 needs to be merged first.

PR #1175, opened 2020-09-01
UI Tester update - Implement locating a nested UI for a custom style ListEditor

aaronayres35:
Implements the seventh bullet on issue #1149 (Implement locating a nested UI for a custom style ListEditor. Consequently, one should be able to navigate into nested editors inside a nested UI.)

This PR replaces #1172

### DIFF
--- a/traitsui/testing/tester/qt4/implementation/list_editor.py
+++ b/traitsui/testing/tester/qt4/implementation/list_editor.py
@@ -9,23 +9,11 @@
 #  Thanks for using Enthought open source!
 #
 from traitsui.testing.tester import locator
+from traitsui.testing.tester.registry_helper import find_by_name_in_nested_ui
 
 from traitsui.qt4.list_editor import (
     CustomEditor,
 )
-
-
-def find_by_name_in_nested_ui(wrapper, location):
-    """ Helper function for resolving from an _IndexedCustomEditor to a
-    TargetByName.
-
-    Parameters
-    ----------
-    wrapper : UIWrapper
-    location : instance of locator.TargetByName
-    """
-    new_interactor = wrapper.locate(locator.NestedUI())
-    return new_interactor.find_by_name(location.name).target
 
 
 class _IndexedCustomEditor:
--- a/traitsui/testing/tester/qt4/implementation/list_editor.py
+++ b/traitsui/testing/tester/qt4/implementation/list_editor.py
@@ -16,6 +16,14 @@ from traitsui.qt4.list_editor import (
 
 
 def find_by_name_in_nested_ui(wrapper, location):
+    """ Helper function for resolving from an _IndexedCustomEditor to a
+    TargetByName.
+
+    Parameters
+    ----------
+    wrapper : UIWrapper
+    location : instance of locator.TargetByName
+    """
     new_interactor = wrapper.locate(locator.NestedUI())
     return new_interactor.find_by_name(location.name).target
 
@@ -39,6 +47,12 @@ class _IndexedCustomEditor:
     @classmethod
     def from_location(cls, wrapper, location):
         """ Helper method to create an _IndexedCustomEditor instance.
+
+        Parameters
+        ----------
+        wrapper : UIWrapper
+            UI Wrapper wrapping the Custom List Editor
+        location : instance of locator.Index
         """
         return cls(
             target=wrapper.target,
--- a/traitsui/testing/tester/qt4/implementation/list_editor.py
+++ b/traitsui/testing/tester/qt4/implementation/list_editor.py
@@ -1,0 +1,107 @@
+#  Copyright (c) 2005-2020, Enthought, Inc.
+#  All rights reserved.
+#
+#  This software is provided without warranty under the terms of the BSD
+#  license included in LICENSE.txt and may be redistributed only
+#  under the conditions described in the aforementioned license.  The license
+#  is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+#  Thanks for using Enthought open source!
+#
+from traitsui.testing.tester import locator
+
+from traitsui.qt4.list_editor import (
+    CustomEditor,
+)
+
+
+def find_by_name_in_nested_ui(wrapper, location):
+    new_interactor = wrapper.locate(locator.NestedUI())
+    return new_interactor.find_by_name(location.name).target
+
+
+class _IndexedCustomEditor:
+    """ Wrapper for a ListEditor (custom) with an index.
+    """
+
+    def __init__(self, target, index):
+        """
+        Parameters
+        ----------
+        target : CustomEditor
+            The Custom List Editor
+        index : int
+            The index of interest.
+        """
+        self.target = target
+        self.index = index
+
+    @classmethod
+    def from_location(cls, wrapper, location):
+        """ Helper method to create an _IndexedCustomEditor instance.
+        """
+        return cls(
+            target=wrapper.target,
+            index=location.index,
+        )
+
+    @classmethod
+    def register(cls, registry):
+        """ Class method to register interactions on a _IndexedCustomEditor
+        for the given registry.
+
+        If there are any conflicts, an error will occur.
+
+        Parameters
+        ----------
+        registry : TargetRegistry
+            The registry being registered to.
+        """
+        registry.register_solver(
+            target_class=CustomEditor,
+            locator_class=locator.Index,
+            solver=cls.from_location,
+        )
+        registry.register_solver(
+            target_class=cls,
+            locator_class=locator.NestedUI,
+            solver=lambda wrapper, _: wrapper.target._get_nested_ui(),
+        )
+
+        registry.register_solver(
+            target_class=cls,
+            locator_class=locator.TargetByName,
+            solver=find_by_name_in_nested_ui,
+        )
+
+    def _get_nested_ui(self):
+        """ Method to get the nested ui corresponding to the List element at
+        the given index.
+        """
+        row, column = divmod(self.index, self.target.factory.columns)
+        # there are two columns for each list item (one for the item itself,
+        # and another for the list menu button)
+        column = 2*column
+        grid_layout = self.target._list_pane.layout()
+        item = grid_layout.itemAtPosition(row, column)
+        if item is None:
+            raise IndexError(self.index)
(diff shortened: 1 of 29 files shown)


traitsui/tests/editors/test_list_editor.py
```diff
@@ -0,0 +1,110 @@
...
+        # getting the correct nested ui 
+        for col in range(1,5):
+            obj = ListTraitTest(people=copy.deepcopy(people), num_columns=col)
+            tester = UITester()
+            with tester.create_ui(obj) as ui:
+                # sanity check
+                self.assertEqual(obj.people[7].name, "Fields")
+                item = tester.find_by_name(ui, "people").locate(locator.Index(7))
+                name_field = item.find_by_name("name")
+                name_field.perform(command.KeySequence("\b\b\b\b\b\bDavid"))
```
rahulporuri, line 78: instead of `"\b\b\b..."` etc, wouldn't it be better to do `"\b"*len("Fields")`?

traitsui/tests/editors/test_list_editor.py
```diff
@@ -0,0 +1,110 @@
...
+                displayed = name_field.inspect(query.DisplayedText())
+                self.assertEqual(obj.people[7].name, "David")
+                self.assertEqual(displayed, obj.people[7].name)
+
+    # The above test should in theory work for wx, but the tester currently 
+    # has a bug for typing "\b" with wx (it types a literal "\x08" to the
+    # textbox)
```
rahulporuri, line 85: does `\b` need to be a raw string `"r'\b'"`

aaronayres35: This doesn't seem to work either, but with the change from the previous PR #1171 just "\b" will work

traitsui/tests/editors/test_list_editor.py
aaronayres35: This does make the reason for the specific number of "\b"s more clear.  However `"\b"*len("Fields") + "David"` no longer feels like a key sequence when you have to like parse the expression mentally.  I don't have a strong preference either way

traitsui/testing/tester/qt4/implementation/list_editor.py
```diff
@@ -0,0 +1,121 @@
...
+    """ Helper function for resolving from an _IndexedCustomEditor to a
+    TargetByName.
+
+    Parameters
+    ----------
+    wrapper : UIWrapper
+    location : instance of locator.TargetByName
+    """
+    new_interactor = wrapper.locate(locator.NestedUI())
+    return new_interactor.find_by_name(location.name).target
```
aaronayres35, line 28: This could probably be pulled out into a separate module eventually like what was done in the draft branch.  I didn't yet as for this pr it felts unnecessary to create a full separate module to house just this function (but I certainly can change that if needed).

traitsui/testing/tester/qt4/implementation/list_editor.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class _IndexedCustomEditor:
```
aaronayres35, line 31: I have been thinking about how to modify this so that potentially repeated code doesn't need to be repeated.  For example trying to extract some skeleton of this class and keep it in the common_ui_targets module, but I could not decide on a clean way to do so.  The specifics of accessing the nested UI will vary from editor to editor

traitsui/tests/editors/test_list_editor.py
```diff
@@ -0,0 +1,89 @@
...
+        for col in range(1,5):
+            obj = ListTraitTest(people=copy.deepcopy(people), num_columns=col)
+            tester = UITester()
+            with tester.create_ui(obj) as ui:
+                # sanity check
+                self.assertEqual(obj.people[7].name, "Fields")
+                item = tester.find_by_name(ui, "people").locate(locator.Index(7))
+                for _ in range(6):
+                    item.find_by_name("name").perform(command.KeyClick("Backspace"))
+                item.find_by_name("name").perform(command.KeySequence("David"))
```
aaronayres35, line 78: Eventually we will be able to do this with just one KeySequence using '\b' characters, but there is a known issue with that so this works for now.

traitsui/testing/tester/qt4/implementation/list_editor.py
kitchoi: I agree this should be pulled out because this logic is actually toolkit agnostic: It deals with the `UI` object. Furthermore the code is repeated - it is a true duplication in this case.
Size of a module is not an issue.

aaronayres35:
This also gives a wx failure on windows:
```
890FAIL: test_locate_element_and_edit (traitsui.tests.editors.test_list_editor.TestCustomListEditor)
891----------------------------------------------------------------------
892Traceback (most recent call last):
893  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_list_editor.py", line 84, in test_locate_element_and_edit
894    self.assertEqual(obj.people[7].name, "David")
895AssertionError: 'Fields' != 'David'
896- Fields
897+ David
898
899
```

like issue #1177 
All failures appear to be coming from the same source (I suspect `EmulateKeyPress` in wx helpers.)

traitsui/tests/editors/test_list_editor.py
```diff
@@ -0,0 +1,94 @@
...
+            tester = UITester()
+            with tester.create_ui(obj) as ui:
+                # sanity check
+                self.assertEqual(obj.people[7].name, "Fields")
+                people_list = tester.find_by_name(ui, "people")
+                item = people_list.locate(locator.Index(7))
+                name_field = item.find_by_name("name")
+                for _ in range(6):
+                    name_field.perform(command.KeyClick("Backspace"))
+                name_field.perform(command.KeySequence("David"))
```
kitchoi, line 82: Minor: I'd have wanted to keep these two were it not because of the ongoing issue with KeyClick... Perhaps for the purpose of this PR we can just keep `DisplayedText()` in the next line. That would be enough to verify we have got the right entry.

traitsui/tests/editors/test_list_editor.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@requires_toolkit([ToolkitName.qt, ToolkitName.wx])
+class TestCustomListEditor(unittest.TestCase):
+
+    def test_locate_element_and_edit(self):
+        # varying the number of columns in the view tests the logic for
+        # getting the correct nested ui
+        for col in range(1, 5):
+            obj = ListTraitTest(people=copy.deepcopy(people), num_columns=col)
```
kitchoi, line 72: Minor: Instead of deepcopy, we can wrap the global list of Person in a function so we always get a new list.
deepcopy does work...but it involves more complexity/logic than necessary.

traitsui/tests/editors/test_list_editor.py
```diff
@@ -0,0 +1,94 @@
...
+                people_list = tester.find_by_name(ui, "people")
+                item = people_list.locate(locator.Index(7))
+                name_field = item.find_by_name("name")
+                for _ in range(6):
+                    name_field.perform(command.KeyClick("Backspace"))
+                name_field.perform(command.KeySequence("David"))
+                displayed = name_field.inspect(query.DisplayedText())
+                self.assertEqual(obj.people[7].name, "David")
+                self.assertEqual(displayed, obj.people[7].name)
+
```
kitchoi, line 86: Could we have a test to assert the IndexError if we attempt to get an index that is out-of-range?
I believe the error occurs when we call `item.find_by_name`, not when we call `locate(Index(...))`. While it is slightly counterintuitive (one might expect `locate` to raise instead), I think that is fine. The argument for raising later is that the GUI might have changed state between calling `locate` and calling `find_by_name`. So even if `locate` succeeds, that does not guarantee `find_by_name` will succeed.

traitsui/testing/tester/qt4/implementation/list_editor.py
kitchoi: Sounds reasonable to me!

kitchoi (commented):
LGTM. Some minor suggestions.

traitsui/testing/tester/qt4/implementation/list_editor.py
```diff
@@ -0,0 +1,109 @@
...
+        Parameters
+        ----------
+        wrapper : UIWrapper
+            UI Wrapper wrapping the Custom List Editor
+        location : instance of locator.Index
+        """
+        return cls(
+            target=wrapper.target,
+            index=location.index,
+        )
```
aaronayres35, line 48: I feel like this method is unnecessary and just makes code harder to read.  I think we might as well just do `solver=lambda wrapper, location: cls(target=wrapper.target, index=location.index)` in line 65

traitsui/tests/editors/test_list_editor.py
aaronayres35: once #1184 is merged, this test should pass.  Will wait to merge this PR until after that has been merged.

aaronayres35: > once #1184 is merged, this test should pass. Will wait to merge this PR until after that has been merged.

It is possible that it would still if the insertion points start incorrectly.  1184 only made the change in default for range editor.  I could see an issue where the backspaces occur before Fields and we get an output of "FieldsDavid".  I am checking this now in a draft PR on a junk branch that I will delete after the CI is finished.  If a fix is needed for the insertion point stuff I will add it to this PR

aaronayres35:
My suspicions in the above discussion were correct.  I added a helper function to set the insertion point the the right not the left end of a textbox on wx.  This feels a bit tangential to this PR ( I can make this change in a separate PR if needed) but without it, this PR will introduce a new test failure.  After #1184 is merged, with this most recent change, there shouldn't be any new errors with this PR as far as I know.

traitsui/tests/editors/test_list_editor.py
kitchoi: We don't know how long it'd take for #1184 to be merged. I'd still prefer decoupling this PR from #1184 so we can review and merge it more efficiently. But if you'd like to wait, that's okay for me.

kitchoi (approved):
LGTM.
One function being overly indented, can be merged after fixing that.

PR #1181, opened 2020-09-02
Normalize KeySequence support for testing textboxes

kitchoi:
Currently the `KeySequence` accepts special keys such as "\b", "\t" for Qt. Under the hood, Qt converts these characters to Qt Key so the support is relatively painless. Similar support for wx.TextCtrl requires a bit more custom work. (See thread starting from https://github.com/enthought/traitsui/pull/1171#discussion_r481055086).

On the other hand, wx.TextCtrl supports some other unicode code points that QTest does not support out-of-the-box (e.g. `'\x1f'` == `chr(31)`).
For testing textboxes, it is probably not necessary to support all possible keys. It may be easier to start with a more limited support and expand it when there are new needs.

This PR
- Restrict supported characters in KeySequence only for the context of testing textbox. Tests all these characters.
      - KeySequence("\b") would not be supported for textboxes. KeyClick("Backspace") should be used instead.
- Somewhat orthogonal, happy to pull out to a separate PR with a separate issue: This PR also normalizes the behaviour for when the textbox already has text. For QLineEdit, we append. For QTextEdit, we insert at the beginning. For wx, we insert text at the beginning too. Here I took the opportunity to normalize these such that the insertion point is at the end of the text if the textbox does not already have focus.

### DIFF
--- a/traitsui/testing/tester/qt4/helpers.py
+++ b/traitsui/testing/tester/qt4/helpers.py
@@ -12,6 +12,7 @@
 from pyface.qt import QtCore, QtGui
 from pyface.qt.QtTest import QTest
 
+from traitsui.testing.tester.compat import check_key_compat
 from traitsui.testing.tester.exceptions import Disabled
 from traitsui.qt4.key_event_to_name import key_map as _KEY_MAP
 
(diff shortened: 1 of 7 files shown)


traitsui/testing/tester/qt4/helpers.py
```diff
@@ -110,6 +111,30 @@ def key_sequence_qwidget(control, interaction, delay):
...
+        e.g. QLineEdit and QTextEdit
+    interaction : instance of command.KeySequence
+        The interaction object holding the sequence of key inputs
+        to be simulated being typed
+    delay : int
+        Time delay (in ms) in which each key click in the sequence will be
+        performed.
+    """
+    for key in interaction.sequence:
+        check_key_compat(key)
```
kitchoi, line 132: For future reference, if we did not check and try to use QTest with a character e.g. "ಶ", then Qt will throw this error:
```
ASSERT: "false" in file /Users/builder/workspace/Buildsystem/test/pisi/tmp/Qt-5.12.6-5/work/qt/qtbase/src/testlib/qasciikey.cpp, line 228
Abort trap: 6
```
If there are needs to support a wider range of characters, then we would need to do something else. Depending on the use case, we may manually edit the text and fire the appropriate Qt event used by the editor.

kitchoi: The test `test_key_sequence_unsupported_key` demonstrates the above error if the check was not added.

PR #1190, opened 2020-09-03
Refactor registering interaction handlers for editable textboxes

kitchoi:
This PR refactors how we add interaction handlers for editable textboxes.

There are small changes to the tests to exercise more interactions in order to ensure the refactoring has not broken things.

With this refactoring, I can add interactions for things like the text-style FontEditor relatively easily (see #1161).

### DIFF
--- a/traitsui/testing/tester/qt4/common_ui_targets.py
+++ b/traitsui/testing/tester/qt4/common_ui_targets.py
@@ -44,22 +44,44 @@ class LocatedTextbox:
         registry : TargetRegistry
             The registry being registered to.
         """
-        handlers = [
-            (command.KeySequence,
-                (lambda wrapper, interaction: helpers.key_sequence_textbox(
-                    wrapper.target.textbox, interaction, wrapper.delay))),
-            (command.KeyClick,
-                (lambda wrapper, interaction: helpers.key_click_qwidget(
-                    wrapper.target.textbox, interaction, wrapper.delay))),
-            (command.MouseClick,
-                (lambda wrapper, _: helpers.mouse_click_qwidget(
-                    wrapper.target.textbox, wrapper.delay))),
-            (query.DisplayedText,
-                lambda wrapper, _: wrapper.target.textbox.displayText()),
-        ]
-        for interaction_class, handler in handlers:
-            registry.register_handler(
-                target_class=cls,
-                interaction_class=interaction_class,
-                handler=handler,
-            )
+        register_editable_textbox_handlers(
+            registry=registry,
+            target_class=cls,
+            widget_getter=lambda wrapper: wrapper.target.textbox,
+        )
+
+
+def register_editable_textbox_handlers(registry, target_class, widget_getter):
+    """ Register common interactions for an editable textbox (in Qt)
+
+    Parameters
+    ----------
+    registry : TargetRegistry
+        The registry being registered to.
+    target_class : subclass of type
+        The type of target being wrapped in a UIWrapper on which the
+        interaction will be performed.
+    widget_getter : callable(wrapper: UIWrapper) -> QWidget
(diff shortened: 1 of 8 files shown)


kitchoi:
Ooooo that's interesting, this error is only seen on Windows:
```
FAIL: test_simple_auto_set_update_text (traitsui.tests.editors.test_text_editor.TestTextEditor)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_text_editor.py", line 173, in test_simple_auto_set_update_text
    name_field.perform(command.KeySequence("NEW"))
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traits\testing\unittest_tools.py", line 119, in __exit__
    raise self.failureException(msg.format(*items))
AssertionError: Change event for name was fired 3 times instead of 4
```
Looks orthogonal. I will open a separate issue and revert the change to the test for now.

kitchoi:
> this error is only seen on Windows:

That's misleading: It is not seen on OSX but I don't know if the error should occur on Linux because our Linux+wx CI job has not been running.

kitchoi:
> Looks orthogonal. I will open a separate issue and revert the change to the test for now.

It is because on Windows + wx, the first input is always being activated and in the case of textbox, the entire text is selected. This makes the test setup different from Mac OS or with Qt. Therefore this is orthogonal to the refactoring.

traitsui/testing/tester/qt4/common_ui_targets.py
```diff
@@ -44,22 +44,44 @@ def register(cls, registry):
...
-                handler=handler,
-            )
+        register_editable_textbox_handlers(
+            registry=registry,
+            target_class=cls,
+            widget_getter=lambda wrapper: wrapper.target.textbox,
+        )
+
+
+def register_editable_textbox_handlers(registry, target_class, widget_getter):
```
aaronayres35, line 54: I very much like this refactor, pulling this logic into its own function is much cleaner.  However, do you think we should move this into the helpers module, or somewhere else?  It seems a little out of place in common_ui_targets IMO, what do you think?

aaronayres35 (approved):
Changes LGTM!  Just one question about where to place the new function

traitsui/testing/tester/qt4/common_ui_targets.py
kitchoi: I agree this function does not sit well with this module. 

But I am a bit hesitant about moving it to `helpers`. The only UITester/UIWrapper piece of knowledge in that module is the use of `interaction` object, and in the places where `interaction` is referred to, they could be replaced with more primitive data such as a string. Perhaps `helpers` should be renamed to something more meaningful to prevent it from becoming a catch-all-I-don't-know-where-this-goes kind of place.

What about... we move this `register_editable_textbox_handlers` to a new module called `registry_helper`? This will mirror the one on `tester` package and is responsible for holding functions that does some common handler registration.

aaronayres35: That sounds good to me!

aaronayres35: Also, I agree that `helpers` might be a bit too generic (all code we write should be helping with something!).  Currently all the code in `helpers` is specific to some form of interaction (keyclicks, mouse clicks, key sequences, and DisplayedText queries) so maybe just something like  `interaction_helpers` or something would be sufficient.  We can leave that for another PR

kitchoi: Cool. I will move this function to a new module.
Agreed the renaming of `helpers` can be a separate PR.

aaronayres35 (approved):
Changes still LGTM!

Issue #1173, opened 2020-09-01
Pre-release checklist for UITester


kitchoi:
Currently the UITester is being developed and tested with TraitsUI's own test suite. Eventually when the API is mature enough, it will need to be made public. There are a number of tasks that need to be evaluated before that. Collecting them here as a reminder and for discussion. Feel free to expand and open separate issues for them if needed.

- [x] Evaluate naming of modules / classes / functions
       - Rename modules (#1248)
       - Rename `UIWrapper.target` (#1249)
- [x] Add `traitsui.testing.api` to expose API components. This also serves as a facade. (#1269)
- [x] Documentation in the user manual with an example for writing an integration test (#1223)
- [x] Move pushing trait change exception handler out. (#1232)
- [x] Update demo examples to import from `traitsui.testing.api` instead (#1274)
- [x] Rename TargetRegister.register_handler and register_solver #1273

rahulporuri:
Regarding naming of modules, do we really need the additional `implementation` submodule under qt4/wx in the tester?

- https://github.com/enthought/traitsui/tree/master/traitsui/testing/tester/wx/implementation
- https://github.com/enthought/traitsui/tree/master/traitsui/testing/tester/qt4/implementation

kitchoi:
We could rename `implementation` to `editors`. I think everything in there are editor domain specific, and everything outside but still within the `qt4` / `wx` namespace are not editor specific but still toolkit specific. Then again, we could also remove that separation. Whichever way to help understanding the structure and to help making the location of new code obvious is good.

kitchoi:
Another renaming to be considered:
- `TargetRegistry.register_solver` -> `TargetRegistry.register_location`
- `TargetRegistry.register_handler` -> `TargetRegistry.register_interaction`

aaronayres35:
The last bullet was completed by PR #1275.  Closing now

PR #1191, opened 2020-09-03
Ui tester updates - add enum editor implementation

### LINKED ISSUE #1187: Implement Enum Editor for UI Tester
There should be implementations for both modes ('radio', and 'list') of a custom style editor, and also for a simple style editor.  
This will require adding the ability to click on a combobox as well as on a wxListBox/QtGui.QListWidget 
(Implementations to do so will like provide more general clicking ability for other widgets - see qt implementation on #1150 which utilizes a `mouse_click_item_view` function).
Also, the Simple style editor allows text entry in the combo box.

aaronayres35:
fixes #1187


traitsui/testing/tester/wx/implementation/enum_editor.py
```diff
@@ -0,0 +1,96 @@
...
+                target_class=cls,
+                interaction_class=interaction_class,
+                handler=handler
+            )
+        for location_class, solver in cls.solvers:
+            registry.register_solver(
+                target_class=cls,
+                interaction_class=location_class,
+                solver=solver,
+            )
```
aaronayres35, line 51: I realized the code for each of the classes here was nearly identical so I refactored to this.  This class has nothing to do with Enum editor specifically though, and should be pulled out into a separate module.  It could easily be used by CheckListEditor as well for example.  I am not exactly sure where to put it though.  It doesn't belong in common_ui_targets as it is not intended to be a target.  It is meant to be subclassed.  In fact it is also not even toolkit specific and could easily be used by qt Enum editor and Check list editor as well.

@kitchoi what do you think about having this as a skeleton class?  Do you think it is beneficial/if so where do you think it should be placed?

traitsui/testing/tester/base_classes.py
```diff
@@ -0,0 +1,44 @@
+#  Copyright (c) 2005-2020, Enthought, Inc.
```
aaronayres35, line 1: Naming things is hard... This module should be renamed, I just wanted to make it something for now.

kitchoi: Agreed this needs to be renamed. Let's see what we ended up with and then we can name it more appropriately before we merge this.

traitsui/testing/tester/base_classes.py
```diff
@@ -0,0 +1,44 @@
...
+class _IndexedEditor:
+    """ Wrapper for Editor and index """
+
+    target_class = None
+    handlers = []
+    solvers = []
+
+    def __init__(self, target, index):
+        self.target = target
+        self.index = index
```
kitchoi, line 23: I agree this refactoring is worth doing, and is a neat idea.

This is referring to the pattern where we choose not to resolve the given locator (e.g. `locator.Index`) until later, instead, we want to keep the target and the locator through.

This would be a pattern useful for things like TableEditor, where we have to keep the location information (e.g. row and column indices for a cell) and editor through until when we perform a mouse click or a key click etc.

This is a pattern different from things like RangeEditor, where we can resolve the textbox widget and throw the location information away.

So I am wondering, what do you think about further abstracting the `index` here to just `location`. Maybe we can rename `target` to `ancestor` or `source` (something like that).  Then this object is not at all "index" specific. We will need to add the `locator_class` as class attribute for subclass to set, but that's about it.

(Related: It might be good to rename `target` anyway so we won't be writing `wrapper.target.target`)

traitsui/testing/tester/qt4/implementation/enum_editor.py
```diff
@@ -0,0 +1,63 @@
...
+        )
+    ]
+
+
+class _IndexedSimpleEditor(_IndexedEditor):
+    target_class = SimpleEditor
+    handlers = [
+        (command.MouseClick, (lambda wrapper, _: helpers.mouse_click_combobox(
+            combobox=wrapper.target.target.control,
+            index=wrapper.target.index,
```
kitchoi, line 48: This `target` is an instance of `_IndexedSimpleEditor` but the `_IndexedSimpleEditor.target_class` is a `SimpleEditor`... the word `target` is overloaded in the same place so this looks a bit confusing. See comment above regarding renaming `target` on `_IndexedEditor`. Say if we rename `_IndexedEditor.target` to `_IndexedEditor.source`, then `_IndexedEditor.target_class` should be renamed to `_IndexedEditor.source_class`.

kitchoi (commented):
Refactoring to a base class makes sense to me!

traitsui/testing/tester/base_classes.py
```diff
@@ -0,0 +1,50 @@
...
+#  under the conditions described in the aforementioned license.  The license
+#  is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+#  Thanks for using Enthought open source!
+#
+
+from traitsui.testing.tester import locator
+
+class _IndexedEditor:
+    """ Wrapper for Editor and index """
```
aaronayres35, line 15: Also needs to be renamed

traitsui/testing/tester/wx/implementation/enum_editor.py
```diff
@@ -0,0 +1,123 @@
...
+                interaction=interaction,
+                delay=wrapper.delay))
+        ),
+        (command.KeySequence, (lambda wrapper, interaction:
+            helpers.key_sequence_text_ctrl(
+                control=wrapper.target.control,
+                interaction=interaction,
+                delay=wrapper.delay))
+        ),
+    ]
```
aaronayres35, line 109: `key_click_text_ctrl` and `key_sequence_text_ctrl` are actually not at all `wxTextCtrl` specific any more after the removal of the `key_click` function and the use of `EmulateKeyPress`.  The only methods they use are just specific to `wxControl` and `wxTextEntry`.  In this case, `wvComboBox` subclasses both (just as `wxTextCtrl` does).  
Thus I think `key_click_text_ctrl` and `key_sequence_text_ctrl` can probably be renamed.  However, we may ultimately want a keyclick and key sequence function specific to text controls if we do ever get the old key_click function working which would allow things like '\b' to be typed, etc.

aaronayres35: Any changes related to this could probably go into a separate PR

aaronayres35:
There are new errors on wx in the test suite on OSX which I did not observe on my machine. looking into them now

kitchoi (commented):
Thank you!  The testing support for EnumEditor will be very useful!

Some suggestions... Since I will be away for some time, @rahulporuri would you be able to take over please?

traitsui/testing/tester/qt4/helpers.py
```diff
@@ -47,6 +47,28 @@ def key_click(widget, key, delay=0):
...
+        If the index is not valid.
+    """
+    if not index.isValid():
+        row = index.row()
+        column = index.column()
+        raise LookupError(
+            "Unabled to locate item with row {!r} and column {!r}.".format(
+                row, column,
+            )
+        )
```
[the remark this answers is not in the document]
kitchoi: Clarifications: The link was only meant to provide related references should that be useful, I don't mean we need all those tests. As long as the test support code is exercised somehow, I think we are good.

traitsui/testing/tester/wx/helpers.py
```diff
@@ -55,6 +55,83 @@ def mouse_click_checkbox(control, delay):
...
+    if not control.IsEnabled():
+        return
+    wx.MilliSleep(delay)
+    click_event = wx.CommandEvent(
+        wx.wxEVT_COMMAND_RADIOBUTTON_SELECTED, control.GetId()
+    )
+    click_event.SetEventObject(control)
+    control.SetValue(not control.GetValue())
+    control.ProcessWindowEvent(click_event)
+
```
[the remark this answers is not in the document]
aaronayres35: I do like this idea for refactor.  However, a refactor like this could impact effectively every keyclick function in helpers.  It may be better to make this change as its own PR entirely, and just leave the repetition for now.

traitsui/testing/tester/wx/helpers.py
```diff
@@ -134,6 +227,18 @@ def key_click_text_ctrl(control, interaction, delay):
...
+        a, b = control.GetTextSelection()
+        if a != b:
+            return (a, b)
+        else:
+            return (-1, -1)
+    else:  # control is wx.TextCtrl
+        if control.GetStringSelection():
+            return control.GetSelection()
+        else:
+            return (-1, -1)
```
aaronayres35, line 241: This fixes the mentioned test failures but it is a very ugly solution.  I could not decide on the best way to make this fix.  @rahulporuri maybe we could have a quick chat to discuss on Monday?

traitsui/testing/tester/qt4/tests/test_helpers.py
```diff
@@ -209,3 +209,15 @@ def test_key_click_disabled(self):
...
+        self.widget = QtGui.QListWidget()
+        self.items = ["a", "b", "c"]
+        self.widget.addItems(self.items)
+        self.good_q_index = self.widget.model().index(1, 0)
+        self.bad_q_index = self.widget.model().index(10, 0)
+
+        self.model = self.widget.model()
+        helpers.check_q_model_index_valid(self.good_q_index)
+        with self.assertRaises(LookupError):
+            helpers.check_q_model_index_valid(self.bad_q_index)
```
rahulporuri, line 223: missing empty line at the end of file.

traitsui/testing/tester/wx/helpers.py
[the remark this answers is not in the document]
aaronayres35: I have opened a separate issue for this (#1194)

traitsui/tests/editors/test_enum_editor.py
```diff
@@ -439,52 +306,45 @@ def test_radio_enum_editor_button_update(self):
...
 
             self.assertEqual(enum_edit.value, "one")
 
-            # The layout is: one, three, four \n two
-            click_radio_button(editor.control, 3)
-            process_cascade_events()
+            radio_editor = tester.find_by_name(ui, "value")
+            radio_editor.locate(locator.Index(3)).perform(command.MouseClick())
 
+            # The layout is: one, three, four \n two
```
aaronayres35, line 317: This is incorrect! Indexing should work in the same order as the enum. See issue #1204

traitsui/testing/tester/wx/helpers.py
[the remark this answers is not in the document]
kitchoi: I think we can limit the scope of this refactoring to only mouse_click_* function that uses this same pattern. In other words, it will affect the existing `mouse_click_checkbox` and no more. 

It is a judgment call whether one should split up refactoring into a separate PR, depending on the circumstances and the scope. A lot of times refactoring will be motivated by adding new code that repeats patterns in existing code, and the refactoring will change existing code. It would be acceptable to include related refactoring in a PR that adds features... unless when the scope got too big to be efficient. In this case, I'd be in favour of doing the refactoring here. Then again I might be imagining a specific scope that is different from how you understand it.

traitsui/testing/tester/wx/helpers.py
```diff
@@ -55,6 +55,83 @@ def mouse_click_checkbox(control, delay):
...
+    if isinstance(control, wx.ComboBox):
+        click_event = wx.CommandEvent(
+            wx.wxEVT_COMMAND_COMBOBOX_SELECTED, control.GetId()
+        )
+    elif isinstance(control, wx.Choice):
+        click_event = wx.CommandEvent(
+            wx.wxEVT_COMMAND_CHOICE_SELECTED, control.GetId()
+        )
+    else:
+        raise TypeError
```
kitchoi, line 83: I did not comment on this earlier thinking the refactoring suggested might make it irrelevant: if this stays, please instantiate the exception with an error message.

traitsui/testing/tester/wx/helpers.py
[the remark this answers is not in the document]
kitchoi: Consider this nominally resolved: I'd leave it to you to decide if the refactoring suggested was too much to fit here.

traitsui/testing/tester/wx/helpers.py
```diff
@@ -135,6 +227,43 @@ def key_click_text_ctrl(control, interaction, delay):
...
+    control : wxComboBox
+        The wx Object to be acted on.
+    interaction : instance of command.KeyClick
+        The interaction object holding the key input
+        to be simulated being typed
+    delay : int
+        Time delay (in ms) in which the key click will be performed.
+    """
+    if not control.IsEditable():
+        raise Disabled("{!r} is disabled.".format(control))
```
kitchoi, line 245: This does not work for combobox when the UI is disabled...
Looks like we should do this instead:
```
if not (control.IsEnabled() and control.IsEditable()):
```

This is the test I expect to pass but currently fails on wx (passes on Qt though):
```

    def test_simple_editor_disabled(self):
        enum_edit = EnumModel(value="two")
        view = View(
            UItem(
                "value",
                style="simple",
                enabled_when="value == 'one'",
                editor=EnumEditor(evaluate=True, values=["one", "two"]),
            ),
        )
        tester = UITester()
        with tester.create_ui(enum_edit, dict(view=view)) as ui:
            combobox = tester.find_by_name(ui, "value")
            with self.assertRaises(Disabled):
                combobox.perform(command.KeyClick("Enter"))

```

(this applied to KeySequence as well)

kitchoi (commented):
Thank you for the changes!  I am sorry for not having batched my earlier comments. Here are a few more. I think I am done for now.

traitsui/testing/tester/wx/implementation/enum_editor.py
```diff
@@ -0,0 +1,195 @@
...
+        i = new_index % (num_rows - 1)
+        j = new_index // (num_empty_entries_last_row)
+        # convert that back to an index found from row major order
+        return num_entries_grid1 + (i * num_empty_entries_last_row + j)
+    else:
+        # find i,j coordinates of index if we counted in column major order
+        i = index % num_rows
+        j = index // num_rows
+    # convert that back to an index found from row major order
+    return i * num_cols + j
```
aaronayres35, line 99: this logic is repeated on Qt as well.  I am thinking I might refactor by pulling this out into a helper function that could be kept in a `utils` module or something like that.  The code is a bit complicated, I tried to make comments to make it more clear.  The code for populating the QGridLayout and the wx.GridSizer is complicated itself.  (I find it strange that the for loops for populating them are based on going in row order, but the default/ standard behavior is to use column order.  This forces the use of the `incr` array and for the population to occur in a row major order, but with the column major ordered elements) see https://github.com/enthought/traitsui/blob/81902cc48f0af498d4f0f1a980dc8074da4d8265/traitsui/qt4/enum_editor.py#L334 and https://github.com/enthought/traitsui/blob/81902cc48f0af498d4f0f1a980dc8074da4d8265/traitsui/wx/enum_editor.py#L357

traitsui/testing/tester/wx/helpers.py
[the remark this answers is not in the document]
aaronayres35: > I think we can limit the scope of this refactoring to only mouse_click_* function that uses this same pattern. In other words, it will affect the existing `mouse_click_checkbox` and no more.

This is the only function with the exact same pattern, however there are many (basically all) mouse_click_* functions that have a very similar pattern.  They seemed similar enough to me that a refactor like this could be applied to many more of the mouseclick functions.  In that sense I think I was imagining a larger scope for this refactor.  
I spent a bit of time trying to work on this, and still have my scratch code saved for later, but decided to put it on pause and deal with later as the scope was kind of expanding as I was working on it (e.g. as I was adding it i realized if I just changed it slightly then such and such other mouse_click functions could also utilize it, etc.)

traitsui/tests/editors/test_enum_editor.py
```diff
@@ -521,3 +426,40 @@ def test_list_evaluate_editor_text(self):
...
+                    mode="list",
+                ),
+                style="custom",
+            ),
+            resizable=True,
+        )
+        self.check_list_enum_none_selected(view)
+
+    def test_list_evaluate_none_selected(self):
+        self.check_list_enum_none_selected(get_evaluate_view("custom", mode="list"))
```
aaronayres35, line 465: I am very confused why these fail.  The similar test passes for the radio editor.  Also, on the displayed list, visually none of the entries are selected.  However, the first assertions fail (i.e. `enum_edit.value ='one'` to start, not `None`).  It may be the case that the List style does not allow None selection but Radio does, but that feels like strange/ incorrect behavior to me.

aaronayres35: I may need to just raise an error in the line you mentioned (i ended up returning "" as it seemed like it was possible to have nothing selected, but I may have been mistaken for the list editor style)

aaronayres35: On Qt, the failing line is: https://github.com/enthought/traitsui/blob/81902cc48f0af498d4f0f1a980dc8074da4d8265/traitsui/qt4/enum_editor.py#L428

aaronayres35: I don't believe this is failing because of anything UI Tester related.  It does seem like a bug in traitsui to me though.  It would be perfectly fine to not allow none of the enum options to be selected, but the inconsistency across styles of editors seems incorrect.  The trait being displayed has a default value of None, which is allowed, it seems strange for the editor to override this.

traitsui/testing/tester/wx/helpers.py
```diff
@@ -135,6 +227,43 @@ def key_click_text_ctrl(control, interaction, delay):
...
+        control.SetInsertionPointEnd()
+    if interaction.key == "Enter":
+        wx.MilliSleep(delay)
+        event = wx.CommandEvent(wx.EVT_TEXT_ENTER.typeId, control.GetId())
+        control.ProcessEvent(event)
+    elif interaction.key == "Backspace":
+        wx.MilliSleep(delay)
+        start_sel, end_sel = control.GetTextSelection()
+        if start_sel != end_sel:
+            control.Remove(start_sel, end_sel)
```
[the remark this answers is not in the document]
aaronayres35: I made the change to `key_click_text_entry` and updated everywhere `key_click_text_ctrl` was used.  Unfortunately, this Pr's was branch was created before #1190 was merged.  Thus, this PR will now break things.  Should I undo the name change for now to avoid breaking anything and then update the name later in a separate PR after this is merged?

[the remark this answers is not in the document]
kitchoi: I think in this case we can first merge master into this branch, resolve the merge conflicts, then we should see that that all we need is to rename `key_click_text_ctrl` to  `key_click_text_entry` in here:
https://github.com/enthought/traitsui/blob/1b2a3933c5313f1d4d2084f197bdb6fc8e4c1ecf/traitsui/testing/tester/wx/registry_helper.py#L38
Given the scope is so small, it seem to me a perfectly reasonable thing to simply do the renaming here.

traitsui/testing/tester/qt4/implementation/enum_editor.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def radio_displayed_selected_text_handler(wrapper, interaction):
+    """ Handler function used to query DisplayedText for EnumRadioEditor.
+
+    Parameters
+    ----------
+    wrapper : UIWrapper
+        The UIWrapper containing that object with text to be displayed.
+    interaction : query.DisplayedText
```
kitchoi, line 126: Reminder: This `DisplayedText` will need updating, also in the description of the function.

traitsui/tests/editors/test_enum_editor.py
```diff
@@ -438,53 +344,71 @@ def test_radio_enum_editor_button_update(self):
...
+                tester = UITester()
+                with tester.create_ui(enum_edit, dict(view=get_radio_view(cols=cols))) as ui:
+                    # sanity check
+                    self.assertEqual(enum_edit.value, "one")
+                    radio_editor = tester.find_by_name(ui, "value")
+                    if is_qt():
+                        radio_editor.target.row_major = row_major
+                        radio_editor.target.rebuild_editor()
+                    radio_editor.locate(locator.Index(3)).perform(command.MouseClick())
+                    self.assertEqual(enum_edit.value, "four")
```
kitchoi, line 359: Now we are testing behaviours that weren't tested before, 🎉 !

traitsui/tests/editors/test_enum_editor.py
```diff
@@ -521,3 +445,41 @@ def test_list_evaluate_editor_text(self):
 
     def test_list_evaluate_editor_index(self):
         self.check_enum_index_update(get_evaluate_view("custom", mode="list"))
+
+    @requires_toolkit([])
```
kitchoi, line 449: Reminder: To put something in this list.

traitsui/qt4/enum_editor.py
```diff
@@ -425,7 +425,9 @@ def update_editor(self):
         """
         control = self.control
         try:
+            print("my value: ", self.value)
             value = self.inverse_mapping[self.value]
+            print("my value2: ", self.value)
```
kitchoi, line 430: Reminder: To remove debugging prints.

kitchoi (commented):
Thank you!
Some new comments.
(Reminder to run flake8 again at the end... I think I am annoyed about this enough I am going to put it up, today!)

traitsui/testing/tester/wx/implementation/enum_editor.py
aaronayres35: One thing to note is that this code will likely be applicable to checklist editor as well i realized so the module `traitsui.testing.tester.editors.enum_editor` might be too specific

kitchoi: Hmmm, `layout` instead of `enum_editor`?

kitchoi: (Eventually I think it should really be given a private module name, so if we regret the name later there is less to worry about. But let's wrap up all the pending PRs first before doing the overall naming changes. See first item of #1173 )

traitsui/tests/editors/test_enum_editor.py
aaronayres35: Use of a format func was already tested separately, so I just left it as None explicitly.  Thus, it gets displayed as an option with the text 'None'.

traitsui/tests/editors/test_enum_editor.py
```diff
@@ -521,3 +448,47 @@ def test_list_evaluate_editor_text(self):
 
     def test_list_evaluate_editor_index(self):
         self.check_enum_index_update(get_evaluate_view("custom", mode="list"))
+
+    def check_list_enum_none_selected(self, view):
+        enum_edit = EnumModelWithNone()
+        view = View(
```
kitchoi, line 454: Whatever `view` provided is overridden here, so the `test_list_enum_none_selected` and `test_list_evaluate_none_selected` are effectively identical. Perhaps we can just change this `check_list_...` to `test_...` and remove the other two?

kitchoi (approved):
Looks great. Thank you very much!
Some minor comments.

aaronayres35:
Looking at the CI I saw a test was failing for windows/wx:
```
FAIL: test_radio_enum_none_selected (traitsui.tests.editors.test_enum_editor.TestRadioEnumEditor)
  File "C:\Users\appveyor\.edm\envs\traitsui-test-3.6-wx\lib\site-packages\traitsui\tests\editors\test_enum_editor.py", line 382, in test_radio_enum_none_selected
    self.assertEqual(enum_edit.value, None)
AssertionError: 'one' != None
```
It seems that for windows/wx the enum traits value is forced to be 'one' whereas for all other platforms/ toolkit combos this is not the case.  I am skipping this test for now on windows/wx, and will merge once CI completes (I would like to have it merged today so that I can use this code in other PRs more easily).  However, we will likely either want to rewrite the test, or add a separate test for the wx/windows specific case.  I am not sure, @kitchoi lmk what you think is best.

PR #1184, opened 2020-09-02
fix UITester wx - explicitly handle a KeyClick('Backspace') and update key_click_text_ctrl

aaronayres35:
fixes #1183 
fixes #1182 
fixes #1177 
This PR explicitly handles the case when a KeyClick command is issued with "Backspace" as the key.  It also makes the same change that was added in PR #1181 for `key_sequence_text_ctrl` to `key_click_text_ctrl`
Finally, it makes the default behavior for entering a textbox to have the insertion point be the rightmost rather than the leftmost point of text.  (this keeps consistent behavior with qt)

PR #1193, opened 2020-09-04
Fix backspacing wx textbox when there is selected text

kitchoi:
This PR fixes the implementation of `KeyClick("Backspace")` on wx.TextCtrl when there is already some selected text.

For reference, we could not get [`wx.TextCtrl.EmulateKeyPress`](https://wxpython.org/Phoenix/docs/html/wx.TextCtrl.html#wx.TextCtrl.EmulateKeyPress) to emulate backspace on Windows, hence we had to emulate it ourselves. See https://github.com/enthought/traitsui/pull/1184 and this [discussion on wxPython forum](https://discuss.wxpython.org/t/wx-textctrl-and-emulatekeypress/28300).

### DIFF
--- a/traitsui/testing/tester/wx/tests/test_helpers.py
+++ b/traitsui/testing/tester/wx/tests/test_helpers.py
@@ -120,11 +120,40 @@ class TestInteractions(unittest.TestCase):
 
     def test_key_click(self):
         textbox = wx.TextCtrl(self.frame)
+        handler = mock.Mock()
+        textbox.Bind(wx.EVT_TEXT, handler)
 
         helpers.key_click_text_ctrl(textbox, command.KeyClick("A"), 0)
+
         self.assertEqual(textbox.Value, "A")
+        self.assertEqual(handler.call_count, 1)
+
+    def test_key_click_backspace(self):
+        textbox = wx.TextCtrl(self.frame)
+        textbox.SetValue("A")
+        handler = mock.Mock()
+        textbox.Bind(wx.EVT_TEXT, handler)
+
         helpers.key_click_text_ctrl(textbox, command.KeyClick("Backspace"), 0)
+
         self.assertEqual(textbox.Value, "")
+        self.assertEqual(handler.call_count, 1)
+
+    def test_key_click_backspace_with_selection(self):
+        textbox = wx.TextCtrl(self.frame)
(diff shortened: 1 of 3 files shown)


traitsui/tests/editors/test_text_editor.py
```diff
@@ -163,12 +163,13 @@ def test_custom_editor_init_and_dispose_no_auto_set(self):
...
     def test_simple_auto_set_update_text(self):
-        foo = Foo()
+        foo = Foo(name="a")
         view = get_view(style="simple", auto_set=True)
         tester = UITester()
         with tester.create_ui(foo, dict(view=view)) as ui:
-            with self.assertTraitChanges(foo, "name", count=3):
+            with self.assertTraitChanges(foo, "name", count=4):
                 name_field = tester.find_by_name(ui, "name")
+                name_field.perform(command.KeyClick("Backspace"))
```
kitchoi, line 172: On Windows, it goes through the branch where there is selected text... because the first input is given focus with all text selected. On OSX, the input is initialized such that it does not already have focus, so we go through the other branch where we are backspacing from the end of text. Since there is only one character, the overall effect is the same.

On second thought, that's perhaps too much complication for the purpose of this test. I will remove this change. The unit test for key click is good enough.

traitsui/testing/tester/wx/tests/test_helpers.py
```diff
@@ -120,11 +120,40 @@ def test_key_sequence_disabled(self):
...
         self.assertEqual(textbox.Value, "")
+        self.assertEqual(handler.call_count, 1)
+
+    def test_key_click_backspace_with_selection(self):
+        textbox = wx.TextCtrl(self.frame)
+        textbox.SetFocus()
+        textbox.SetValue("ABCDE")
+        textbox.SetSelection(0, 4)
+        # sanity check
+        self.assertEqual(textbox.GetStringSelection(), "ABCD")
```
kitchoi, line 148: Pedantic: This is not really a test objective so if the setup is not what we expect, we should raise a different exception to cause the test to "error", not to "fail".  I am not a purist and the comment explains the intent of this assertion already so I am going to leave it here. Happy to remove it too if others find this confusing.

rahulporuri (approved):
LGTM.

PR #1199, opened 2020-09-07
Fix for menu actions with on_perform being performed twice

### LINKED ISSUE #741: TableEditor menu actions with on_perform are performed twice
Running the following script, right clicking the table, and pressing "Click me!" prints:
```
Called with 1 args!
Called with 0 args!
```

I believe the one-arg perform comes from the BaseTableEditor's `_perform` method, and the zero-arg perform comes from PyFace's Action's `perform` method.

https://github.com/enthought/traitsui/blob/48fc57f504b284d5c420b16b4330356550b448f3/traitsui/editors/table_editor.py#L520-L522

https://github.com/enthought/pyface/blob/03cbcfbe6e36167576a97462fa86ea9ef1b0ea79/pyface/action/action.py#L126-L135

```python
import random

from traits.api import HasTraits, List, Int, Instance
from traitsui.api import ModelView, TableEditor, Menu, Action, ObjectColumn, View, Item


class Result(HasTraits):
    value = Int

    def _value_default(self):
        return random.randint(1, 100)


class MyModel(HasTraits):
    results = List(Instance(Result))

    def _results_default(self):
        return [Result() for _ in range(20)]


def my_table(view):
    return TableEditor(
        menu=Menu(
            Action(
                name="Click me!",
                on_perform=view.do_perform,
            )
        ),
        columns=[
            ObjectColumn(
                name="value",
                label="A number",
                editable=False,
            ),
        ],
    )


class MyView(ModelView):
    model = Instance(MyModel)

        return View(
            Item(
                "model.results",
                editor=my_table(self),
            )
        )

    def do_perform(self, *args):
        print("Called with {} args!".format(len(args)))


if __name__ == "__main__":
    model = MyModel()
    view = MyView(model=model)

    view.configure_traits()
```

jjenthought:
This should fix #741, though I'm not entirely sure where to add tests to confirm it! Having it work in one canopy data app where this came up probably isn't sufficient..
 
The problem seemed to be that `Action`s with an `on_perform` attribute hit :
https://github.com/enthought/traitsui/blob/884414547b634ce30421df29e817bb6beaa1036c/traitsui/qt4/editor.py#L260-L261
and called the function set as `on_perform`, but then afterwards hit:
https://github.com/enthought/traitsui/blob/884414547b634ce30421df29e817bb6beaa1036c/traitsui/qt4/editor.py#L263
which basically does the same thing again, only in the `Action` class this time.
https://github.com/enthought/pyface/blob/4b91627c8a8ab2680f62e85af261d1da54675558/pyface/action/action.py#L123-L132

Just returning after the first one would mean `on_perform` would take precedence over a overridden `perform` method if an `Action` had both of these. If creating `Action`s with both these properties was an expected usage, this PR won't be enough!

### DIFF
--- a/traitsui/qt4/editor.py
+++ b/traitsui/qt4/editor.py
@@ -259,6 +259,7 @@ class Editor(UIEditor):
 
         if action.on_perform is not None:
             action.on_perform(selection)
+            return
 
         action.perform(selection)
 
--- a/traitsui/tests/editors/test_table_editor.py
+++ b/traitsui/tests/editors/test_table_editor.py
@@ -3,8 +3,7 @@ from unittest.mock import Mock
 
 from traits.api import HasTraits, Instance, Int, List, Str, Tuple
 
-from traitsui.api import EvalTableFilter, Item, ObjectColumn, TableEditor, View
-from traitsui.menu import Action, Menu
+from traitsui.api import Action, EvalTableFilter, Item, ObjectColumn, TableEditor, View
 from traitsui.tests._tools import (
     create_ui,
     is_qt,
--- a/traitsui/tests/editors/test_table_editor.py
+++ b/traitsui/tests/editors/test_table_editor.py
@@ -1,8 +1,10 @@
 import unittest
+from unittest.mock import Mock
 
 from traits.api import HasTraits, Instance, Int, List, Str, Tuple
 
 from traitsui.api import EvalTableFilter, Item, ObjectColumn, TableEditor, View
+from traitsui.menu import Action, Menu
 from traitsui.tests._tools import (
     create_ui,
     is_qt,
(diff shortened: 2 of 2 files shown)


kitchoi (approved):
I doubt the `perform` method is meant to be called twice, and returning early is consistent with the implementation in table editor:
https://github.com/enthought/traitsui/blob/9727b1ad53b18fcc84b58808824dbe072081f583/traitsui/editors/table_editor.py#L519-L521

In some places, the second invocation of `action.perform` isn't even there:
https://github.com/enthought/traitsui/blob/148457a08b4f0fc14ccff6dc2166fa8f0251643c/traitsui/wx/tree_editor.py#L1494-L1496

>  though I'm not entirely sure where to add tests to confirm it! Having it work in one canopy data app where this came up probably isn't sufficient..

If you already have a test written, it would be good to add it somewhere, the location of the test can be changed later if necessary.

kitchoi:
And thanks for this.

jjenthought:
> If you already have a test written, it would be good to add it somewhere, the location of the test can be changed later if necessary.

I haven't written anything beyond the separate code fragment I was using as a check, I'm not sure how testing changes in just the qt implementation of `editor.py` would fit in with everything else in the test suite. I'll upload the test I had, but it's just for the table editor unfortunately.

kitchoi (approved):
Still LGTM. Thank you very much for the test.

PR #1207, opened 2020-09-09
UI Tester updates - implement ui_base

### LINKED ISSUE #1198: Add ui_base implementation for UI Tester
Note that once this is completed (and the find_by_id pr #1195 is merged) I believe we should be able to test the https://github.com/enthought/traitsui/blob/master/traitsui/examples/demo/Applications/converter.py demo using UI Tester.

aaronayres35:
fixes #1198 
adding ui_base implementation and implicit test via testing the converter.py application demo.  

Explicit tests should be added for ui_base.  At the moment I'm unsure of the best way to go about that ~~(hence this PR currently being a draft).~~ @kitchoi I marked this PR as ready for review, but I still feel like the ui_base code needs more explicit tests. I would be happy to have a chat tomorrow about it if you have the time.


integrationtests/test_all_examples.py
```diff
@@ -279,6 +281,14 @@ def run_file(file_path):
         exec(content, globals)
 
 
+def load_demo(file_path, variable_name="demo"):
```
kitchoi, line 284: Docstring please.

traitsui/testing/tester/qt4/implementation/ui_base.py
```diff
@@ -0,0 +1,36 @@
...
+    If there are any conflicts, an error will occur.
+
+    Parameters
+    ----------
+    registry : TargetRegistry
+    """
+    handlers = [
+        (command.MouseClick, (lambda wrapper, _:  helpers.mouse_click_qwidget(
+                              wrapper.target.control, wrapper.delay))),
+        (query.DisplayedText, lambda wrapper, _: wrapper.target.control.text())
```
kitchoi, line 29: This handler for `DisplayedText` not used in the test, maybe it is needed in the future but I think it is better to defer it until we actually have a use case to use it. It is better to avoid speculative generality, this is especially true when it comes to something that will become a published API.

integrationtests/test_all_examples.py
```diff
@@ -317,3 +327,31 @@ def test_run(self):
...
+    """ Test examples with more interactions."""
+
+    @requires_toolkit([ToolkitName.qt, ToolkitName.wx])
+    def test_converter(self):
+        # Test converter.py in examples/demo/Applications
+        filepath = os.path.join(
+            DEMO, "Applications", "converter.py"
+        )
+        demo = load_demo(filepath, "popup")
+        tester = UITester(delay=500)
```
kitchoi, line 342: This `delay` is meant to be removed?

kitchoi (commented):
Thank you very much. Some minor comments.

The test support is used in integration tests, which are now only run with Linux on CI.
Could we add a small test that exercise clicking the UI button that can be exercised cross platforms?

kitchoi:
Ah I see what you mean by this now:

> I marked this PR as ready for review, but I still feel like the ui_base code needs more explicit tests. 

Yep I agree.

traitsui/testing/tester/qt4/implementation/ui_base.py
kitchoi: Or if we change the following test to use DisplayedText, then this represents an existing use case and we can keep this handler here:
https://github.com/enthought/traitsui/blob/24bcf02fe2216237fd207b4f679629416d073754/traitsui/tests/test_ui.py#L191

kitchoi:
Looks like we can modify a few lines in `traitsui.tests.test_ui.TestUI.test_destroy_after_ok_wx` and `traitsui.tests.test_ui.TestUI.test_destroy_after_ok_qt` to use the test support code added here. 

```
            # press the OK button and close the dialog
            ok_button = tester.find_by_id(ui, "OK")
            self.assertEqual(ok_button.inspect(query.DisplayedText()), "OK")
            ok_button.perform(command.MouseClick())
```

That would address my own comments about using `DisplayedText` and having it tested cross platform, I don't know if that would address your own comment too.

kitchoi:
> The test support is used in integration tests, which are now only run with Linux on CI.
Could we add a small test that exercise clicking the UI button that can be exercised cross platforms?

I was wrong. The integration tests are run on Windows and OSX CI builds. But not for Wx, because the test command for wx fails early, precluding the command for integration tests to run.

integrationtests/test_all_examples.py
```diff
@@ -279,6 +281,24 @@ def run_file(file_path):
         exec(content, globals)
 
 
+def load_demo(file_path, variable_name="demo"):
+    """ Loads a demo example from given file_path. Extracts the relevant
+    object via variable_name.
+    Parameters
```
kitchoi, line 287: There is a missing newline before `Parameters`.  It is missing in the `run_file` function too. Feel free to add that too as a drive-by (small enough to not worth a separate PR).

integrationtests/test_all_examples.py
```diff
@@ -279,6 +281,24 @@ def run_file(file_path):
...
+    ----------
+    file_path : str
+        The file_path of the file to be loaded
+    variable_name : str
+        The key in the global symbol state corresponding to the object of
+        interest for the demo.
```
kitchoi, line 293: This function returns something, please add a "Returns" block here.  We expect the return object to be an instance of `HasTraits` because we will need to call `edit_traits` on it. That should be mentioned.

kitchoi (commented):
LGTM. Some suggestions on docstrings.

kitchoi (approved):
LGTM!

PR #1208, opened 2020-09-10
UI Tester updates - Instance editor implementation

### LINKED ISSUE #1206: Implement Instance Editor for UI Tester
Instance Editor is another of the more commonly used editors, and thus would be good to add soon.

aaronayres35:
fixes #1206 
Adds a simple implementation for Instance editor, and modifies tests to now use UI Tester.  Also adds an additional test to see that the displayed editor updates if a trait changes externally to the editor.  

Note: the code for the wx instance editor had to be changed very slightly in order to gain access to the ui of interest.  This idea was take from kit's draft branch #1150

### DIFF
--- a/traitsui/testing/tester/qt4/implementation/instance_editor.py
+++ b/traitsui/testing/tester/qt4/implementation/instance_editor.py
@@ -16,12 +16,24 @@ from traitsui.qt4.instance_editor import (
 
 
 def _get_netsed_ui_simple(target):
+    """ Obtains a nested UI within a Simple Instance Editor.
+
+    Parameters
+    ----------
+    target : instance of SimpleEditor
+    """
     if target._dialog_ui is None:
         target._button.click()
     return target._dialog_ui
 
 
 def _get_netsed_ui_custom(target):
+    """ Obtains a nested UI within a Custom Instance Editor.
+
+    Parameters
+    ----------
+    target : instance of CustomEditor
+    """
     return target._ui
 
 
--- a/traitsui/testing/tester/qt4/implementation/instance_editor.py
+++ b/traitsui/testing/tester/qt4/implementation/instance_editor.py
@@ -1,0 +1,39 @@
+#  Copyright (c) 2005-2020, Enthought, Inc.
+#  All rights reserved.
+#
+#  This software is provided without warranty under the terms of the BSD
+#  license included in LICENSE.txt and may be redistributed only
+#  under the conditions described in the aforementioned license.  The license
+#  is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+#  Thanks for using Enthought open source!
+#
+from traitsui.testing.tester import command, locator
+from traitsui.testing.tester.registry_helper import register_nested_ui_solvers
+from traitsui.qt4.instance_editor import (
+    CustomEditor,
+    SimpleEditor
+)
+
+def _get_netsed_ui_simple(target):
+    if target._dialog_ui is None:
+        target._button.click()
+    return target._dialog_ui
+
+
+def _get_netsed_ui_custom(target):
+    return target._ui
+
+
+def register(registry):
+    """ Register interactions for the given registry.
+
+    If there are any conflicts, an error will occur.
+
+    Parameters
+    ----------
+    registry : TargetRegistry
+        The registry being registered to.
+    """
+    register_nested_ui_solvers(registry, SimpleEditor, _get_netsed_ui_simple)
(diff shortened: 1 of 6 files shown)


traitsui/testing/tester/qt4/implementation/instance_editor.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def _get_netsed_ui_simple(target):
+    if target._dialog_ui is None:
+        target._button.click()
+    return target._dialog_ui
+
+
+def _get_netsed_ui_custom(target):
+    return target._ui
```
aaronayres35, line 25: need doc strings (same with wx).  I will add shortly

traitsui/tests/editors/test_instance_editor.py
```diff
@@ -19,28 +20,59 @@ class EditedInstance(HasTraits):
...
+            self.assertEqual(displayed, "hello")
+            edited_inst.value = "bye"
+            displayed = value_txt.inspect(query.DisplayedText())
+            self.assertEqual(displayed, "bye")
+
+    def test_custom_editor_resynch_editor(self):
+        return self.check_resynch_editor('custom')
+
+    def test_simple_editor_resynch_editor(self):
+        return self.check_resynch_editor('simple')
```
kitchoi, line 71: These `return` can be removed.

traitsui/tests/editors/test_instance_editor.py
```diff
@@ -19,28 +20,59 @@ class EditedInstance(HasTraits):
...
+        return self.check_editor_update_value('custom')
+
+    def check_resynch_editor(self, style):
+        edited_inst = EditedInstance(value='hello')
+        obj = NonmodalInstanceEditor(inst=edited_inst)
+        tester = UITester()
+        with tester.create_ui(obj, dict(view=get_view(style))) as ui:
+            instance = tester.find_by_name(ui, "inst")
+            editor = instance.locate(locator.NestedUI())
+            value_txt = editor.find_by_name("value")
```
kitchoi, line 60: Could we write `value_txt = tester.find_by_name(ui, "inst").find_by_name("value")` here instead?

traitsui/tests/editors/test_instance_editor.py
```diff
@@ -19,28 +20,59 @@ class EditedInstance(HasTraits):
...
+            value_txt.perform(command.KeySequence("abc"))
+            if style == 'simple':
+                # the below code requires ui_base implementation to work
+                # ok_button = editor.find_by_id("OK")
+                # ok_button.perform(command.MouseClick())
+                press_ok_button(editor.target)
```
kitchoi, line 44: This action looks redundant: the trait should be updated even without pressing the OK button, unless we need to switch focus for wx to trigger the change?

traitsui/wx/instance_editor.py
```diff
@@ -527,6 +527,7 @@ def edit_instance(self, event):
             # have its own:
             if ui.history is None:
                 ui.history = self.ui.history
+        return ui
```
kitchoi, line 530: Following the pattern in the Qt implementation, it might be better to assign `ui` to a private attribute on the object and make sure its `dispose` method is called, this will allow the mouse click to be a separate interaction.

kitchoi (commented):
Thank you!  Some comments for discussions, see #1212 which accompanies the review comments.

traitsui/testing/tester/qt4/implementation/instance_editor.py
```diff
@@ -0,0 +1,51 @@
...
+
+def _get_netsed_ui_simple(target):
+    """ Obtains a nested UI within a Simple Instance Editor.
+
+    Parameters
+    ----------
+    target : instance of SimpleEditor
+    """
+    if target._dialog_ui is None:
+        target._button.click()
```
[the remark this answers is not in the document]
kitchoi: Now these two lines can be removed as they are redundant in tests where a mouse click is performed prior to querying for the dialog UI.  I am sorry I probably should have done that in #1206 😓 .

If the test code tries to navigate into the dialog UI without performing a mouse click first, the dialog UI is None and we get an error anyway complaining no location solvers are registered on `None`. It might be nicer to raise early with a different error message but I think it might be a bit overkill.

kitchoi (commented):
I am sorry I omitted to remove the click action in Qt `_get_nested_ui_simple` in #1206... One last comment.

kitchoi (approved):
LGTM!

PR #1209, opened 2020-09-11
Ui tester range editor updates

### LINKED ISSUE #1188: Expand UI Tester Range Editor implementation
As mentioned here: https://github.com/enthought/traitsui/pull/1171#discussion_r480143531 we will want to include an implementation for SimpleSpinEditor
This will require additional solvers and potentially special handling of the SpinBox/SpinCtrl object.  

Additionally, we will want to add the functionality for locating and interacting with a slider, and for clicking on any buttons (e.g. in the SimpleSpin Editor or LargeRangeSliderEditor)

aaronayres35:
addresses part of #1188 
When adding the wx implementation fo the SimpleSpin range editor I ran into the same problem that was observed in https://github.com/enthought/traitsui/pull/1191#issuecomment-687259212 which were introduced in 
#1193 .

It would be convenient to have a function for key sequence and key click that works for all wx textbooks.  We recently added a refactor in #1190 to introduce the function `register_editable_textbox_handlers`.  However, on wx this functions name is a little misleading as it only works for wx.TextCtrl not other editable textboxes such as a wx.ComboBox, or a wx.SpinCtrl.  

My main dilemma now is that to implement this on wx, I will either need to register handlers for the SimpleSpinEditor itself (which deviates from the pattern of all the other editors in which they utilize a LocatedTextbox as a middle man) or I will need to modify the LocatedTextbox implementation to handle SpinCtrls specially.  Both seem like ugly solutions to me, I am still trying to decide on the best approach.  

I could also try to modify the Key_click_text_ctrl function to not be TextCtrl specific but that introduces a whole slew of other problems


aaronayres35:
I am very tempted to just stop splitting this PR into two.  This PR could keep the slider additions, and possibly the SimpleSpin implementation on Qt.  Another issue/PR could bee opened to handle the wx SimpleSpin implementation.  
Not even considering the wx problems, even if everything worked fine it would be a bit of a lengthy PR anyway.

traitsui/testing/tester/qt4/implementation/range_editor.py
```diff
@@ -36,10 +41,32 @@ def resolve_location_slider(wrapper, location):
...
+    wrapper : UIWrapper
+        Wrapper containing the SimpleSpin Range Editor target.
+    location : locator.WidgetType
+        The location we are looking to resolve.
+    """
+    if location == locator.WidgetType.textbox:
+        # FIX ME: THIS CODE DOESNT BELONG HERE
+        textbox = wrapper.target.control.lineEdit()
+        textbox.end(False)
+        return LocatedTextbox(textbox=textbox)
```
aaronayres35, line 69: This was another insertion point issue.  It seems qts default behavior was different here for whatever reason.  This should probably be moved to inside helpers though

traitsui/tests/editors/test_range_editor.py
```diff
@@ -115,6 +121,10 @@ def check_set_with_text_after_empty(self, mode):
     def test_simple_slider_editor_set_with_text_after_empty(self):
         return self.check_set_with_text_after_empty(mode='slider')
 
+    @requires_toolkit([ToolkitName.qt])
+    def test_simple_spin_editor_set_with_text_after_empty(self):
+        return self.check_set_with_text_after_empty(mode='spinner')
```
aaronayres35, line 125: This was failing on wx with the very common taceback of: 
```
wx._core.wxAssertionError: C++ assertion "GetEventHandler() == this" failed at /Users/robind/projects/bb2/dist-osx-py36/build/ext/wxWidgets/src/common/wincmn.cpp(475) in ~wxWindowBase(): any pushed event handlers must have been removed
```

aaronayres35:
> I am very tempted to just stop splitting this PR into two. This PR could keep the slider additions, and possibly the SimpleSpin implementation on Qt. Another issue/PR could bee opened to handle the wx SimpleSpin implementation.
> Not even considering the wx problems, even if everything worked fine it would be a bit of a lengthy PR anyway.

I ended up doing this, but I did include the simple spin for Qt.  @kitchoi if you would prefer I can remove this and add it back in another pr where I would add the wx implementation as well.  Or we could just leave it here with this PR and add the wx implementation alone later.

aaronayres35:
> Edit: Actually, after the Enum editor PR is merged, key_click_text_ctrl is now key_click_text_entry and is not text ctrl specific. Once that is merged, I will try to simply add spin editor for wx here.

I don't know why I had remembered wx.SpinCtrl as subclassing wx.TextEntry, but it doesn't so this issue is still annoying to solve, and will require one of the solutions I proposed in the description for this PR.

traitsui/testing/tester/qt4/implementation/range_editor.py
aaronayres35: This kind of problem was seen before, and addressed by adding:
```
if not control.hasFocus():
        key_click(widget=control, key="End", delay=0)
``` 
to `keey_sequence_textbox`.  However, adding the same thing to `key_click_qwidget` does not seem to solve this problem here, and we also shouldn't call the method used above: `textbox.end(False)` because this function is supposed to work with general qwidgets, and those may not have the .end() method.

aaronayres35: maybe this code is fine where it is...?

traitsui/testing/tester/wx/helpers.py
```diff
@@ -316,3 +316,26 @@ def key_sequence_text_ctrl(control, interaction, delay):
     for char in interaction.sequence:
         wx.MilliSleep(delay)
         control.WriteText(char)
+
+
+def key_click_slider(control, interaction, delay):
+    valid_keys = {"Left", "Right", "Up", "Down", "Page Up", "Page Down"}
```
kitchoi, line 322: I also realize something fun... on OSX, Up / Page Up increments the value. On another Linux machine of mine, Up / Page Up decrements the value. I did not notice that before.

This reminds me of the system setting to do with natural scrolling on trackpad / mouse wheel... some people think mouse wheel up should do a page down, where some people think a mouse wheel up should do a page up.  Perhaps this is yet another personal preference and we should just stick with Left and Right, which is unambiguous.

kitchoi (commented):
Thank you for looking into this!  Wx SpinCtrl claims to be like TextCtrl but it has not really implemented the interface of TextCtrl, so its support will have to be implemented differently.

If we couldn't get wx support in, I'd be in favour of deferring the spinner support via the default registry. For example, we can move the Qt implementation you have here to a local register and use it in traitsui tests locally.  Later we can expand the support for wx, and once we are happy that the API is consistent cross toolkits, then we can move the implementations over to the default registry for downstream projects to use. The rationale being that once an API is published for external uses, that introduces limitations on what can be changed. If we publish it prematurely, e.g. not having it tested out for more than one toolkit, we might find ourselves having to workaround the published API in order to accommodate differences between Qt and Wx.

Some preliminary comments, and I can review this again when this is finalized.

aaronayres35:
For what ever reason I did not have the option to reply to this comment directly. 

>Also I see a spinbox as one widget. Therefore it is somewhat surprising to me to have to call locate(WidgetType.textbox) in order to modify the text. The test code for mode="spinner" may not need to be identical to the test code written for mode="spinner".
I was thinking that the spinner also contains an up and down button so one could locate the buttons or the textbox as the different "sub-widgets" but I see what you mean about it really basically being one widget.  Also, Im assuming you meant code for mode'"spinner" does not need to be same as code for mode="slider".

>One approach maybe to reuse register_editable_textbox_handlers on SimpleSpinEditor with widget_getter=lambda wrapper: wrapper.target.control.lineEdit()?
I think that would make sense.

traitsui/testing/tester/wx/helpers.py
aaronayres35: This is when interacting with the UI manually I assume?  Also are you suggesting to remove "Up", "Page Up", etc from the list of valid keys, or just the tests (I read that as from the list of valid keys but wanted to confirm).  I had not thought of the case where "Up" would decrement.  I did like the ability to "Page Up"/"Page Down" for the difference is step/page step size but if it would be confusing to users it could be removed.  Could we also just document that "Up" corresponds to increment for the tester?

kitchoi:
> Also, Im assuming you meant code for mode'"spinner" does not need to be same as code for mode="slider".

Yes. Stupid typo on my part, sorry! 

> Even the RangeTextEditor does as well, and it certainly only contains the textbox. Maybe that should be changes as well so that it doesn't go tthrough locatedTextbox? what do you think?

Yes you are right! I did not notice that when we added the RangeTextEditor. Yes I agree RangeTextEditor does not need to go through LocatedTextBox.

traitsui/testing/tester/wx/helpers.py
kitchoi: Yes my comment was from manual testing.

> I did like the ability to "Page Up"/"Page Down" for the difference is step/page step size but if it would be confusing to users it could be removed. Could we also just document that "Up" corresponds to increment for the tester?

Yes I think documenting that "Up" corresponds to increment would also be reasonable!

traitsui/tests/editors/test_range_editor.py
```diff
@@ -68,30 +69,75 @@ def check_set_with_text_valid(self, mode):
...
+            displayed = number_field_text.inspect(query.DisplayedText())
+            self.assertEqual(model.value, 10)
+            self.assertEqual(displayed, str(model.value))
+
+    # SimpleSpinEditor not yet implemented on Wx.
+    @requires_toolkit([ToolkitName.qt])
+    def test_simple_spin_editor_set_with_text_valid(self):
+        from traitsui.testing.tester.registry import TargetRegistry
+        from traitsui.testing.tester.qt4.implementation.range_editor import (
+            _register_simple_spin
```
kitchoi, line 116: Since this is only used by this module, could we just define this function locally in this module?
The preceding underscore is used to signal when something is private. It being imported from outside goes against this convention.

If we will elevate the implementation to the default registry and to use it for RangeEditor tests, this function will be discovered here and be moved. Much like the `set_combobox_index`, `finish_combobox_text_entry` etc removed from test_enum_editor when the tester implementation was added.

traitsui/tests/editors/test_range_editor.py
```diff
@@ -113,16 +159,122 @@ def check_set_with_text_after_empty(self, mode):
...
+            # Delete all contents of textbox
+            for _ in range(5):
+                number_field_text.perform(command.KeyClick("Backspace"))
+            number_field_text.perform(command.KeySequence("11"))
+            number_field_text.perform(command.KeyClick("Enter"))
+            displayed = number_field_text.inspect(query.DisplayedText())
+            self.assertEqual(model.value, 11)
+            self.assertEqual(displayed, str(model.value))
+
+    # SimpleSpinEditor not yet implemented on Wx.
```
kitchoi, line 193: This comment reads to me as "the feature mode='spinner' is not implemented on Wx".
I think it would be better to say "the tester support code is not yet implemented for Wx SimpleSpinEditor".

traitsui/testing/tester/wx/helpers.py
```diff
@@ -316,3 +316,48 @@ def key_sequence_text_ctrl(control, interaction, delay):
...
+    """
+    valid_keys = {"Left", "Right", "Up", "Down", "Page Up", "Page Down"}
+    if interaction.key not in valid_keys:
+        raise ValueError(
+            "Unexpected Key. Supported keys are: {}".format(sorted(valid_keys))
+        )
+    if not control.HasFocus():
+        control.SetFocus()
+    value = control.GetValue()
+    wx.MilliSleep(delay)
```
kitchoi, line 346: Nit: It might be better to move this down to go before `SetValue` so if we are going run into the ValueError, there is no delay.

kitchoi (commented):
LGTM. Only minor suggestions.

kitchoi (approved):
Thank you!

aaronayres35:
Looks like we missed an error on wx: 
```
ERROR: test_wx_text_editing (traitsui.tests.editors.test_range_editor_text.TestRangeEditorText)
  File "/Users/appveyor/.edm/envs/traitsui-test-3.6-wx/lib/python3.6/site-packages/traitsui/tests/editors/test_range_editor_text.py", line 77, in test_wx_text_editing
    text = tester.find_by_name(ui, "number").locate(locator.WidgetType.textbox)
  File "/Users/appveyor/.edm/envs/traitsui-test-3.6-wx/lib/python3.6/site-packages/traitsui/testing/tester/ui_wrapper.py", line 86, in locate
    target=self._get_next_target(location),
  File "/Users/appveyor/.edm/envs/traitsui-test-3.6-wx/lib/python3.6/site-packages/traitsui/testing/tester/ui_wrapper.py", line 217, in _get_next_target
    supported=list(supported),
traitsui.testing.tester.exceptions.LocationNotSupported: Location <enum 'WidgetType'> is not supported for <class 'traitsui.wx.range_editor.RangeTextEditor'>. Supported these: []
```

It is in a separate testt file from tthe other range editor tests, so I had not noticed it.  It is a very easy fix, we just dont go through locator.WidgetType.textbox any more for the Range Text editor.  I will open a separate PR to fix.

kitchoi:
We forgot to manually run flake8 before merging too.

There was a race condition from PR merging: The last CI build for this PR was run before #1222 was merged. This PR happened to introduce an error from flake8, which is now causing CI to fail on #1225

aaronayres35:
> We forgot to manually run flake8 before merging too.
> 
> There was a race condition from PR merging: The last CI build for this PR was run before #1222 was merged. This PR happened to introduce an error from flake8, which is now causing CI to fail on #1225

I was very confused why 1225 was failing. Should I do the flake8 fix on #1226?

kitchoi:
Apparently there are new flake8 errors introduced by #1219 too.

> Should I do the flake8 fix on #1226?

Up to you, whichever makes sense and is efficient.

(Actually, I might have the race condition the other way round - the PRs that are introducing flake8 errors might have been merged already. Anyway, it is still race condition with PRs interacting with each other.)

PR #1213, opened 2020-09-15
Adding inline comments to EnumRadioEditor and CheckList editor code for populating grid

aaronayres35:
The code was a bit difficult to understand (see PR #1191 for reference) so I added a few quick comments to. hopefully make it a little clearer for someone looking in the future.  Eventually we may want to refactor this code all together so that it is simpler to follow.

### DIFF
--- a/traitsui/qt4/check_list_editor.py
+++ b/traitsui/qt4/check_list_editor.py
@@ -199,15 +199,23 @@ class CustomEditor(SimpleEditor):
         n = len(labels)
         cols = self.factory.cols
         rows = (n + cols - 1) // cols
+        # incr will keep track of how to increment index so that as we traverse
+        # the grid in row major order, the elements are added to appear in
+        # column major order
         incr = [n // cols] * cols
         rem = n % cols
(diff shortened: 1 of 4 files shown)


traitsui/qt4/check_list_editor.py
```diff
@@ -199,15 +199,23 @@ def rebuild_editor(self):
...
         cols = self.factory.cols
         rows = (n + cols - 1) // cols
+        # incr will keep track of how to increment index so that as we traverse
+        # the grid in row major order, the elements are added to appear in
+        # column major order
         incr = [n // cols] * cols
         rem = n % cols
         for i in range(cols):
             incr[i] += rem > i
         incr[-1] = -sum(incr[:-1]) + 1
```
aaronayres35, line 209: It is very strange that this line isn't `incr[-1] = -(reduce(lambda x, y: x + y, incr[:-1], 0) - 1)` like it is in all the other files which do the same thing.

kitchoi (approved):
LGTM!

PR #1215, opened 2020-09-15
UI Tester checklist editor indexing fix

### LINKED ISSUE #1214: UI Tester CheckList Editor implementation should use column_major_to_row_major and update tests
The Checklist editor does the same things that the radio style enum editor does when populating a grid. The same changes that were added to PR #1191 should be added for checklist editor as well, along with tests to ensure the behavior is correct.

aaronayres35:
fixes #1214 
Effectively this does exactly what was done for the enum editor in pr #1191 
It handles the strange behavior of the checklist grid being populated in row major with order with the elements arranged in column major order.


kitchoi (approved):
LGTM

PR #1216, opened 2020-09-16
TST: Update a test on Radio enum editor to use UITester

kitchoi:
One of the existing tests (`test_radio_enum_editor_button_update`) for the radio enum editor checks the UI states are updated when the trait state is updated.
The test was added prior to the introduction of UITester.

The implementation of the test does this by getting all the states of the radio buttons, but since radio buttons are "auto-exclusive" for one-of-many selections, the implementation for querying for the selected item achieves the same goal.

This PR:
- Update the test to use UITester features. This demonstrates one of the two patterns of UI testing: Updating the trait results in UI state change.
- Remove `test_radio_enum_displayed_selected_text` because its purpose is for exercising the logic of `SelectedText`. The test updated here now uses `SelectedText` so `test_radio_enum_displayed_selected_text` is redundant.

Note: The other, unchanged `test_radio_enum_editor_pick` demonstrates the second pattern of UI testing - Updating the UI state via user interaction updates the trait.
Note: The auto-exclusive behaviour is not coded in traitsui, it is a feature provided by Qt and Wx Radio button widgets. But if in doubt, we can always update the tester implementation to do some sanity check there - but we will be testing Qt / Wx rather than testing traitsui.

### DIFF
--- a/traitsui/tests/editors/test_enum_editor.py
+++ b/traitsui/tests/editors/test_enum_editor.py
@@ -316,18 +316,13 @@ class TestSimpleEnumEditor(unittest.TestCase):
 @requires_toolkit([ToolkitName.qt, ToolkitName.wx])
 class TestRadioEnumEditor(unittest.TestCase):
 
-    @contextlib.contextmanager
-    def setup_gui(self, model, view):
-        with create_ui(model, dict(view=view)) as ui:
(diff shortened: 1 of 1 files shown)


aaronayres35 (approved):
LGTM!

PR #1217, opened 2020-09-16
Maint: Reorganize optional dependencies in etstool.py and setup.py

kitchoi:
When I was trying to add flake8 to CI, I notice that the existing organization of dependencies in `etstool.py` will force me to add flake8 to the `dependencies` set. This led me to realize that the dependencies set has mixed runtime dependencies and CI dependencies.

This PR:
- Reorganize the dependencies in `etstool.py` to clarify what dependencies are used for what, with a goal of matching those used by `setup.py`
- Add an optional set of dependencies called `"editors"` to be used by `setup.py` to specify the dependencies on optional features:
      - `pandas` is required by DataFrameEditor. (Note that `DataFrameEditor` is not exposed in `api`)
      - `numpy` is required by ArrayEditor and DataFrameEditor. `traitsui.api` explicitly allows `numpy` to be absent. If numpy absent, `traitsui.api` won't include DataFrameEditor

I was tempted to go the extra mile to add a CI job to run with the minimal set of dependencies, but then I decided against it.  This PR should have no behavioural change to how CI is run.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -98,16 +98,10 @@ DEFAULT_TOOLKIT = 'null'
 # The requirement is to be interpreted by EDM
 TRAITS_VERSION_REQUIRES = os.environ.get("TRAITS_REQUIRES", "")
 
+# Required runtime dependencies. Should match install_requires in setup.py
 dependencies = {
-    "numpy",
     "traits" + TRAITS_VERSION_REQUIRES,
-    "pandas",
-    "pygments",
-    "pip",
-    "nose",
-    "coverage",
-    "configobj",
-    "docutils",
+    # pyface is installed from source via ci-src-requirements.txt
 }
 
 # NOTE : pyface is always installed from source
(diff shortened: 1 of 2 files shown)


etstool.py
```diff
@@ -98,46 +98,61 @@
...
+    },
+    'pyqt5': {
+        'pyqt5',
+        'pygments',
+    },
     # XXX once wxPython 4 is available in EDM, we will want it here
-    'wx': set(),
-    'null': set()
+    'wx': {
+        'numpy',
```
rahulporuri, line 129: Not sure why `numpy` is a dependency of `wx` toolkit. I understand that this is emulating what we see in `setup.py` but still.

rahulporuri (approved):
LGTM. Tested locally with pyqt5 toolkit and works as expected.

etstool.py
kitchoi: From what is covered by tests, numpy is required for `traitsui.theme` via `traitsui.wx.image_slice`.
I guess there are more... e.g. I can find `numpy` being imported at the top-level by `pyface.wx.color`.
So at the moment, it does not seem at all disposable for wx.

rahulporuri: ohh yeah. makes sense. 👍

PR #1218, opened 2020-09-16
Skip test module for DataFrameEditor if numpy is absent

kitchoi:
Currently the test module for DataFrameEditor will fail with ImportError if one runs `python -m unittest discover -t . traitsui` with an environment installed from PyPI, Qt5 and without NumPy.

NumPy is an optional dependency, e.g. the `ArrayEditor` is skipped in `traitsui.api` if importing it should fail due to absence of numpy.

This PR skips the test module if NumPy cannot be imported, in the same way that the module is skipped if pandas is not available.


rahulporuri (approved):
LGTM

PR #1219, opened 2020-09-16
UI Tester - use _BaseSourceWithLocation for Checklist and List editor implementations

### LINKED ISSUE #1197: Fully utilize recent refactors to UI Tester
In recent PRs #1191 (soon to be merged) and #1190 some refactoring was done, but these refactors are not utilized everywhere they can be.  The new `_BaseSourceWithLocation` base class added in 1191 could be used for CheckList Editor and likely List editor as well (it is currently only used for Enum editor).  Additionally the new `register_editable_textbox_handlers` from 1190 can likely be used in other places as well

aaronayres35:
fixes #1197 
I believe that `register_editable_textbox_handlers` is actually already currently used where it should be.  Thus, this PR simply changes the checklist and List editor implementations to utilize the new `_BaseSourceWithLocation` base class added in #1191 

Note this was originally sitting on top of PR #1215 (hence the couple extra commits in the commit history).  That is now merged though, so the files changed are correct.

### DIFF
--- a/traitsui/testing/tester/qt4/implementation/list_editor.py
+++ b/traitsui/testing/tester/qt4/implementation/list_editor.py
@@ -9,6 +9,7 @@
 #  Thanks for using Enthought open source!
 #
 from traitsui.testing.tester import command, locator
+from traitsui.testing.tester.common_ui_targets import _BaseSourceWithLocation
 from traitsui.testing.tester.registry_helper import register_nested_ui_solvers
 from traitsui.testing.tester.qt4 import helpers
 from traitsui.qt4.list_editor import (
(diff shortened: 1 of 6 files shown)


traitsui/testing/tester/qt4/implementation/list_editor.py
```diff
@@ -64,49 +43,25 @@ def register(cls, registry):
...
             The registry being registered to.
         """
-        registry.register_solver(
-            target_class=NotebookEditor,
-            locator_class=locator.Index,
-            solver=cls.from_location,
-        )
-
+        super().register(registry)
         register_nested_ui_solvers(
```
kitchoi, line 47: The `solvers` list in `_BaseSourceWithLocation` was removed in #1191 because it wasn't used then. Now we have a use case for it. Would you like to add it back? Then we don't need to worry about overriding the base class method.

aaronayres35: That makes sense to me, but then it is a bit awkward as we wouldn't be able to use `register_nested_ui_solvers` as it is currently written.  Instead we would need to set something like:
```
solver = [(locator.NestedUI, solver ...) , (locator.TargetByName, solver...), (locator.TargetById, ...)]
```
I can think of a few cleaner solutions to this so that `register_nested_ui_solvers` could still be used, but am unsure what would be best.

kitchoi: Offline discussion: I didn't think this through. The `solvers` list as a class variable does not allow us to reuse refactored functions that register many things. The current approach with extending the method in the subclass is correct.

kitchoi (approved):
LGTM.

From our discussion:
We observe that `handlers` on the subclasses are registered by defining a class variable whereas `solvers` are registered by extending the `register` method.  This is asymmetric.

If in the future a new use case arises where we need to extend the `register` method for registering handlers via refactored helper function, we should consider removing the `handlers` class variable such that there is only one way for subclasses to provide additional handlers: extend the `register` method.

The base class and subclasses are not part of the public API so we can afford to do these internal refactoring later.

PR #1221, opened 2020-09-16
Adding cleaned up UI Tester examples for each of the implemented editors

### LINKED ISSUE #1220: UI Tester Examples moved from draft PR into new PR and cleaned up
Currently the UI Tester examples are sitting in a draft branch and could use some cleaning up before being added to the code base.  A new PR should be opened to add these cleaned up examples so that they can be used as a reference in the docs if needed. They are also just useful as a reference for using the tester in general.

aaronayres35:
fixes #1220 
Adds examples for using the UI Tester on the various editors for which there is currently implementations. 

Instance Editor still needs to be added.


traitsui/testing/tester/examples/demo_examples.py
```diff
@@ -0,0 +1,53 @@
...
+        assert output_amount.inspect(query.DisplayedText()) == "1.0"
+
+        for _ in range(3):
+            input_amount.perform(command.KeyClick("Backspace"))
+        input_units.locate(locator.Index(5)).perform(command.MouseClick())
+        output_units.locate(locator.Index(3)).perform(command.MouseClick())
+
+        assert output_amount.inspect(query.DisplayedText())[:3] == "1.6"
+
+        time.sleep(2)
```
aaronayres35, line 53: this is also implemented in https://github.com/enthought/traitsui/blob/master/integrationtests/test_all_examples.py, but it feels appropriate here as a good example of using the tester.  Should this code be removed from test_all_examples, or removed here, or neither?

aaronayres35: In fact, many of the other examples here for UI Tester could be effectively replaced by simply loading the corresponding TraitsUI demo and utilizing the tester on it like is done here.  The examples directly borrow the models/HasTraits classes etc. from the TraitsIO examples

aaronayres35: @kitchoi what are your thoughts about pulling this load_demo function into a module, and then for each example of the UI Tester, we could just load the corresponding TraitsUI demo for the editor, and interact with it via the tester, to show how the tester can be used on each editor.  I think this would be cleaner that how it is currently being done, and it is also seems  nice that the examples are consistent.

aaronayres35: Ah, I started doing this, but there are actually a few annoyances.  For example, in https://github.com/enthought/traitsui/blob/master/traitsui/examples/demo/Standard_Editors/CheckListEditor_simple_demo.py the readonly and custom style editors are both named 'checklist' so using `find_by_name` with the tester does not work.  It seems to make some of the examples more tester friendly it might require adjusting the already existing examples in TraitsUI.  In the most recent commit I made this change to the Button Editor example just for reference as to how this change would look.  If we decide against loading the traitsui examples I can simply revert this commit. 
Going to pause working on this for now until I confirm the best way to proceed.

traitsui/testing/tester/examples/range_example.py
```diff
@@ -0,0 +1,54 @@
...
+    for mode in ['slider', 'xslider', 'logslider', 'text']:
+        with tester.create_ui(model, dict(view=get_view(mode))) as ui:
+            number_field = tester.find_by_name(ui, "value")
+            text = number_field.locate(locator.WidgetType.textbox)
+
+            if is_windows and is_wx() and mode == 'text':
+                # For RangeTextEditor on wx and windows, the textbox
+                # automatically gets focus and the full content is selected.
+                # Insertion point is moved to keep the test consistent
+                text.target.textbox.SetInsertionPointEnd()
```
kitchoi, line 46: If we are exposing this test as a showcase example, I'd very much like to implement `KeyClick("End")` for wx TextEntry.

kitchoi (commented):
This is a tricky one...

If we add these examples in a `if __name__ == "__main__":` block and not run them in CI, then we have the same problem we had before UITester was introduced. We rely on manually running each of these Python scripts to make sure they still work...which happens very very rarely.

This is how some examples in `traitsui/examples/demo` created years ago ended up being broken: No one is running every single one of them regularly across toolkits and platforms whenever we change the library code. This is also how this block becoming broken until you fixed it in #1171:
https://github.com/enthought/traitsui/blob/cbfeeffff58ee7f11c30daa22b2f13779be3917c/traitsui/tests/editors/test_range_editor_text.py#L85-L89
(I'd actually have just removed the block.)

Manual testing cannot, and should not, be completely removed from the testing pipeline, but we should aim at automating as many test logic as possible.

UITester allows projects to move more manual testing logic from their `__name__ == "__main__"` into testing frameworks such that tests are frequently and automatically run by CI. Showing examples in these `__name__ == "__main__"` blocks not exercised in CI seems a bit ironic.

May I suggest that we move these tests to `test_all_examples`, like `TestInteractExample.test_converter`? Then we can point users to them and say "TraitsUI has some more complex application example and this is how automated integration tests are written for them with the UITester".

Having them run in CI means we won't be setting any `delay` in the tests, and this is closer to the expected usage: The `delay` is meant to be a temporary thing for debugging and sanity checking. It can be mentioned in the user manual.

rahulporuri:
> May I suggest that we move these tests to test_all_examples, like TestInteractExample.test_converter? Then we can point users to them and say "TraitsUI has some more complex application example and this is how automated integration tests are written for them with the UITester".

My only recommendation is to leave each editor example in it's own small script instead of putting them all in a single `test_all_examples` script - that way, if we want to show more complicated/advanced examples of how the UITester can be used in the context of this specific editor, we can do so easily in the appropriate context.

rahulporuri:
and also to split up each example into it's own standalone PR so that the work/review/rework/approval can happen in parallel - and take up less real estate in the GitHub PR history.

kitchoi:
Organizing the tests into separate modules is a good idea... one thing that might be a bit awkward is the fact that `integrationtests` is not a package, so refactoring and reusing the `load_demo` function is going to require a bit more tweaking. 

No action: In fact, we might want to consider having the tests in the `examples/demo` directory so that their association with the demos  is more obvious. But that will require a bit more work to tweak `unittest` `discover` to find them without making the `demo` a Python package, so I think I am just thinking out loud.  Expanding logic for testing the demo is already a huge win.

kitchoi:
> No action: In fact, we might want to consider having the tests in the examples/demo directory so that their association with the demos is more obvious. But that will require a bit more work to tweak unittest discover to find them without making the demo a Python package, so I think I am just thinking out loud. Expanding logic for testing the demo is already a huge win.

Forgot to mention: I was thinking having the example test being visible from etsdemo 😉

aaronayres35:
> Organizing the tests into separate modules is a good idea... one thing that might be a bit awkward is the fact that `integrationtests` is not a package, so refactoring and reusing the `load_demo` function is going to require a bit more tweaking.

Just to confirm, are you suggesting the examples go into separate modules within `integration tests`?

aaronayres35:
So in the most recent commit I moved the button editor example into a file called test_button_examples in integrationtests.  I also made it into a unittest test case, rather than using __name__ == "__main__".   The load_demo function will need to be refactored still (it is currently just sitting in testing/tester/examples).  I just wanted to confirm that I am on the same page given the above discussion.  
Basically, we want to:
-remove __name__ == "__main__", and any delays from these "examples" (they are examples of the tester, but tests of the demos)
- move these into their own scripts (one for each editor) inside `integrationtests` for now (ultimately we will want to move them into thee examples/demo directory, but not yet)
- deal with the `load_demo` refactor which is only tricky because `integrationtests` is not a package.

Also, breaking this all up into individual PRs is a good idea.  I am thinking once I confirm the plan with you both, I can extract each example into its own PR and just close this one?

kitchoi:
>> Organizing the tests into separate modules is a good idea... one thing that might be a bit awkward is the fact that integrationtests is not a package, so refactoring and reusing the load_demo function is going to require a bit more tweaking.
>
> Just to confirm, are you suggesting the examples go into separate modules within integrationtests?

I think I was agreeing to splitting the tests into separate modules because it is easier to open separate PRs without causing conflicts, but at the same time I am not in favour of it due to the complexity required for reusing `load_demo` for what it is worth.

My thought process...
- The `load_demo` is only needed for testing demo loaded from scripts in `examples/demo`.  Neither `integrationtests` nor the `examples/demo` folder are packages. It is only used for TraitsUI own CI for maintaining non-package data that contains Python code. It is a utility like `etstool.py`. It is a stretch to place it in `traitsui.testing`, or anywhere in `traitsui` package.
- You cannot add `load_demo` to the `integrationtests` folder and have it importable without fiddling with Python import path (terrible!) because `integrationtests` is not a package.

`integrationtests` has not been meant to be a package, I doubt it is ever meant to be one. If we turn it into a package such that we can add `load_demo` there, we will then have to sanity check `setup.py` to make sure it does not end up being picked up by `setup.py`.

Taking a step back, all of these just for reusing a single function that serves a very specific and rare use case, is it worth it?

Considering if we were to move individual tests for the demo inside `examples/demo`,  so that it is also visible from the etsdemo application, we can configure unittest discover and solve this problem in a different way, maybe with the `load_tests` protocol.

As much as I am in favour of having individual demo's test as separate modules for organization/PR-splitting purposes, it is too much for now just to be able to reuse this `load_demo`.  So I'd rather add them into the single `test_all_examples.py` and use a `load_demo` define there.

@rahulporuri @aaronayres35 What do you think?

aaronayres35:
> @rahulporuri @aaronayres35 What do you think?

Thats fine with me! If in the future we decide we want to move them, we can cross that bridge when we get to it.

aaronayres35:
closing based off the above discussion.  I will open smaller PRs for each individual example

PR #1225, opened 2020-09-17
UI Tester wx mouseclick refactor

### LINKED ISSUE #1194: UI Tester wx helpers for mouse clicks can be refactored
The various mouse clicking functions for wx are all very similar and should be refactored.  See discussion here: https://github.com/enthought/traitsui/pull/1191#discussion_r483716007

aaronayres35:
Fixes #1194 

Introduces new `_create_event` helper function, and `mouse_click` decorator.  
Also, note that the decorator enforces arguments to mouse click functions to be key word and not positional arguments.  When calling these mouse click functions we typically already passed arguments with a keyword, but in a few cases we had passed them by position.  Those have been replaced in this PR.

### DIFF
--- a/traitsui/testing/tester/wx/helpers.py
+++ b/traitsui/testing/tester/wx/helpers.py
@@ -45,7 +45,7 @@ def mouse_click(func):
 
     Parameters
     ----------
-    func : callable
+    func : callable(*, control, delay, **kwargs)
         The mouse click function to be decorated.
 
     Returns
--- a/traitsui/testing/tester/wx/helpers.py
+++ b/traitsui/testing/tester/wx/helpers.py
@@ -16,24 +16,62 @@ from traitsui.testing.tester.exceptions import Disabled
 
 
 def _create_event(event_type, control):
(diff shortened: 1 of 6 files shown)


traitsui/testing/tester/wx/helpers.py
```diff
@@ -15,6 +15,62 @@
...
+
+def mouse_click(func):
+    """ Decorator function for mouse clicks. Decorated functions will return
+    if they are not enabled. Additionally, this handles the delay for the
+    click.
+
+    Parameters
+    ----------
+    func : callable
+        The mouse click function to be decorated.
```
kitchoi, line 49: Nit: Although the returned callable has its signature documented, it would be nice to also specify the required signature for the function being decorated: it must accept, at a minimum, the `control` and `delay` as arguments.  `func : callable(*, control, delay, **kwargs)`. This saves the developer from having to parse the code in order to know how to use this decorator.

(Maybe this was also part of what you were asking about earlier... Sorry about that.)

kitchoi (approved):
LGTM

PR #1226, opened 2020-09-17
Fixing test that was broken by #1209

aaronayres35:
This solves the problem mentioned here: https://github.com/enthought/traitsui/pull/1209#issuecomment-694313188

We did not make a necessary change to a test on #1209 after changing the RangeEditor implementation


kitchoi (approved):
LGTM

PR #1229, opened 2020-09-17
Adding 'End' as a possible value for Key click text entry on wx

### LINKED ISSUE #1228: UI Tester wx Keyclicks do not support "End"
Currently the only supported keys are individual characters, "Backspace", and "Enter".  Adding "End" would be useful for dealing with varying toolkit/editor discrepancies of where the default insertion point is when navigating into a textbox. 
For a use case example see: https://github.com/enthought/traitsui/pull/1221#discussion_r490066759

aaronayres35:
fixes #1228 
simply moves the text entry's insertion point to the end if a key click is called with "End".
Also a simple test for this is added to this PR.


kitchoi (approved):
LGTM!

kitchoi:
(I believe `KeyClick("End")` is already working as expected for Qt.)

PR #1231, opened 2020-09-18
Reraise exception captured when the UI is disposed from create_ui

kitchoi:
The observation in https://github.com/enthought/traitsui/pull/1161#discussion_r476589718 led me to this: `create_ui` context manager should capture and reraise exceptions occurred from the GUI event loop while disposing a UI.

However this does not resolve the strangeness observed in https://github.com/enthought/traitsui/pull/1161#discussion_r476589718
The event that causes the error (which was fixed by #1161) was processed much much later.

### DIFF
--- a/traitsui/testing/tester/tests/test_ui_tester.py
+++ b/traitsui/testing/tester/tests/test_ui_tester.py
@@ -11,6 +11,7 @@
 
 import unittest
 
+from pyface.api import GUI
 from traits.api import (
     Button, Instance, HasTraits, Str,
 )
@@ -56,6 +57,23 @@ class TestUITesterCreateUI(unittest.TestCase):
         with tester.create_ui(order, dict(view=view)) as ui:
             pass
         self.assertTrue(ui.destroyed)
+
+    def test_create_ui_reraise_exception(self):
+        tester = UITester()
+        order = Order()
+        view = View(Item("submit_button"))
+
+        with self.assertRaises(RuntimeError), \
+                self.assertLogs("traitsui", level="ERROR"):
+
+            with tester.create_ui(order, dict(view=view)) as ui:
+
+                def raise_error():
+                    raise ZeroDivisionError()
+
+                GUI().invoke_later(raise_error)
+
+        self.assertIsNone(ui.control)
 
 
 @requires_toolkit([ToolkitName.qt, ToolkitName.wx])
--- a/traitsui/tests/_tools.py
+++ b/traitsui/tests/_tools.py
@@ -232,13 +232,14 @@ def create_ui(object, ui_kwargs=None):
         # At the end of a test, there may be events to be processed.
         # If dispose happens first, those events will be processed after
         # various editor states are removed, causing errors.
-        process_cascade_events()
-        try:
-            ui.dispose()
-        finally:
-            # dispose is not atomic and may push more events to the event
-            # queue. Flush those too.
+        with reraise_exceptions():
             process_cascade_events()
+            try:
+                ui.dispose()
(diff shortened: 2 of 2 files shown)


rahulporuri (approved):
I'm okay with the change but i'm afraid this will break test suites which were silently passing earlier.

kitchoi:
Most tests should have `reraise_exceptions` outside of `create_ui` already, e.g. this:
https://github.com/enthought/traitsui/blob/80a31705d974edd9624ceb34cd3e83a6f643960a/traitsui/tests/editors/test_drop_editor.py#L47-L48

It is, of course, not identical to moving the context manager inside create_ui.  But I think moving it inside create_ui is more correct. We should only be using the context manager in tests where we are asking the GUI event loop to process events. Unfortunately the `reraise_exceptions`  (used to be called `store_exceptions_on_all_threads`) historically been used for more than just that, it also captures exceptions from trait change (see #1232).

It has been a gradual process to add `reraise_exceptions` in tests and to ensure the UIs are disposed.

traitsui/testing/tester/tests/test_ui_tester.py
```diff
@@ -57,6 +58,23 @@ def test_ui_disposed(self):
...
+
+        with self.assertRaises(RuntimeError), \
+                self.assertLogs("traitsui", level="ERROR"):
+
+            with tester.create_ui(order, dict(view=view)) as ui:
+
+                def raise_error():
+                    raise ZeroDivisionError()
+
+                GUI().invoke_later(raise_error)
```
aaronayres35, line 74: you say in the PR description: "create_ui context manager should capture and reraise exceptions occurred from the GUI event loop while disposing a UI."  
Just to clarify, because `raise_error` raises a `ZeroDivisionError` but we have `with self.assertRaises(RuntimeError)` we know that the raise_error ultimately triggered the `reraise_exceptions` that was added in the file below (as that 'captures all exceptions occurred in the context and then reraises a RuntimeError if there are any exceptions captured.').  

If thats true this all LGTM, I just want to make sure Im understanding correctly.
Also, do we know anything specific about when a call to `.invoke_later()` will be processed (aside from that it will be later)?

aaronayres35 (approved):
LGTM! Just a question for my own understanding

traitsui/testing/tester/tests/test_ui_tester.py
kitchoi: > Just to clarify, because raise_error raises a ZeroDivisionError but we have with self.assertRaises(RuntimeError) we know that the raise_error ultimately triggered the reraise_exceptions that was added in the file below (as that 'captures all exceptions occurred in the context and then reraises a RuntimeError if there are any exceptions captured.').

Yes that is correct.
Always reraising a new RuntimeError is intentional, as the opposite of reraising whatever is captured is problematic in this situation. See https://github.com/enthought/traitsui/pull/1099

> Also, do we know anything specific about when a call to .invoke_later() will be processed (aside from that it will be later)?

The callable is queued to be processed by the GUI event loop. So given the way I have written this test, it should occur from the first `process_cascade_events` in the `finally` block. 
If I mock the UI.dispose to call `invoke_later(raise_error)`, then the error will occur from the second `process_cascade_events` in the `finally`. Mocking `UI.dispose` is a bit harder because `UI` is an instance of `HasStrictTraits` and it is not as trivial to mock its method compared to mocking any methods of a plain simple Python object.

kitchoi:
Tests for Qt are passing, and the allowed test failures for Wx are the same as before. Merging.

PR #1235, opened 2020-09-18
Rename register_nested_ui_solvers to register_traitsui_ui_solvers

kitchoi:
Follow this https://github.com/enthought/traitsui/pull/1227#discussion_r490837104

This PR renames `register_nested_ui_solvers` to `register_traitsui_ui_solvers`
The function does not care if the `traitsui.ui.UI` instance is nested or not. The name is slightly long, but verbosity for understanding trumps ~conciseness~.

Note: Could not use the name `register_ui_solvers` here because the function is only about `traitsui.ui.UI`. The word "UI" is too broad and is used in many other context.


aaronayres35 (approved):
LGTM!

PR #1237, opened 2020-09-18
adding UI Tester test/example for ButtonEditor_simple_demo.py

aaronayres35:
This is the first "example" (which is actually a test of a traitsui example/demo) to be added for UI Tester. 
Addresses part of #1220

### DIFF
--- a/integrationtests/test_all_examples.py
+++ b/integrationtests/test_all_examples.py
@@ -346,11 +346,26 @@ class TestExample(unittest.TestCase):
                     )
                 )
 
-
+@requires_toolkit([ToolkitName.qt, ToolkitName.wx])
 class TestInteractExample(unittest.TestCase):
     """ Test examples with more interactions."""
 
-    @requires_toolkit([ToolkitName.qt, ToolkitName.wx])
+    def test_button_editor_simple_demo(self):
+        # Test ButtonEditor_simple_demo.py in examples/demo/Standard_Edtiors
+        filepath = os.path.join(
+            DEMO, "Standard_Editors", "ButtonEditor_simple_demo.py"
+        )
+        demo = load_demo(filepath, "demo")
+
+        tester = UITester()
+        with tester.create_ui(demo) as ui:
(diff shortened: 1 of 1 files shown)


integrationtests/test_all_examples.py
```diff
@@ -346,11 +346,26 @@ def test_run(self):
...
+        )
+        demo = load_demo(filepath, "demo")
+
+        tester = UITester()
+        with tester.create_ui(demo) as ui:
+            button = tester.find_by_name(ui, "my_button_trait")
+            for _ in range(5):
+                button.perform(command.MouseClick())
+            click_counter = tester.find_by_name(ui, "click_counter")
+            displayed_count = click_counter.inspect(query.DisplayedText())
```
aaronayres35, line 366: There is still the issue on OSX/qt where the click_counter does not actually display a change, however this test still passes... Seems a little misleading given the test, but I think the tester, and query.DisplayedText() are generally working how we want them to.

integrationtests/test_all_examples.py
```diff
@@ -346,11 +346,26 @@ def test_run(self):
...
+        filepath = os.path.join(
+            DEMO, "Standard_Editors", "ButtonEditor_simple_demo.py"
+        )
+        demo = load_demo(filepath, "demo")
+
+        tester = UITester()
+        with tester.create_ui(demo) as ui:
+            button = tester.find_by_name(ui, "my_button_trait")
+            for _ in range(5):
+                button.perform(command.MouseClick())
```
kitchoi, line 364: I'd suggest adding `self.assertEqual(demo.click_counter, index + 1)` following the mouse click.
This demonstrates one of the two common testing patterns: Modify GUI state resulting in trait changes

Likewise, adding `demo.click_counter = 10` (deliberately not 5) and then inspect the displayed value is updated to "10" demonstrates the second of the two common testing patterns: Modify trait updates the GUI state.

kitchoi (commented):
LGTM. Some suggestions.

kitchoi (approved):
LGTM

PR #1239, opened 2020-09-18
adding UI Tester test/example for ButtonEditor_demo.py

aaronayres35:
This adds on to what was done in PR #1237 with another example.  
This PR also showcases how traitsui's UI Tester can work with pyface's ModalDialogTester.  
I am unsure if how I utilized them together here was the best way to do so.  I wanted to test that clicking the buttons in the original UI did trigger the Modal Dialog to popup.  That is why I defined the functions that I did and used those to instantiate the ModalDialogTester.
Originally I had tried using `demo._fire_event_fired`, or `message = load_demo(filepath, "message")` as the functions to instantiate the ModalDialogTester as these functions both open the dialog.  (`message` is really the function doing this, and its returned `ui.result` is what we would expect such a function to return - what we would expect to appear as the ModalDialogTester instance's `_result` attribute).  But the way I did it made more sense to me for the purposes of this test. 

This may very well not be the correct way to do it though, so any feedback is very welcome!

### DIFF
--- a/integrationtests/test_all_examples.py
+++ b/integrationtests/test_all_examples.py
@@ -304,7 +304,6 @@ def load_demo(file_path, variable_name="demo"):
         content = f.read()
     globals_ = globals().copy()
     exec(content, globals_)
-    #print(globals_)
     return globals_[variable_name]
 
 
--- a/integrationtests/test_all_examples.py
+++ b/integrationtests/test_all_examples.py
@@ -304,6 +304,7 @@ def load_demo(file_path, variable_name="demo"):
         content = f.read()
     globals_ = globals().copy()
     exec(content, globals_)
+    #print(globals_)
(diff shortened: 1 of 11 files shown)


kitchoi (approved):
LVGTM!

integrationtests/test_all_examples.py
```diff
@@ -371,7 +404,7 @@ def test_button_editor_simple_demo(self):
             demo.click_counter = 10
             displayed_count = click_counter.inspect(query.DisplayedText())
             self.assertEqual(displayed_count, '10')
-
```
kitchoi, line 374: Looks like we can remove `integrationtests/test_all_examples.py` from `setup.cfg` after this change. Then CI will help us keeping this file pep8-clean. Not a request change. But feel free to do as a drive-by.

PR #1240, opened 2020-09-18
Reverse dependency so that traitsui.tests._tools depend on traitsui.testing

### LINKED ISSUE #1238: Make traitsui.tests._tools depend on traitsui.testing
Currently a few functions in `traitsui.tests._tools` are imported by `traitsui.testing`:
`create_ui`
`reraise_exceptions`
`process_cascade_events`

This should be fixed.

My understanding is that `traitsui.tests._tools` is entirely private for supporting traitsui's own test code.  `traitsui.testing` not only supports traitsui's own test code, it is also intended to be a public library. 

The problem with the current dependency direction is that one may think they can change any of the above functions without affecting downstream projects.

The fix is to move the above functions into the `traitsui.testing` package and have `traitsui.tests._tools` import from the `testing` package if needed.

kitchoi:
Closes #1238

- The implementation of `traitsui.tests._tools.create_ui` is now moved to `UITester.create_ui`. The `traitsui.tests._tools.create_ui` is kept to limit the scope of this PR (otherwise I need to change a lot of test)
- `traitsui.tests._tools.process_cascade_events` is moved to `traitsui.testing.gui`. Its test is also moved. Likewise, the alias in `_tools` is kept, for now to limit scope.
- `traitsui.tests._tools.reraise_exceptions` is moved to `traitsui.testing.exception_handling`. Its test is also moved. Likewise, the alias in `_tools` is kept, for now to limit scope.

### DIFF
--- a/traitsui/testing/gui.py
+++ b/traitsui/testing/gui.py
@@ -1,0 +1,40 @@
+#  Copyright (c) 2005-2020, Enthought, Inc.
+#  All rights reserved.
+#
+#  This software is provided without warranty under the terms of the BSD
+#  license included in LICENSE.txt and may be redistributed only
+#  under the conditions described in the aforementioned license.  The license
+#  is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+#  Thanks for using Enthought open source!
+#
+""" Module containing GUI event processing utility for testing purposes.
+"""
+
+from traits.etsconfig.api import ETSConfig
+
+
+def process_cascade_events():
+    """ Process all posted events, and attempt to process new events posted by
+    the processed events.
+
+    Cautions:
+    - An infinite cascade of events will cause this function to enter an
+      infinite loop.
+    - There still exists technical difficulties with Qt. On Qt4 + OSX,
+      QEventLoop.processEvents may report false saying it had found no events
+      to process even though it actually had processed some.
+      Consequently the internal loop breaks too early such that there are
+      still cascaded events unprocessed. Problems are also observed on
+      Qt5 + Appveyor occasionally. At the very least, events that are already
+      posted prior to calling this function will be processed.
+      See enthought/traitsui#951
+    """
+    if ETSConfig.toolkit.startswith("qt"):
+        from pyface.qt import QtCore
+        event_loop = QtCore.QEventLoop()
+        while event_loop.processEvents(QtCore.QEventLoop.AllEvents):
+            pass
+    else:
+        from pyface.api import GUI
+        GUI.process_events()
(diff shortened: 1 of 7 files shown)


traitsui/testing/gui.py
```diff
@@ -0,0 +1,40 @@
...
+    - There still exists technical difficulties with Qt. On Qt4 + OSX,
+      QEventLoop.processEvents may report false saying it had found no events
+      to process even though it actually had processed some.
+      Consequently the internal loop breaks too early such that there are
+      still cascaded events unprocessed. Problems are also observed on
+      Qt5 + Appveyor occasionally. At the very least, events that are already
+      posted prior to calling this function will be processed.
+      See enthought/traitsui#951
+    """
+    if ETSConfig.toolkit.startswith("qt"):
```
aaronayres35, line 33: what is the plan for where `is_qt`, `is_wx` etc. will reside?

traitsui/testing/tests/test_exception_handling.py
```diff
@@ -0,0 +1,74 @@
...
+""" Tests for traitsui.testing.exception_handling """
+
+import unittest
+
+from pyface.api import GUI
+
+from traits.api import HasTraits, Int
+from traitsui.tests._tools import (
+    requires_toolkit,
+    ToolkitName,
```
aaronayres35, line 21: will these eventually be moved into `traitsui.testing`?  I feel like they don't need to if they aren't a dependency outside of the tests in `traitsui.testing`, correct?

aaronayres35 (approved):
LGTM, just a couple quick questions - not suggested changes.

traitsui/testing/gui.py
aaronayres35: Edit: I believe they are going to stay where they are and that makes sense to me.

traitsui/testing/tests/test_exception_handling.py
kitchoi: > will these eventually be moved into traitsui.testing?

I am not planning to.

I think I want to think of `traitsui.testing` being used by both traitsui AND projects using of traitsui.
requires_toolkit and ToolkitName are utilities only used by traitsui's own tests, so by keeping them out of the way of `traitsui.testing`, it is more obvious to developers that we can change them without worrying about affecting downstream projects in any way.

That is why I did not reuse `is_qt` in `traitsui.testing.gui`. It is such a teeny tiny function that does not warrant the additional coupling.

>  I feel like they don't need to if they aren't a dependency outside of the tests in traitsui.testing, correct?

I am not sure if I understand the question fully. Perhaps part of this was my choice of wording in #1238
In #1238, I wrote:
> Currently a few functions in traitsui.tests._tools are imported by traitsui.testing:

When I wrote "imported by `traitsui.testing`", I have in my mind excluded `traitsui.testing.*.tests.*`.
My concern in #1238 was mainly about having externally facing testing library behaviour being implemented in a private module that is obviously internally facing. `traitsui.testing.*.tests.*` are not library, so the concern does not apply there.

PR #1241, opened 2020-09-18
adding UI Tester test/example for CheckListEditor_simple_demo.py

aaronayres35:
addresses part of #1220 

This PR, in addition to adding the UI Tester test/example for CheckListEditor_simple_demo.py, modifies the `mouse_click_qwidget` function to address a test failure observed here and in other places.  (see comments).  It fixed the problem by handling clicking of objects that subclass `QAbstractButton` separately.

### DIFF
--- a/traitsui/testing/tester/qt4/helpers.py
+++ b/traitsui/testing/tester/qt4/helpers.py
@@ -8,6 +8,8 @@
 #
 #  Thanks for using Enthought open source!
 #
+import time 
+
 from pyface.qt import QtCore, QtGui
 from pyface.qt.QtTest import QTest
 
@@ -105,11 +107,15 @@ def mouse_click_qwidget(control, delay):
     delay : int
         Time delay (in ms) in which click will be performed.
     """
-    QTest.mouseClick(
-        control,
-        QtCore.Qt.LeftButton,
-        delay=delay,
-    )
+    if isinstance(control, QtGui.QAbstractButton):
+        time.sleep(delay/1000)
+        control.click()
+    else:
+        QTest.mouseClick(
+            control,
+            QtCore.Qt.LeftButton,
+            delay=delay,
+        )
 
 
 def mouse_click_tab_index(tab_widget, index, delay):
@@ -150,11 +156,7 @@ def mouse_click_qlayout(layout, index, delay):
     if not 0 <= index < layout.count():
         raise IndexError(index)
     widget = layout.itemAt(index).widget()
-    QTest.mouseClick(
-        widget,
-        QtCore.Qt.LeftButton,
-        delay=delay,
-    )
+    mouse_click_qwidget(widget, delay)
 
 
 def mouse_click_item_view(model, view, index, delay):
--- a/traitsui/testing/tester/qt4/helpers.py
+++ b/traitsui/testing/tester/qt4/helpers.py
@@ -8,7 +8,7 @@
 #
 #  Thanks for using Enthought open source!
 #
-import time 
+import time
 
 from pyface.qt import QtCore, QtGui
 from pyface.qt.QtTest import QTest
--- a/traitsui/testing/tester/qt4/helpers.py
+++ b/traitsui/testing/tester/qt4/helpers.py
@@ -8,8 +8,6 @@
 #
 #  Thanks for using Enthought open source!
 #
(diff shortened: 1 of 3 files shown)


aaronayres35:
Calling `QTest.mouseClick` defaults to clicking the center of the widget. Calling `widget.rect()` gave a result of `QRect(0, 0, 214, 20)`.  Thus the mouse click occurred at (107, 10).  By hardcoding where to click, i.e. by running 
```
QTest.mouseClick(
        widget,
        QtCore.Qt.LeftButton,
        delay=delay,
        pos=QtCore.QPoint(10, 10)
    )
```
the failing test passes, so it seems that specifying the correct position to be clicked fixes this issue.  

I'm currently trying to identify how to do this in a non hardcoded way

aaronayres35:
> I'm currently trying to identify how to do this in a non hardcoded way

Rather than specifying the correct position to click, I simply handled clicking subclasses of `QAbstractButton` separately by calling the `click` method.  Because of the way the code is written (in that we use very generic helper functions like `mouse_click_qwidget` and `mouse_click_qlayout`) it is a bit awkward to handle different cases, like the QCheckbox here, specifically.  I could have made a separate `mouse_click_qbutton` function, but that also would be a bit awkward, so I went with the solution here.  I'm happy to discuss alternatives though.

traitsui/testing/tester/qt4/helpers.py
```diff
@@ -105,11 +107,15 @@ def mouse_click_qwidget(control, delay):
     delay : int
         Time delay (in ms) in which click will be performed.
     """
-    QTest.mouseClick(
-        control,
-        QtCore.Qt.LeftButton,
-        delay=delay,
-    )
+    if isinstance(control, QtGui.QAbstractButton):
+        time.sleep(delay/1000)
```
aaronayres35, line 111: I'm not a huge fan of this `isinstance` check, or the use of time.sleep here.  At the time of writing this they felt like the best solution.

traitsui/testing/tester/qt4/helpers.py
```diff
@@ -150,11 +156,7 @@ def mouse_click_qlayout(layout, index, delay):
     if not 0 <= index < layout.count():
         raise IndexError(index)
     widget = layout.itemAt(index).widget()
-    QTest.mouseClick(
-        widget,
-        QtCore.Qt.LeftButton,
-        delay=delay,
-    )
+    mouse_click_qwidget(widget, delay)
```
aaronayres35, line 159: one reason it would be a bit annoying to have instead made a separate `mouse_click_qbutton` function is that in places like here we may have a layout that contains buttons, or a layout containing other widgets.  In that cases here one would in theory need to have the isinstance check I added in `mouse_click_q_widget`.  
currently `mouse_click_qlayout` is only used for enum_editor and checklist_editor, and in both of those cases the layout is filled with buttons.  So right now we could just call a new `mouse_click_qbutton` function here, but I wasn't sure if there might be a use case for this function in which the layout wasn't populated with button.

kitchoi: I'd suggest refactoring the first three lines that returns a widget (which occurs in a number of places), then  in the implementation of the CheckListEditor where it knows for sure the widget is a checkbox, call the refactored function to obtain the widget and then call the `click` method.
The code path we need for this handler needs to know it is dealing with checkboxes, the only reasonable place for that information to go it in the CheckListEditor.

kitchoi: Ah no, scrap that. I did not see the `if isinstance(control, QtGui.QAbstractButton):` change above.

traitsui/testing/tester/qt4/helpers.py
kitchoi: `QTest.qSleep` would be better which is designed for this (you might need to check the delay is larger than 0).
The `isinstance(control, QtGui.QAbstractButton)` is okay because we are checking against an interface, not a concrete implementation. We might want to comment on why we are not using `mouseClick` directly: `mouseClick` assumes the center of the widget, which is not the appropriate default location when the widget is a button.

kitchoi (approved):
LGTM.
You might also want to update the main description as this no longer fails on Qt.

PR #1242, opened 2020-09-18
replacing WidgetType enum members with standalone types

### LINKED ISSUE #1233: Replace WidgetType enum members with standalone types
For the use case of RangeEditor, we allow test code to locate a subwidget by its type, using this enum:
https://github.com/enthought/traitsui/blob/734e1037986cacc9e158e98bd885c7ab4b167d3b/traitsui/testing/tester/locator.py#L67-L80

While this is good enough for what we needed for RangeEditor, it assumes a subwidget can be uniquely identified by type. e.g. If there are multiple subwidgets of the same type, just an enum member will not be good enough.

Proposed action is to split up `WidgetType.textbox` and `WidgetType.slider` into their own classes (they can be empty for now) so we keep the options of extension open.

aaronayres35:
fixes #1233 
Removes the `locator.WidgetType` enum class, and replaces it  with more flexible standalone classes for each widget type.

### DIFF
--- a/traitsui/testing/tester/locator.py
+++ b/traitsui/testing/tester/locator.py
@@ -61,6 +61,6 @@ class Slider:
 
 
 class Textbox:
-    """ A locator for locating amnested textbox widget within a UI.
+    """ A locator for locating a nested textbox widget within a UI.
     """
     pass
--- a/traitsui/testing/tester/locator.py
+++ b/traitsui/testing/tester/locator.py
@@ -56,22 +56,6 @@ class TargetById:
         self.id = id
 
 
-class WidgetType(enum.Enum):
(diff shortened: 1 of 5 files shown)


kitchoi (approved):
LGTM

traitsui/testing/tester/locator.py
```diff
@@ -56,17 +54,13 @@ def __init__(self, id):
...
     """
+    pass
 
-    # A textbox within a UI
-    textbox = "textbox"
 
-    # A slider within a UI
-    slider = "slider"
+class Textbox:
+    """ A locator for locating amnested textbox widget within a UI.
```
kitchoi, line 64: Oops. Typo:
```suggestion
    """ A locator for locating a nested textbox widget within a UI.
```

PR #1243, opened 2020-09-18
adding UI Tester test/example for EnumEditor_demo.py

aaronayres35:
Addresses part of #1220 
Note that in order to fully test this demo, I needed to add trivial implementations for `traitsui.wx.editor_factory.TextEditor' and `traitsui.wx.editor_factory.ReadonlyEditor' (likewise for qt).


aaronayres35:
Just seeing this CI failure now, tests pass on my machine.  I may look at this a little today, if not I will investigate tomorrow morning.  PR is not really ready for review now with this failure, but feel free to review whenever is most convenient

traitsui/testing/tester/wx/default_registry.py
```diff
@@ -14,6 +14,8 @@
 from traitsui.testing.tester.wx.implementation import (
     button_editor,
     check_list_editor,
+    editor_factory_readonly_editor,
+    editor_factory_text_editor,
```
kitchoi, line 18: Nit: I think I'd prefer the module matching the that in `traitsui.qt4` so that it is easier to find them. This is a bit like the convention of naming test files according to the module being tested. But then there are always exceptions to the rules. Is there a reason for these two to be separated?

traitsui/testing/tester/wx/implementation/editor_factory_readonly_editor.py
```diff
@@ -0,0 +1,59 @@
...
+    other times use wx.StaticText.
+    '''
+    if isinstance(wrapper.target.control, wx.TextCtrl):
+        return wrapper.target.control.GetValue()
+    elif isinstance(wrapper.target.control, wx.StaticText):
+        return wrapper.target.control.GetLabel()
+    raise TypeError("readonly_displayed_text_handler expected a UIWrapper with"
+                    " a ReadonlyEditor as a target. ReadonlyEditor control"
+                    " should always be either wx.TextCtrl, or wx.StaticText."
+                    " {} was found".format(wrapper.target.control))
```
kitchoi, line 42: I think I have seen this function before that is close to identical. Is it time to refactor?

traitsui/testing/tester/qt4/helpers.py
```diff
@@ -105,11 +107,15 @@ def mouse_click_qwidget(control, delay):
     delay : int
         Time delay (in ms) in which click will be performed.
     """
-    QTest.mouseClick(
-        control,
-        QtCore.Qt.LeftButton,
-        delay=delay,
-    )
+    if isinstance(control, QtGui.QAbstractButton):
+        time.sleep(delay/1000)
```
kitchoi, line 111:
```
if delay > 0:
    QTest.qSleep(delay)
```
might be cleaner? We are asking Qt to sleep rather than asking Python to sleep.

integrationtests/test_all_examples.py
```diff
@@ -429,3 +429,57 @@ def test_converter(self):
...
+            selected_list = list_enum.inspect(query.SelectedText())
+            displayed_text = text.inspect(query.DisplayedText())
+            displayed_readonly = readonly.inspect(query.DisplayedText())
+
+            displayed_selected = [
+                displayed_simple, disp_simple_text, selected_radio,
+                selected_list, displayed_text, displayed_readonly
+            ]
+            for _ in displayed_selected:
+                self.assertEqual(_, 'Foo')
```
kitchoi, line 485: Nit: `_` as a variable name is by convention used for unused variable, i.e. it is understood as a placeholder. Here the variable is used, so we should probably name it, otherwise it reads a bit funny for those who are used to the convention.

kitchoi (commented):
Mostly LGTM!  Some minor comments.

traitsui/testing/tester/wx/default_registry.py
aaronayres35: > Is there a reason for these two to be separated?

Not a good one, I will make the change to mirror that of `traitsui.qt4` as that makes more sense.

traitsui/testing/tester/wx/implementation/editor_factory_readonly_editor.py
aaronayres35: It is exactly identical (even up to the name) as the function in the text_editor implementation.  Yes, I agree it should be pulled out somewhere and used in both places.  I'm unsure where though.  Of the current options maybe the wx `registry_helper` makes most sense?  It could also go in its own new module (but we do have a lot of modules with only one function or class in them already).  What do you think?

kitchoi: Since this function only uses `wrapper.target.control` and does not use `interaction`, it might fit in the more `helpers` module instead?

aaronayres35: Yeah I think i agree.  I originally had thought helpers seemed incorrect as currently those functions are all directly related to mouse clicks/ key clicks, but they are basically just functions that are called when `.perform()` is called.  It makes sense that this would be there as well as it will be called when `.inspect()` is called.

traitsui/testing/tester/wx/helpers.py
```diff
@@ -391,3 +391,30 @@ def key_click_slider(control, interaction, delay):
         wx.wxEVT_SCROLL_CHANGED, control.GetId(), position
     )
     wx.PostEvent(control, event)
+
+
+def readonly_displayed_text_handler(wrapper, interaction):
```
kitchoi, line 396: The function does not need to know about ReadonlyEditor or query.DisplayedText or UIWrapper. This allows the functions to be more easily reused in other context. e.g. All you need to call this function is a wx.TextCtrl or wx.StaticText. If the widget is not placed on `wrapper.target.control`, but somewhere else, you can still reuse this function.

So I'd probably change the argument to just `(control)` (the wx widget), especially since the `interaction` isn't used at all. The function can be renamed such that it is more obvious about that the fact it is serving a read-only textbox. In the case of wx, one can do so via wx.TextCtrl or wx.StaticText. This function serves to normalize the lack of common interface between the two.

kitchoi (approved):
LGTM

Issue #1223, opened 2020-09-17
User manual / guide for UITester


kitchoi:
This issue is for extending the Sphinx documentation for developers _using_ UITester.

Note that some of that has been written in the `UITester` class docstring, but there is still room to extend. In particular, the documentation should
(1) Guide how to find supported locations and interactions (related to #1211)
(2) Provide an example or some examples for testing a complex application in the context of an integration test.

Part of #1173

PR #1246, opened 2020-09-21
Add help method to UIWrapper for self-discovery

### LINKED ISSUE #1211: Allow self discovery of supported locations and interactions on UIWrapper
To test a particular UI component using UITester, one needs to know what objects can be passed to the `UIWrapper.locate` for locating elements and `UIWrapper.perform` (or `UIWrapper.inspect`) for performing user interactions. These objects vary from context to context. Currently there isn't a way to know what are supported apart from trying a value and getting an exception with the supported values included in the error message... that is quite awkward to use!

It is perhaps best for a library to be self-documenting. While it would also be useful to maintain a laundry list of these supported locations/interactions in the Sphinx documentation per-editor, it is hard to maintain and keep up-to-date. Furthermore, the UITester can be extended to support more custom locations and interactions (via `TargetRegistry`). Static documentation would not be able to help with discovering these custom extensions.

Proposal:
- Add `UIWrapper.supported_locations` property (or a query method) to return a list of supported locations, as are accepted by the `locate` method.
- Add `UIWrapper.supported_interactions` property (or a query method) to return a list of supported interactions as accepted by the `perform` and `inspect` methods.

Goal:  The API will be self-documenting and one will then be able to explore functionalities either by printing from a half-written test or from a Python prompt:
```
>>> ui = .... # created somehow, maybe HasTriaits.edit_traits
>>> tester = UITester()
>>> wrapper = tester.find_by_name(ui, "name")
>>> wrapper.supported_interactions    # or something like that.
[<class 'traitsui.testing.tester.command.MouseClick'>]
```
Then one can follow the documentation of the location / interaction objects that are already there.

kitchoi:
Closes #1211

Instead of implementing properties on `UIWrapper`, this PR implements a `help` method to print help messages when it is invoked.

Its current use case is only for developers to discover supported features while they are writing tests.

e.g. if I add the following in this [test](https://github.com/enthought/traitsui/blob/1b2a3933c5313f1d4d2084f197bdb6fc8e4c1ecf/traitsui/tests/editors/test_list_editor.py#L122-L128):
```
            print("Help on the list editor")
            list_.help()

            print("Help on an item in the list editor")
            list_.locate(locator.Index(1)).help()
```

Then I see this being printed:
<details>
    <summary> printed content </summary>

```
Help on the list editor
Interactions
------------
No interactions are supported.

Locations
---------
<class 'traitsui.testing.tester.locator.Index'>
    A locator for locating a target that is uniquely specified by a single
    0-based index.
    
    Attributes
    ----------
    index : int
        0-based index

Help on an item in the list editor
Interactions
------------
<class 'traitsui.testing.tester.command.MouseClick'>
    An object representing the user clicking a mouse button.
    Currently the left mouse button is assumed.
    
    In most circumstances, a widget can still be clicked on even if it is
    disabled. Therefore unlike key events, if the widget is disabled,
    implementations should not raise an exception.

Locations
---------
<class 'traitsui.testing.tester.locator.TargetById'>
    A locator for locating the next UI target using an id.
    
    Attributes
    ----------
    id : str

<class 'traitsui.testing.tester.locator.TargetByName'>
    A locator for locating the next UI target using a name.
    
    Attributes
    ----------
    name : str
```

</details>

I plan to use this feature on the user manual (see #1223)

### DIFF
--- a/traitsui/testing/tester/registry.py
+++ b/traitsui/testing/tester/registry.py
@@ -9,6 +9,8 @@
 #  Thanks for using Enthought open source!
 #
 
+import inspect
+
 from traitsui.testing.tester.exceptions import (
     InteractionNotSupported,
     LocationNotSupported,
@@ -54,6 +56,20 @@ class _TargetToKeyRegistry:
                 available_keys=list(action_to_handler),
             )
         return action_to_handler[key]
+
+    def get_keys(self, target_class):
+        """ Return all the keys for the given target.
+
+        Parameters
+        ----------
+        target_class : subclass of type
+            The type of a UI target being operated on.
+
+        Returns
+        -------
+        keys : set
+        """
+        return set(self._target_to_key_to_value.get(target_class, []))
 
 
 class TargetRegistry:
@@ -215,11 +231,55 @@ class TargetRegistry:
         handler : callable(UIWrapper, interaction) -> any
             The function to handle the particular interaction on a target.
             ``interaction`` should be an instance of ``interaction_class``.
+
+        Raises
+        ------
+        InteractionNotSupported
+            If the given target and interaction types are not supported
+            by this registry.
         """
         return self._interaction_registry.get_value(
             target_class=target_class,
             key=interaction_class,
         )
+
+    def get_interactions(self, target_class):
+        """ Returns all the interactions supported for the given target type.
+
+        Parameters
+        ----------
+        target_class : subclass of type
+            The type of a UI target being operated on.
+
+        Returns
+        -------
+        interaction_classes : set
+            Supported interaction types for the given target type.
+        """
+        return self._interaction_registry.get_keys(target_class=target_class)
+
+    def get_interaction_doc(self, target_class, interaction_class):
+        """ Return the documentation for the given target and locator type.
+
+        Parameters
+        ----------
+        target_class : subclass of type
+            The type of a UI target being operated on.
+        interaction_class : subclass of type
+            Any class.
+
+        Raises
+        ------
+        InteractionNotSupported
+            If the given target and interaction types are not supported
+            by this registry.
+        """
+        self._interaction_registry.get_value(
+            target_class=target_class,
+            key=interaction_class,
+        )
+        # This maybe configurable in the future via register_handler
+        return inspect.getdoc(interaction_class)
 
     def register_solver(self, target_class, locator_class, solver):
         """ Register a solver for resolving the next UI target for the given
@@ -266,3 +326,39 @@ class TargetRegistry:
             target_class=target_class,
             key=locator_class,
         )
+
+    def get_locations(self, target_class):
+        """ Returns all the location types supported for the given target type.
+
+        Parameters
+        ----------
+        target_class : subclass of type
+            The type of a UI target being operated on.
+
+        Returns
+        -------
+        locators_classes : set
+            Supported locator types for the given target type.
+        """
+        return self._location_registry.get_keys(target_class=target_class)
+
+    def get_location_doc(self, target_class, locator_class):
+        """ Return the documentation for the given target and locator type.
+
+        Parameters
+        ----------
+        target_class : subclass of type
+            The type of a UI target being operated on.
+        locator_class : subclass of type
+            Any class.
+
+        Raises
+        ------
+        LocationNotSupported
+        """
+        self._location_registry.get_value(
+            target_class=target_class,
+            key=locator_class,
+        )
+        # This maybe configurable in the future via register_solver
+        return inspect.getdoc(locator_class)
--- a/traitsui/testing/tester/registry.py
+++ b/traitsui/testing/tester/registry.py
@@ -268,6 +268,10 @@ class TargetRegistry:
         interaction_class : subclass of type
             Any class.
 
+        Returns
+        -------
+        doc : str
+
         Raises
         ------
         InteractionNotSupported
@@ -321,6 +325,7 @@ class TargetRegistry:
         Raises
         ------
         LocationNotSupported
+            If the given locator and target types are not supported.
         """
         return self._location_registry.get_value(
             target_class=target_class,
@@ -352,9 +357,14 @@ class TargetRegistry:
         locator_class : subclass of type
             Any class.
 
+        Returns
+        -------
+        doc : str
+
         Raises
         ------
         LocationNotSupported
+            If the given locator and target types are not supported.
         """
         self._location_registry.get_value(
             target_class=target_class,
--- a/traitsui/testing/tester/tests/test_registry.py
+++ b/traitsui/testing/tester/tests/test_registry.py
@@ -108,7 +108,7 @@ class TestInteractionRegistry(unittest.TestCase):
         with self.assertRaises(ValueError):
             registry.register_handler(SpecificEditor, UserAction, handler)
 
-    def test_error_get_interaction_coc(self):
+    def test_error_get_interaction_doc(self):
         # The registry is empty
         registry = TargetRegistry()
         with self.assertRaises(InteractionNotSupported):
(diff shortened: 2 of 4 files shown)


traitsui/testing/tester/registry.py
```diff
@@ -215,12 +231,60 @@ def get_handler(self, target_class, interaction_class):
...
+        ------
+        InteractionNotSupported
+            If the given target and interaction types are not supported
+            by this registry.
+        """
+        self._interaction_registry.get_value(
+            target_class=target_class,
+            key=interaction_class,
+        )
+        # This maybe configurable in the future via register_handler
```
kitchoi, line 285: I was thinking of adding an optional argument `help` to `register_handler` so that implementations can modify the help message with more context specific documentation. e.g. For the slider in RangeEditor, we can document the fact that the `KeyClick` command only supports "Up", "Down", ... keys. But that would be a nice-to-have. Getting the documentation from the class is not so bad - the generic description is already quite useful.

Similarly for `register_solver` for registering location solver.
(I can remove these comments too if the reviewer finds them out-of-place.)

traitsui/testing/tester/ui_wrapper.py
```diff
@@ -73,6 +73,44 @@ def __init__(self, target, registries, delay=0):
         self._registries = registries
         self.delay = delay
 
+    def help(self):
+        """ Print help messages.
+        (This function is intended for interactive use.)
+        """
+        interactions = dict()
+        locations = dict()
+        for registry in self._registries[::-1]:
```
rahulporuri, line 82: im not sure why we are iterating in reverse order

traitsui/testing/tester/ui_wrapper.py
```diff
@@ -73,6 +73,44 @@ def __init__(self, target, registries, delay=0):
...
+                interactions[type_] = registry.get_interaction_doc(
+                    self.target.__class__, type_)
+
+            for type_ in registry.get_locations(self.target.__class__):
+                locations[type_] = registry.get_location_doc(
+                    self.target.__class__, type_)
+
+        print("Interactions")
+        print("------------")
+        pairs = sorted(interactions.items(), key=lambda value: repr(value[0]))
```
rahulporuri, line 93: not sure why we are sorting by `repr(value[0])` or what `repr(value[0])` looks like.

traitsui/testing/tester/ui_wrapper.py
```diff
@@ -73,6 +73,44 @@ def __init__(self, target, registries, delay=0):
...
+        for interaction_class, doc in pairs:
+            print(repr(interaction_class))
+            print(
+                *("    " + line for line in doc.split("\n")),
```
rahulporuri, line 97: If I understand correctly, the `"          "` adds indentation to the help documentation. IMO it is cleaner and more readable to use `textwrap.indent` instead and replace `"        "` with `r"\t"` (maybe)

traitsui/testing/tester/ui_wrapper.py
```diff
@@ -73,6 +73,44 @@ def __init__(self, target, registries, delay=0):
...
+                sep="\n",
+            )
+            print()
+
+        print()
```
rahulporuri, line 102: instead of having empty print function calls which are adding a new line if i understand correctly, why not print
```
----------
Locations
----------
```

that way, you are emphasising the text and we won't need empty print calls. Same comment regd "Interactions".

rahulporuri (commented):
I don't have any comments on the implementation at the moment but I made some comments about the how the help looks.

traitsui/testing/tester/ui_wrapper.py
kitchoi: Because the interaction types may not support being compared.
e.g. `sorted([float, str])` will raise a TypeError.

All we care is that the help message show the objects in a deterministic order (it is easier to test). Because the help message is printing `repr(interaction_class))`, sorting them using `repr` seems reasonable, but it is really not _that_ important.

traitsui/testing/tester/ui_wrapper.py
kitchoi: Yep, this needs a test.

traitsui/testing/tester/ui_wrapper.py
kitchoi: It was motivated by the situation when there are no interactions supported. If I did not have a newline, it would look like this, which is a little uglier.

```
Interactions
------------
Locations
---------
<class 'traitsui.testing.tester.locator.Index'>
    A locator for locating a target that is uniquely specified by a single
    0-based index.
    
    Attributes
    ----------
    index : int
        0-based index

Help on an item in the list editor
Interactions
------------
<class 'traitsui.testing.tester.command.MouseClick'>
    An object representing the user clicking a mouse button.
    Currently the left mouse button is assumed.
    
    In most circumstances, a widget can still be clicked on even if it is
    disabled. Therefore unlike key events, if the widget is disabled,
    implementations should not raise an exception.

```

I could change this so that there is at least a sentence for saying there is nothing. It adds a bit more logic but it might look nicer.

kitchoi: Done. Now the example where there are no interactions look like this:
```
Interactions
------------
No interactions are supported.

Locations
---------
<class 'traitsui.testing.tester.locator.Index'>
    A locator for locating a target that is uniquely specified by a single
    0-based index.

    Attributes
    ----------
    index : int
        0-based index
```

Is the "-------" above the section still needed?

rahulporuri: nope. looks good.

traitsui/testing/tester/ui_wrapper.py
rahulporuri: IMO it is still worth adding a one line comment here that reversing means ordering based on priority.

rahulporuri (approved):
LGTM

traitsui/testing/tester/ui_wrapper.py
```diff
@@ -73,6 +74,57 @@ def __init__(self, target, registries, delay=0):
...
+        (This function is intended for interactive use.)
+        """
+        # mapping from interaction types to their documentation
+        interaction_to_doc = dict()
+
+        # mapping from location types to their documentation
+        location_to_doc = dict()
+
+        # Let higher priority registry override the documentation from lower
+        # priority registry.
```
rahulporuri, line 88:
```suggestion
        # Order registries by their priority.
```

traitsui/testing/tester/tests/test_registry.py
```diff
@@ -108,6 +108,33 @@ def handler(wrapper, interaction):
         with self.assertRaises(ValueError):
             registry.register_handler(SpecificEditor, UserAction, handler)
 
+    def test_error_get_interaction_coc(self):
```
aaronayres35, line 111: typo: should be `test_error_get_interaction_doc`

traitsui/testing/tester/tests/test_registry.py
```diff
@@ -146,3 +173,27 @@ def solver(wrapper, location):
...
+            locator_class=Locator,
+            solver=lambda w, l: 1,
+        )
+
+        help_text = registry.get_location_doc(
+            target_class=float, locator_class=Locator,
+        )
+        self.assertEqual(help_text, "Some default documentation.")
+
+    def test_error_get_interaction_coc(self):
```
aaronayres35: same typo here

traitsui/testing/tester/ui_wrapper.py
```diff
@@ -73,6 +74,57 @@ def __init__(self, target, registries, delay=0):
...
+        for registry in self._registries[::-1]:
```
aaronayres35, line 88: Just for my own understanding, the concept of priority is needed because two registries may have handlers or solvers registered for the same `type_` below.  So, by going in reverse order we override documentation until we have the docs for the highest priority registry?  This would then be the registry used if someone asks for an interaction or location with type_

aaronayres35 (approved):
A little late, but LGTM! just a couple typos.  Also disregard my question, I believe I understand fully now

traitsui/testing/tester/ui_wrapper.py
kitchoi: That's right!

PR #1250, opened 2020-09-21
Rename UIWrapper.target to UIWrapper._target

### LINKED ISSUE #1249: Rename UIWrapper.target to UIWrapper._target
The `target` attribute on `UIWrapper` should be private to users of the tester, but public/visible to developers who implement testing logic using the target.
By renaming the attribute with a preceding underscore, when test code tries access the attribute, it will be more obvious that the test code is depending on implementation details of a UI component.

Part of #1173

aaronayres35:
fixes #1249 
This pr replaces UIWrapper.target with UIWrapper._target. I also tried to make the change where appropriate in the documentation (in docstrings). 

Note this may cause some currently open PRs to have test failures (e.g. #1243).  They will simply need to update and uses of .target to ._target, which should be an easy fix.

### DIFF
--- a/traitsui/testing/tester/common_ui_targets.py
+++ b/traitsui/testing/tester/common_ui_targets.py
@@ -69,7 +69,7 @@ class _BaseSourceWithLocation:
         registry.register_solver(
             target_class=cls.source_class,
             locator_class=cls.locator_class,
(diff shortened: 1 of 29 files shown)


kitchoi (approved):
LGTM

PR #1251, opened 2020-09-21
adding UI Tester test/example for InstanceEditor_demo.py

aaronayres35:
part of #1220 

This PR adds a very simply test for the instance editor demo.  
One thing to note, I originally intended to test by using the simple style first and then custom.  However, once the tester clicks on the button for the simple style and the new ui for the instance pops up, I don't believe we currently have a way to close that ui.  Might be something we eventually want to add

### DIFF
--- a/integrationtests/test_all_examples.py
+++ b/integrationtests/test_all_examples.py
@@ -467,3 +467,10 @@ class TestInteractExample(unittest.TestCase):
             name.perform(command.KeySequence("ABC"))
             name.perform(command.KeyClick("Enter"))
             self.assertEqual(demo.sample_instance.name, "ABC")
+
+            demo.sample_instance.name = "XYZ"
+            simple_displayed = name.inspect(query.DisplayedText())
+            custom_name = custom.find_by_name("name")
+            custom_displayed = custom_name.inspect(query.DisplayedText())
+            self.assertEqual(simple_displayed, "XYZ")
+            self.assertEqual(custom_displayed, "XYZ")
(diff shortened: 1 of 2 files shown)


kitchoi (approved):
LGTM

kitchoi:
> However, once the tester clicks on the button for the simple style and the new ui for the instance pops up, I don't believe we currently have a way to close that ui. Might be something we eventually want to add

We can close it. Using the `simple = tester.find_by_id(ui, "simple")` you have in your test, after launching the dialog, `simple.find_by_id("OK").perform(MouseClick())` should close the dialog.

I have tried this in another context and if I remember correctly, it works for Qt, but for wx, it triggers some orthogonal layout persistent / restoration issues.

So this should work:
```
simple = tester.find_by_id(ui, "simple")
simple.perform(MouseClick()) # launch the dialog
simple.find_by_id("OK").perform(MouseClick())  # close the dialog
simple.perform(MouseClick()) # launch a new dialog!
```

When the `create_ui` exits the context, the parent UI should be disposed, and with it, the dialog UI. Closing the parent UI without closing the dialog UI first is also a valid usage, so it should not be a problem not trying to dialog UI in the test. Closing the dialog UI first would be another unit test for the InstanceEditor.

kitchoi:
Ah I assumed the example we have will have the OK button for the dialog. I was wrong, it does not have the OK button. Not having a OK button is a valid layout too. Then you are right, there is currently no way to close the dialog from the tester. We will need another action for that. Maybe in the future.

PR #1252, opened 2020-09-21
adding UI Tester test/example for List_editor_notebook_selection_demo.py

aaronayres35:
part of #1220 

Adds a simple test to see that we can navigate to a tab and change traits of a person, and also that the index trait for the demo is updated.

### DIFF
--- a/integrationtests/test_all_examples.py
+++ b/integrationtests/test_all_examples.py
@@ -445,3 +445,21 @@ class TestInteractExample(unittest.TestCase):
             self.assertEqual(demo.checklist, ["three"])
             item3.perform(command.MouseClick())
             self.assertEqual(demo.checklist, [])
+
+    def test_list_editor_notebook_selection_demo(self):
+        # Test List_editor_notebook_selection_demo.py in examples/demo/Advanced
+        filepath = os.path.join(
+            DEMO, "Advanced", "List_editor_notebook_selection_demo.py"
+        )
+        demo = load_demo(filepath, "demo")
+
+        tester = UITester()
+        with tester.create_ui(demo) as ui:
+            people_list = tester.find_by_name(ui, "people")
+            person2 = people_list.locate(locator.Index(2))
+            person2.perform(command.MouseClick())
+            self.assertEqual(demo.index, 2)
+            age = person2.find_by_name("age")
+            age.perform(command.KeyClick("Backspace"))
+            age.perform(command.KeyClick("9"))
+            self.assertEqual(demo.people[2].age, 39)


aaronayres35:
One thing I noticed when writing this test:  I was originally going to add a test for https://github.com/enthought/traitsui/blob/master/traitsui/examples/demo/Standard_Editors/ListEditor_demo.py
This is the test I wrote: 
<details>
```
def test_list_editor_demo(self):
        # Test ListEditor_demo.py in examples/demo/Standard_Edtiors
        filepath = os.path.join(
            DEMO, "Standard_Editors", "ListEditor_demo.py"
        )
        demo = load_demo(filepath, "demo")

        tester = UITester()
        with tester.create_ui(demo) as ui:
            custom_list = tester.find_by_id(ui, "custom")
            item1 = custom_list.locate(locator.Index(1))
            for _ in range(6):
                item1.perform(command.KeyClick("Backspace"))
            item1.perform(command.KeySequence("Othello"))
            item1.perform(command.KeyClick("Enter"))
            self.assertEqual(demo.playlist,
                ["The Merchant of Venice", "Othello", "MacBeth"])
```
</details>
But I realize that the current list editor implementation for the tester, we only provide the ability to navigate to nested UI.  However, with a list of just strings, I was unable to modify an element of the list.  I would need to basically call `item1.find_by_name(___)` or something to access the text box corresponding to the second element in the list.

kitchoi:
> One thing I noticed when writing this test: I was originally going to add a test for https://github.com/enthought/traitsui/blob/master/traitsui/examples/demo/Standard_Editors/ListEditor_demo.py
But I realized that the current list editor implementation for the tester, we only provide the ability to navigate to nested UI. However, with a list of just strings, I was unable to modify an element of the list. I would need to basically call item1.find_by_name(___) or something to access the text box corresponding to the second element in the list.
This is the test I wrote:

Yeah, I think this is hitting a feature that has yet to be implemented.

Looks like we can simplify the implementation of the custom ListEditor to support both the new test you wrote and what we already support for when the item is a nested UI. Now that we have an implementation for the custom InstanceEditor, the location solver for custom ListEditor for when it is given an index can return the editor, rather than the `editor._ui`. Then the logic for the custom InstanceEditor will follow.

Something like this (most of that is taken from [here](https://github.com/enthought/traitsui/blob/59b5fd7de5005cb5116bc97cad840175da131b44/traitsui/testing/tester/qt4/implementation/list_editor.py#L87), I just removed the final `._ui`) should work, and we can remove `_IndexedCustomListEditor`:
```

def _get_next_target(list_editor, index):
    row, column = divmod(index, list_editor.factory.columns)
    # there are two columns for each list item (one for the item itself,
    # and another for the list menu button)
    column = 2*column
    grid_layout = list_editor._list_pane.layout()
    item = grid_layout.itemAtPosition(row, column)
    if item is None:
        raise IndexError(index)
    if list_editor.scrollable:
        list_editor.control.ensureWidgetVisible(item.widget())
    return item.widget()._editor

def register(registry):
    ...
    registry.register_solver(
        target_class=CustomEditor,
        locator_class=locator.Index,
        solver=lambda wrapper, location: (
            _get_next_target(wrapper.target, location.index)
        )
    )
```

A similar change can apply to wx. I will open an issue.

kitchoi (approved):

PR #1256, opened 2020-09-22
Move UI Tester tests of TraitsUI examples into examples/demos

### LINKED ISSUE #1255: Move UI Tester examples (tests of TraitsUI examples) from test_all_examples.py to sit next to corresponding example
The examples of the tester which are tests of the traitsui examples were initially added to the file `test_all_examples.py` in `integrationtests`.  However, after #1254 was merged, they can be moved to sit next to the examples that they test so they are more visible/discoverable while still being run by CI.

aaronayres35:
fixes #1255 
This PR does not change any code in the tests themselves, it simply moves them so that they sit in the tests folder within the directory matching the example that they are testing.  It follows the example done in #1254 

It also removes the `load_demo` function as (as mentioned in 1254) we can instead use `runpy.run_path()`


traitsui/examples/demo/Advanced/tests/test_List_editor_notebook_selection_demo.py
```diff
@@ -0,0 +1,47 @@
...
+FILENAME = "List_editor_notebook_selection_demo.py"
+
+#: Path of the demo script
+DEMO_PATH = os.path.join(os.path.dirname(__file__), "..", FILENAME)
+
+
+class TestListEditorNotebookSelectionDemo(unittest.TestCase):
+
+    def test_list_editor_notebook_selection_demo(self):
+        # Test List_editor_notebook_selection_demo.py in examples/demo/Advanced
```
aaronayres35, line 28: these comments are no longer needed, I will remove them

traitsui/examples/demo/Standard_Editors/tests/test_ButtonEditor_demo.py
```diff
@@ -0,0 +1,68 @@
...
+
+# FIXME: Import from api instead
+# enthought/traitsui#1173
+from traitsui.testing.tester import command
+from traitsui.testing.tester.ui_tester import UITester
+
+from traitsui.tests._tools import (
+    requires_toolkit,
+    ToolkitName,
+)
```
kitchoi, line 23: Demo examples are where we point developers to when we say "here is how you do XYZ". So we should expect this code being copied and pasted, then adapted by external projects.
Importing from the private `_tools` is not recommended, so we would need to handle the absence of `ModalDialogTester` on wx differently.

This might be what we can do here instead:
https://github.com/enthought/pyface/blob/4fda22878e64d930ce0d2bebac69907ea71b31b9/pyface/tests/test_dialog.py#L26-L29

Then skip the test if `no_modal_dialog_tester` is true.
https://github.com/enthought/pyface/blob/4fda22878e64d930ce0d2bebac69907ea71b31b9/pyface/tests/test_dialog.py#L109

kitchoi: > Importing from the private _tools is not recommended

It is okay for traitsui to import traitsui's private module, so it is okay this test was in traitsui test suite not designated as demo examples.

kitchoi (approved):
LGTM

PR #1258, opened 2020-09-22
Start of UI Tester module renaming

aaronayres35:
Part of #1248 

This PR does not do any reorganization, it simply renames a few of the modules with a leading underscore.  Namely, 
`testing.exception_handling`, `testing.gui`, `testing.tester.compat`, `testing.tester.editor.layout` and both `helpers.py` files were renamed `_interaction_helpers.py`.  The amount of changes in this PR is already quite substantial (although they are all trivial), and that is why I stopped here.  I broke up some of the name changes by commit so hopefully that makes them slightly easier to review.  If this PR is too long I can break this down further into smaller PRs.

Issue #1248, opened 2020-09-21
Rename and reorganize modules in traitsui.testing


kitchoi:
Part of #1173
We have decided to let renaming happen later. Now that we have a number of use cases exercised, we are in a better place to reorganize and rename the modules so that:
(1) It is more obvious to developers what is related to what. Easier to find things, and easier to know where to add things
(2) It is more obvious what is public to vs hidden from users of the library.

@aaronayres35 and I discussed this topic, and this is the new layout we have so far. 

<details>
    <summary> Proposed new layout </summary>

(Note the following assumes #1247 has happened.)

```
./testing
* api.py
* _exception_handling.py
* _gui.py
* ./tester
      * ./_ui_tester_register
         * (Everything here are to do with the default registry in UITester to provide builtin support for TraitsUI GUI components)
         * ./qt4
            * ./_traitsui
               * (The intention here is that this matches the structure in traitsui.qt4)
               * button_editor.py
               * …
               * range_editor.py
                  * class LocatedTextbox:
                  * class LocatedSlider:
               * text_editor.py
               * ui_base.py
            * _interaction_helpers.py
            * _registry_helpers.py
            * default_registry.py
         * ./wx
            * ./_traitsui
               * (The intention here is that this matches the structure in traitsui.qt4)
               * button_editor.py
               * …
               * range_editor.py
                  * class LocatedTextbox:
                  * class LocatedSlider:
               * text_editor.py
               * ui_base.py
            * _interaction_helpers.py
            * _registry_helpers.py
            * default_registry.py
         * _common_ui_targets.py
         * _compat.py
         * _layout.py
         * _traitsui_ui.py
            * Previously ../registry_helpers.py
            * def register_traitsui_ui_solvers(...):
         * default_registry.py
            * def get_default_registry(...):
            * (Move _get_ui_registry from ui_tester.py to inside get_default_registry)
   * command.py
   * exceptions.py
   * locator.py
   * query.py
   * registry.py
   * ui_tester.py
   * ui_wrapper.py
```
</details>

PR #1259, opened 2020-09-22
2nd round of of UI Tester module renaming 

aaronayres35:
Part of #1248, a continuation of #1258 

This PR also does not do any reorganization, it simply renames the still needing to be changed modules as specified by 1248.
Again the amount of changes in this PR is quite large.  Name changes were in general made as 1 change per commit (in some cases if there were sister modules e.g. `wx/implementation` and `qt4/implementation` the changes were done in the same commit).


rahulporuri:
I think there's an extra unintentional underscore here - `traitsui/testing/tester/wx/tests/test__interaction_helpers.py`

rahulporuri (approved):
LGTM except for the one file with a possible unintentional extra underscore in the filename.

rahulporuri:
still LGTM

PR #1261, opened 2020-09-22
Move LocatedTextbox and LocatedSlider to RangeEditor implementation for testing

### LINKED ISSUE #1247: Move LocatedTextbox and LocatedSlider to RangeEditor implementation for testing
Currently `LocatedTextbox` and `LocatedSlider` are implemented as reusable targets `traitsui.testing.qt4.common_ui_targets` and `traitsui.testing.wx.common_ui_targets`. They are only used by RangeEditor at the moment.

The current configuration means the `register` functions in `traitsui.testing.qt4.implementation.range_editor` and `traitsui.testing.wx.implementation.range_editor` are not standalone: If I want to have a registry that only support RangeEditor, using that `register` function is not enough, I still need to call `LocatedTextbox.register` and `LocatedSlider.register`.

Even if there are use cases where we can locate a textbox/slider and then throw away the location information, it might still be the case that creating an intermediate object locally within the context of an editor is more readable and less obscure than trying to refactor it.

Proposal is to move LocatedTextbox and LocatedSlider back into `range_editor` and register the interactions for these objects there.

aaronayres35:
fixes #1247 

This PR moves the classes `LocatedTextbox` and `LocatedSlider` from `wx/common_ui_targets` and `qt4/common_ui_targets` to `wx/implementation/range_editor` and `qt4/implementation/range_editor` respectively (and then deleted the common_ui_target files).

Now, the range_editors register methods are stand alone, and the LocatedTexbox/LocatedSlider register methods are not called directly in `default_registry` but now indirectly, through the range_editor register methods.


traitsui/testing/tester/wx/_traitsui/range_editor.py
```diff
@@ -15,12 +15,70 @@
...
+        given registry.
+
+        If there are any conflicts, an error will occur.
+
+        Parameters
+        ----------
+        registry : TargetRegistry
+            The registry being registered to.
+        """
+        registry_helper.register_editable_textbox_handlers(
```
kitchoi, line 45:
```suggestion
        _registry_helper.register_editable_textbox_handlers(
```

I think the tests are catching these errors.

traitsui/testing/tester/qt4/_traitsui/range_editor.py
```diff
@@ -16,12 +16,70 @@
...
+        given registry.
+
+        If there are any conflicts, an error will occur.
+
+        Parameters
+        ----------
+        registry : TargetRegistry
+            The registry being registered to.
+        """
+        registry_helper.register_editable_textbox_handlers(
```
kitchoi, line 46:
```suggestion
        _registry_helper.register_editable_textbox_handlers(
```

traitsui/testing/tester/qt4/_traitsui/range_editor.py
```diff
@@ -16,12 +16,70 @@
...
 )
 
-from traitsui.testing.tester import locator
-from traitsui.testing.tester.qt4 import _registry_helper
-from traitsui.testing.tester.qt4.common_ui_targets import (
-    LocatedSlider,
-    LocatedTextbox
-)
+from traitsui.testing.tester import command, locator
+from traitsui.testing.tester.qt4 import _interaction_helpers, _registry_helper
```
kitchoi, line 20: An observation entirely out-of-scope for this PR: strange mismatch between "helpers" and "helper" there. Since they are now obviously private, we should not need to worry too much about breaking external code if we want to rename them after release. The names still provide good enough hints to maintain encapsulation and to guide code discovery. So the little difference here bother those who suffer from OCD (who cares about them 😂 )

kitchoi (commented):
Two renaming are still required.

traitsui/testing/tester/qt4/_traitsui/range_editor.py
aaronayres35: I agree, I also dislike the mismatch.  We can make a fix in a separate PR

kitchoi (approved):
LGTM.

PR #1262, opened 2020-09-23
Fix and expand docstring for UIWrapper

kitchoi:
This PR makes changes to the docstring of UIWrapper so that they provide more information.

### DIFF
--- a/traitsui/testing/tester/ui_wrapper.py
+++ b/traitsui/testing/tester/ui_wrapper.py
@@ -60,9 +60,6 @@ class UIWrapper:
 
     Attributes
     ----------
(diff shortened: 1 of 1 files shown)


traitsui/testing/tester/ui_wrapper.py
```diff
@@ -60,9 +60,6 @@ class UIWrapper:
 
     Attributes
     ----------
-    target : any
```
kitchoi, line 63: `target` has since be renamed to `_target` so it is no longer a public attribute.

rahulporuri (approved):
LGTM

Issue #1232, opened 2020-09-18
reraise_exceptions


kitchoi:

kitchoi:
After #1231 is merged and after moving `push_exception_handler`/`pop_exception_handler` to test case setup/teardown, tests that use both `reraise_exceptions()` and `create_ui` like this:
https://github.com/enthought/traitsui/blob/80a31705d974edd9624ceb34cd3e83a6f643960a/traitsui/tests/editors/test_drop_editor.py#L47-L48

can just be `with create_ui(...): ...`

aaronayres35:
> After #1231 is merged and after moving `push_exception_handler`/`pop_exception_handler` to test case setup/teardown, tests that use both `reraise_exceptions()` and `create_ui` like this:
> https://github.com/enthought/traitsui/blob/80a31705d974edd9624ceb34cd3e83a6f643960a/traitsui/tests/editors/test_drop_editor.py#L47-L48
> 
> can just be `with create_ui(...): ...`

Once #1270 is merged should we open a separate issue to make all these replacements?

kitchoi:
I think what I said there was too simplistic. My apologies. I had a look at the occurrences of `reraise_exceptions` we have.
Some of these `reraise_exceptions` are now obviously redundant. But some do need to stay because not all test logics go via UITester (strictly speak, UIWrapper) [where we are wrapping GUI event processing with the same context manager](https://github.com/enthought/traitsui/blob/aa16c9cabc6d4d4904f30103f09b32227a8199fc/traitsui/testing/tester/ui_wrapper.py#L322-L332).

It costs little extra to keep the redundant ones, apart from the fact that they are now noise to developers reading the test code. To me it isn't important enough to say there is an issue about keeping them, but that could be just me.

PR #1266, opened 2020-09-23
Add trait exception handler in TraitsUI tests setup

kitchoi:
Part of #1232

This PR introduces a base class `BaseTestMixin` for ~all~ TraitsUI test cases to subclass. This is such that all the tests will have the trait change exception handler set, otherwise exceptions from trait change handler will only result in a warning. These instances are almost always programming errors that the tests should capture and re-raise.

The practice is mixed in `reraise_exceptions`, in order to remove that logic for #1232, the logic is first added to all the test cases. In the next PR, I will remove the logic from reraise_exceptions and it should not change test behaviour.

There are also a few drive-by flake8 fixes wherever they are small and re-reviewable.

Note that some tests that never use `reraise_exceptions` also receive this treatment, as I believe this is the right thing for all the _trait-using_ tests to check for errors from trait change handlers.


traitsui/tests/test_actions.py
```diff
@@ -85,7 +86,13 @@ def _qt_click_button(ui):
...
 
 
-class TestActions(unittest.TestCase):
+class TestActions(BaseTestMixin, unittest.TestCase):
+
+    def setUp(self):
+        BaseTestMixin.setUp(self)
+
+    def tearDown(self):
+        BaseTestMixin.tearDown(self)
```
aaronayres35, line 95: Do we need to override these? This is more of a general python question, if a subclass inherits from 2 base classes that have a method of the same name, which method is used when it is called on a subclass?  In this case I believe the `unittest.TestCase` setUp and tearDown methods do nothing by default.  Im guessing by the fact that you included this, that in this case the programmer would need to specify by overwriting the method?

traitsui/tests/test_editor.py
```diff
@@ -182,7 +184,15 @@ def create_editor(
...
-class TestEditor(GuiTestAssistant, unittest.TestCase):
+class TestEditor(BaseTestMixin, GuiTestAssistant, unittest.TestCase):
+
+    def setUp(self):
+        BaseTestMixin.setUp(self)
+        GuiTestAssistant.setUp(self)
+
+    def tearDown(self):
+        GuiTestAssistant.tearDown(self)
+        BaseTestMixin.tearDown(self)
```
aaronayres35, line 195: similar comment to before.  Why did we not need to call `GuiTestAssistant.setUp(self)` before but now we do?

aaronayres35 (approved):
LGTM! just a couple questions

traitsui/tests/test_actions.py
kitchoi: For `TestActions`, I can totally do this instead:
```
class TestActions(BaseTestMixin, unittest.TestCase):
    # setUp and tearDown are not overridden.
    # BaseTestMixin.setUp and BaseTestMixin.tearDown are called by virtue of inheritance.
```
This would be similar to the tests that previously only subclass `GuiTestAssistant` and `TestCase`. This works the same way:
```
class TestSomething(GuiTestAssistant, unittest.TestCase):
    # GuiTestAssistant.setUp and GuiTestAssistant.tearDown by virtue of inheritance.
```
But as soon as I add another base class that also offers `setUp` and `tearDown`, I need to expand the subclass's `setUp` and `tearDown` to call all base classes' setUp/tearDown.

e.g. as soon as I add `BaseTestMixin`, I need to remember to call `GuiTestAssistant.setUp` and `GuiTestAssistant.tearDown`:
```
class TestSomething(BaseTestMixin, GuiTestAssistant, unittest.TestCase):
    def setUp(self):
        BaseTestMixin.setUp(self)
        GuiTestAssistant.setUp(self)

    def tearDown(self):
        GuiTestAssistant.tearDown(self)
        BaseTestMixin.tearDown(self)
```

If I did not do this, i.e. if I had done this instead:
```
class TestSomething(BaseTestMixin, GuiTestAssistant, unittest.TestCase):
    # setUp and tearDown are not overridden
```
Only `BaseTestMixin.setUp` and `BaseTestMixin.tearDown` will be called. `GuiTestAssistant.setUp` and `GuiTestAssistant.tearDown` that were previously called will NOT be called. That would be an error.

This is why even though I do not need to override `TestActions.setUp` and `TestActions.tearDown`, I still override them to call `BaseTestMixin`'s setUp and tearDown so that the next person adding another base class doesn't need to think that hard. The computer sees the same thing, but this taxes the human less and lowers the risk of human errors.

rahulporuri: >This is why even though I do not need to override TestActions.setUp and TestActions.tearDown, I still override them to call BaseTestMixin's setUp and tearDown so that the next person adding another base class doesn't need to think that hard. The computer sees the same thing, but this taxes the human less and lowers the risk of human errors.

Personally I wouldn't have done this because it's not common to add baseclasses (or change baseclasses) to testcases.

rahulporuri (approved):
LGTM

rahulporuri:
@kitchoi not sure why all of the files touched in this PR weren't flake8-d? I'm not sure if all of them are compliant already - and if they are, they can be removed from `setup.cfg`.

kitchoi:
> not sure why all of the files touched in this PR weren't flake8-d? I'm not sure if all of them are compliant already - and if they are, they can be removed from setup.cfg.

The files weren't flake8 clean before CI does flake8 check. I inspected the ones that are skipped in `setup.cfg` and cleaned the few I deemed easy enough to review along with this PR (this [commit](22a11eebcd4f0675d25772734302fcab7f7ac272)). For those, I removed them from `setup.cfg`. For others, I decided that those changes would be too obscure/hidden to mix them into this PR so I defer including them as drive-by.

traitsui/tests/test_actions.py
kitchoi: Yeah I tend to think this is closer to a personal preference which I won't ask others to do, but I'd do myself :)

PR #1264, opened 2020-09-23
UI Tester Module Reorganization

aaronayres35:
Part of #1248 

This PR pulls numerous modules and files into a new sub directory `_ui_tester_register` as specified by 1248.  The movement of modules is broken up to be one change per commit to hopefully make it a bit easier to review. 

After this I believe 1248 is very nearly finished, we just need to "Move _get_ui_registry from ui_tester.py to inside get_default_registry" in default_registry.  

~~Edit: actually I am just going to make that change in one last commit on this PR so that this PR can close up 1248~~ Scratch that I think it is better to put that in a smaller separate PR

PR #1267, opened 2020-09-23
Use current package in default_registry.py instead of hard coded path when locating toolkit specific default_registry module

aaronayres35:
This is a very short PR to address the issue that was brought up here: https://github.com/enthought/traitsui/pull/1264#discussion_r493637964


kitchoi (approved):
LGTM

PR #1268, opened 2020-09-23
UI Tester update List Editor implementation to not only work with nested uis

### LINKED ISSUE #1253: Implementation for testing custom list editor assumes a nested UI that may not exist
See comment in https://github.com/enthought/traitsui/pull/1252#issuecomment-696385370
Currently the implementation for testing the custom list editor assumes the item is viewed via a nested UI. This is not always true.

e.g. the following test should work, but currently the test raises ~InteractionNotSupported~ `AttributeError` because `._ui` is not available on a simple text editor.
```

class PlayList(HasStrictTraits):
    play_list = List(Str)

class TestListEditor(unittest.TestCase):
    def test_list_of_text(self):
        obj = PlayList(
            play_list=["The Merchant of Venice", "Hamlet", "MacBeth"]
        )
        tester = UITester()
        view = View(Item("play_list", editor=ListEditor(), style="custom"))
        with tester.create_ui(obj, dict(view=view)) as ui:
            custom_list = tester.find_by_name(ui, "play_list")
            item1 = custom_list.locate(locator.Index(1))
            item1.perform(command.KeySequence(" Returns"))
            self.assertEqual(obj.play_list,
                ["The Merchant of Venice", "Hamlet Returns", "MacBeth"]
            )
```

We should be able to fix this, and support the case for nested UI by simply returning the nested editor from the location solver used for the custom list editor and the Index locator.
(See this https://github.com/enthought/traitsui/pull/1252#issuecomment-696606132).

This might mean `locate(Index(...))` may raise IndexError instead of deferring to when we perform an interaction. I think that is okay (probably better).

aaronayres35:
fixes #1253 

This PR modifies the UI Tester implementation of Custom List Editor so that it works with Lists of objects that don't have nested UIs (e.g. just a list of Strings).  It does so by the means described on issue 1253 and here: https://github.com/enthought/traitsui/pull/1252#issuecomment-696385370

### DIFF
--- a/traitsui/examples/demo/Standard_Editors/ListEditor_demo.py
+++ b/traitsui/examples/demo/Standard_Editors/ListEditor_demo.py
@@ -21,13 +21,13 @@ class ListEditorDemo(HasTraits):
 
     # Items are used to define display, one per editor style:
     list_group = Group(
-        Item('play_list', style='simple', label='Simple', height=75),
+        Item('play_list', style='simple', label='Simple', height=75, id='simple'),
         Item('_'),
-        Item('play_list', style='custom', label='Custom'),
+        Item('play_list', style='custom', label='Custom', id='custom'),
         Item('_'),
-        Item('play_list', style='text', label='Text'),
+        Item('play_list', style='text', label='Text', id='text'),
         Item('_'),
-        Item('play_list', style='readonly', label='ReadOnly')
+        Item('play_list', style='readonly', label='ReadOnly', id='readonly')
     )
 
     # Demo view:
--- a/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/list_editor.py
+++ b/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/list_editor.py
@@ -71,7 +71,7 @@ def _get_next_target(list_editor, index):
     list_editor : CustomEditor
         The custom style list editor in which the target is contained.
     index : int
-        the index of the target of interest in the list 
+        the index of the target of interest in the list
 
     Returns
     -------
@@ -88,6 +88,7 @@ def _get_next_target(list_editor, index):
     if list_editor.scrollable:
         list_editor.control.ensureWidgetVisible(item.widget())
     return item.widget()._editor
+
 
 def register(registry):
     """ Register interactions for the given registry.
--- a/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/list_editor.py
+++ b/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/list_editor.py
@@ -75,7 +75,8 @@ def _get_next_target(list_editor, index):
 
     Returns
     -------
-    The obtained target.
+    traitsui.editor.Editor
+        The obtained target.
     """
     row, column = divmod(index, list_editor.factory.columns)
     # there are two columns for each list item (one for the item itself,
--- a/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/list_editor.py
+++ b/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/list_editor.py
@@ -63,48 +63,31 @@ class _IndexedNotebookEditor(_BaseSourceWithLocation):
         return self.source._uis[self.location.index][1]
 
 
-class _IndexedCustomEditor(_BaseSourceWithLocation):
-    """ Wrapper for a ListEditor (custom) with an index.
+def _get_next_target(list_editor, index):
+    """ Gets the target at a given index from a Custom List Editor.
+
+    Parameters
+    ----------
+    list_editor : CustomEditor
+        The custom style list editor in which the target is contained.
+    index : int
+        the index of the target of interest in the list 
+
+    Returns
+    -------
+    The obtained target.
     """
-    source_class = CustomEditor
-    locator_class = locator.Index
-
-    @classmethod
-    def register(cls, registry):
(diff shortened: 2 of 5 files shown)


traitsui/examples/demo/Standard_Editors/ListEditor_demo.py
```diff
@@ -21,13 +21,13 @@ class ListEditorDemo(HasTraits):
...
+        Item('play_list', style='simple', label='Simple', height=75, id='simple'),
         Item('_'),
-        Item('play_list', style='custom', label='Custom'),
+        Item('play_list', style='custom', label='Custom', id='custom'),
         Item('_'),
-        Item('play_list', style='text', label='Text'),
+        Item('play_list', style='text', label='Text', id='text'),
         Item('_'),
-        Item('play_list', style='readonly', label='ReadOnly')
+        Item('play_list', style='readonly', label='ReadOnly', id='readonly')
```
aaronayres35, line 30: One thing I found very strange/misleading is that these styles of editor dont really exist.  All of these are exactly the same and in fact, in the demo a user can modify the 'readonly' one just like the rest.  The code for getting the list editor is simply:
```
def _get_custom_editor_class(self):
        if self.use_notebook:
            return toolkit_object("list_editor:NotebookEditor")
        return toolkit_object("list_editor:CustomEditor")
```
This is not directly relevant to this PR, but it seemed off to me

aaronayres35: Note they are defined:
https://github.com/enthought/traitsui/blob/5a030766dd1168ddeb14f2d481a08e484e65f763/traitsui/qt4/list_editor.py#L412-L423
https://github.com/enthought/traitsui/blob/5a030766dd1168ddeb14f2d481a08e484e65f763/traitsui/wx/list_editor.py#L473-L484
but based on the above code it seems they aren't ever going to be used?

kitchoi: You are right to find this strange.

The design of the EditorFactory is that there are four allowed values for "style": "simple", "custom", "text", "readonly". (defined [here](https://github.com/enthought/traitsui/blob/6f25745e3a9035b17e5512998dbc76b91be53f2d/traitsui/ui_traits.py#L56-L57)).
And each style should map to one editor class. The default implementation is like this:
https://github.com/enthought/traitsui/blob/460742421bcdddfec124a35287843e8a0eae0b49/traitsui/editor_factory.py#L222-L283

Trouble is, sometimes we need more than four editor classes for an ever growing number of options, but the allowed values in "style" is fixed. So the editor factory ends up having to branch out from one of these four styles (usually "custom", for its name is the least specific).

e.g. RangeEditor has this `mode` option, and to select an editor class, you need to combine "style" and "mode":
https://github.com/enthought/traitsui/blob/460742421bcdddfec124a35287843e8a0eae0b49/traitsui/editors/range_editor.py#L216-L271

traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/list_editor.py
```diff
@@ -63,47 +63,31 @@ def _get_nested_ui(self):
...
+    Parameters
+    ----------
+    list_editor : CustomEditor
+        The custom style list editor in which the target is contained.
+    index : int
+        the index of the target of interest in the list
+
+    Returns
+    -------
+    The obtained target.
```
kitchoi, line 78: Nit: This docstring can be more helpful here by specifying the returned value is an instance of `traitsui.editor.Editor`. (We studied the code to know that, it helps the next person who hasn't studied the code to quickly know what it is.)

traitsui/testing/tester/_ui_tester_registry/wx/_traitsui/list_editor.py
```diff
@@ -62,44 +62,29 @@ def _get_nested_ui(self):
...
+    Parameters
+    ----------
+    list_editor : CustomEditor
+        The custom style list editor in which the target is contained.
+    index : int
+        the index of the target of interest in the list
+
+    Returns
+    -------
+    The obtained target.
```
kitchoi, line 77: Same here.

kitchoi (approved):
LGTM.

traitsui/examples/demo/Standard_Editors/ListEditor_demo.py
aaronayres35: That makes sense, thank you!

So, because the ToolkitEditorFactory in https://github.com/enthought/traitsui/blob/5a030766dd1168ddeb14f2d481a08e484e65f763/traitsui/editors/list_editor.py#L170-L173 
subclasses EditorFactory, it has the `_get_simple_editor_class`, `_get_custom_editor_class`, `_get_text_editor_class`, and `_get_readonly_editor_class` methods.  The only one overridden is `_get_custom_editor_class` where there is another check added to see if someone wants to `use_notebook` and they can't add another style because only the original 4 are allowed.  My initial source of confusion was that all 4 styles appear exactly the same in the demo.  The one that stood out as seeming wrong was the readonly style, as if you run the demo, you can freely modify the text, and see the trait is updated as the other styles all immediately reflect any changes you make.  This didn't seem very 'readonly' and I initially thought for that it wasn't actually the ReadonlyEditor being used.  After looking more closely I realize that this actually intentional as the list being readonly only sets the mutable trait to false.  Thus the readonly style does not provide the additional buttons to 'Add Before' or 'Add After' etc, but the things in the list can still be modified.

PR #1269, opened 2020-09-23
Populating the traitsui.testing.api module

aaronayres35:
Part of #1173 

Note to keep this PR small and easier to review, this PR simply adds the import statements to `traitsui.testing.api.py`.  It does not change any other code to actually use the api (see the fifth bullet on 1173). These changes will be made in a separate PR after this is merged.  

I don't believe anything else belongs in the api atm?  One thing I wasn't sure about was if `UIWrapper` should be included, but I could not think of a case where a user would need a wrapper separate from having a UITester.  In all the test code we have written so far we never needed to import `UIWrapper`.  However, I very well may be wrong and it can easily be added here.

### DIFF
--- a/traitsui/testing/api.py
+++ b/traitsui/testing/api.py
@@ -8,3 +8,32 @@
 #
 #  Thanks for using Enthought open source!
 #
+from .tester.command import (
+    MouseClick,
+    KeyClick,
+    KeySequence
+)
+
+from .tester.exceptions import (
+    Disabled,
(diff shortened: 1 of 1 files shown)


kitchoi (approved):
LGTM.

Agreed we don't have to expose UIWrapper in the api.py as no ones is importing it. (The UIWrapper is nonetheless very public by virtue of public access via UITester.)

PR #1270, opened 2020-09-24
Remove trait change exception handler in reraise_exceptions

### LINKED ISSUE #1232: reraise_exceptions

kitchoi:
Closes #1232
(Part of #1173)

This PR
- Removes setting trait change exception handler in `reraise_exceptions`.
- Apply `BaseTestMixin` to one more test case that would have lost the setting if this wasn't done before/when `reraise_exceptions` lost this behaviour. Most of the other test cases that need this should have received the treatment from #1266.
(I missed this one test case there. 😞   I searched for these test cases by finding all references to `reraise_exceptions`, but I overlooked `traitsui.qt4.tests` package back then.)


aaronayres35 (approved):
Changes LGTM!

Issue #1257, opened 2020-09-22
Add tests for all TraitsUI examples/demos using UI Tester


aaronayres35:
Some of these have already been added (see #1220), but there are many more examples that still need to be tested.

Some of these examples may rely on functionality that has not yet been added to UI Tester, in which case those tests will need to be delayed.  Going through them may serve as a good means for finding what new features would be useful to implement for the tester. 

Additionally, these tests will serve as good examples of how to use UI Tester itself, and could possibly be referenced in documentation if needed.

kitchoi:
> could possibly be referenced in documentation if needed.

Some existing ones are now referenced in the User Manual "Testing TraitsUI Applications" section. So new references can be added there too.

PR #1277, opened 2020-09-24
adding UI Tester test/example for RangeEditor_demo.py

aaronayres35:
part of #1257 

This PR adds a test for `RangeEditor_demo.py`.  One thing of note: by the way the demo is constructed, the traits for the various ranges are each assigned their own tab in the window.  Currently the tester has not implemented a way to click on these tabs, so if one were to sett a `delay` and try to seee the test, in order tto follow the test, they would need to manually click on the tabs to see the test occurring.  

Also note, this PR contains a couple simple flake8 changes for some of the other demos, which I found when checking flake8 for this test.  They are completely unrelated to this PR, so I am happy to remove them if a reviewer would prefer that.

One other thing to note: a keyclick of Page Up or Page Down on a slider corresponds to a 1/10 change of the range shown.  Note for some of the int range that number is a faction (e.g. 2.4 for the medium range) which gets rounded.  Also, for a `LargeRangeSliderEditor` the displayed range does not match the range of the trait, so the size of Page Up/Page Down is a little tricky to follow.


kitchoi (approved):
LGTM!

PR #1278, opened 2020-09-24
UI Tester - add boolean editor

aaronayres35:
This PR adds an implementation for BooleanEditor to UI Tester.  It also modifies the current `test_boolean_editor.py` file by modifying it to use UI Tester, and adding another simple test.  Lastly, it adds tests for the examples `BooleanEditor_demo` and `BooleanEditor_simple_demo`. The previously list changes are made in roughly 1 change per commit.

### DIFF
--- a/traitsui/examples/demo/Standard_Editors/tests/test_BooleanEditor_simple_demo.py
+++ b/traitsui/examples/demo/Standard_Editors/tests/test_BooleanEditor_simple_demo.py
@@ -11,7 +11,7 @@ import runpy
 import unittest
 
 from traitsui.testing.api import (
-    DisplayedText, KeyClick, KeySequence, MouseClick, UITester
+    DisplayedText, IsChecked, KeyClick, KeySequence, MouseClick, UITester
 )
 
 #: Filename of the demo script
@@ -50,9 +50,8 @@ class TestBooleanEditorSimpleDemo(unittest.TestCase):
             displayed = readonly.inspect(DisplayedText())
             self.assertEqual(displayed, "True")
 
-            displayed_count_changes = count_changes.inspect(DisplayedText())
-            self.assertEqual(displayed_count_changes, '3')
-            self.assertEqual(displayed_count_changes, str(demo.count_changes))
+            simple_is_checked = simple.inspect(IsChecked())
+            self.assertEqual(simple_is_checked, True)
 
 
 # Run the test(s)
--- a/traitsui/examples/demo/Standard_Editors/tests/test_BooleanEditor_simple_demo.py
+++ b/traitsui/examples/demo/Standard_Editors/tests/test_BooleanEditor_simple_demo.py
@@ -1,0 +1,63 @@
+"""
+This example demonstrates how to test interacting with a boolean
+using BooleanEditor.
+
+The GUI being tested is written in the demo under the same name (minus the
+preceding 'test') in the outer directory.
+"""
+
+import os
+import runpy
+import unittest
+
+from traitsui.testing.api import (
+    DisplayedText, KeyClick, KeySequence, MouseClick, UITester
+)
+
+#: Filename of the demo script
+FILENAME = "BooleanEditor_simple_demo.py"
+
+#: Path of the demo script
+DEMO_PATH = os.path.join(os.path.dirname(__file__), "..", FILENAME)
+
+
+class TestBooleanEditorSimpleDemo(unittest.TestCase):
+
+    def test_boolean_editor_simple_demo(self):
+        demo = runpy.run_path(DEMO_PATH)["demo"]
+
+        tester = UITester()
+        with tester.create_ui(demo) as ui:
+            simple = tester.find_by_id(ui, 'simple')
+
+            readonly = tester.find_by_id(ui, 'readonly')
+            text = tester.find_by_id(ui, 'text')
+            count_changes = tester.find_by_name(ui, "count_changes")
+
+            self.assertEqual(demo.my_boolean_trait, False)
+            simple.perform(MouseClick())
+            self.assertEqual(demo.my_boolean_trait, True)
+
+            for _ in range(4):
+                text.perform(KeyClick("Backspace"))
+            text.perform(KeySequence("False"))
+            text.perform(KeyClick("Enter"))
+            self.assertEqual(demo.my_boolean_trait, False)
+
+            displayed_count_changes = count_changes.inspect(DisplayedText())
+            self.assertEqual(displayed_count_changes, '2')
+            self.assertEqual(displayed_count_changes, str(demo.count_changes))
+
+            demo.my_boolean_trait = True
+            displayed = readonly.inspect(DisplayedText())
+            self.assertEqual(displayed, "True")
+
+            displayed_count_changes = count_changes.inspect(DisplayedText())
+            self.assertEqual(displayed_count_changes, '3')
+            self.assertEqual(displayed_count_changes, str(demo.count_changes))
+
+
+# Run the test(s)
+unittest.TextTestRunner().run(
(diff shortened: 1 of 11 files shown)


traitsui/tests/editors/test_boolean_editor.py
```diff
@@ -43,6 +43,19 @@ def test_init_dispose(self):
...
+        view = View(Item("true_or_false", editor=BooleanEditor()))
+        obj = BoolModel()
+
+        tester = UITester()
+        with tester.create_ui(obj, dict(view=view)) as ui:
+            # sanity check
+            self.assertEqual(obj.true_or_false, False)
+            checkbox = tester.find_by_name(ui, "true_or_false")
+            checkbox.perform(MouseClick())
+            self.assertEqual(obj.true_or_false, True)
```
kitchoi, line 61: There are a number of recommended testing practices: [test pyramid](https://martinfowler.com/bliki/TestPyramid.html), [testing trophy](https://kentcdodds.com/blog/write-tests). But broadly speaking, we like tests that are
(1) fast and cheap to run, because then developers will be more likely to run them frequently, get test feedbacks often and catch errors sooner.
(2) small and focused so that its failure help pinning down the source of bug more quickly. This also allows tests targeting different behaviour to be run independently of each other. If all are lumped into a single test, the first failure stops the rest from being run so you don't get to see the whole picture early.

The test in the demo, if run on its own, is not slow, but it does lump a lot of things together. Furthermore, it does not live with the `traitsui.tests.editors` where developers find editor specific tests. Indeed, the tests in the demo domain serve different purposes: it protects the demo script and it serves to demonstrate writing tests. The demo test changes if the demo changes, which could be caused by educational reasons. The demo could well have lived in a separate repository outside of TraitsUI. With that, we should not rely on the demo tests to protect the editor's behaviour.

Imagine a developer changing the code for BooleanEditor, they will likely notice `test_boolean_editor` to be the tests for BooleanEditor, not the one in the demo. The test in the demo only gets picked up when they run the integration test suite. The integration test suite is a lot slower that one may not run as often.

Suggestion: Could you move the the test logic you have in the demo here, and organize them such that different test objectives can be run independently (e.g. separating simple editor and custom editor, separating trait change -> GUI change and GUI change -> trait change)? Then the demo test can also be simplified.

(Unrelated comment: When we have to add tests as an afterthought to code that doesn't have tests, we tend to have to go backwards and have to write bigger characteristics tests first to preserve behaviour before we can refactor the code to make it easier to test in small bits and pieces.)

traitsui/tests/editors/test_boolean_editor.py
```diff
@@ -43,6 +43,19 @@ def test_init_dispose(self):
         # Test init and dispose of the editor.
         view = View(Item("true_or_false", editor=BooleanEditor()))
         obj = BoolModel()
-        with reraise_exceptions(), \
-                create_ui(obj, dict(view=view)):
+
+        tester = UITester()
+        with tester.create_ui(obj, dict(view=view)):
```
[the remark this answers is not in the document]
kitchoi: After removing this test, could we also add wx back to `requires_toolkit`? The goal of fixing #752 seems far fetched, and the test you added should pass with wx.

traitsui/examples/demo/Standard_Editors/tests/test_BooleanEditor_simple_demo.py
```diff
@@ -0,0 +1,63 @@
...
+
+class TestBooleanEditorSimpleDemo(unittest.TestCase):
+
+    def test_boolean_editor_simple_demo(self):
+        demo = runpy.run_path(DEMO_PATH)["demo"]
+
+        tester = UITester()
+        with tester.create_ui(demo) as ui:
+            simple = tester.find_by_id(ui, 'simple')
+
```
kitchoi, line 32: Nit: This line seems redundant.

traitsui/examples/demo/Standard_Editors/tests/test_BooleanEditor_simple_demo.py
```diff
@@ -0,0 +1,63 @@
...
+            readonly = tester.find_by_id(ui, 'readonly')
+            text = tester.find_by_id(ui, 'text')
+            count_changes = tester.find_by_name(ui, "count_changes")
+
+            self.assertEqual(demo.my_boolean_trait, False)
```
kitchoi, line 37: It might read more naturally if this assertion is removed. This kind of sanity checking can be done inside unit tests that are not demo tests.

kitchoi (commented):
Thank you!
The test support code looks good. Some general comments on where we add tests for the editor.

traitsui/tests/editors/test_boolean_editor.py
aaronayres35: Hi @kitchoi, thank you for the review!  I added more simple unit tests to `test_boolean_editor` as you suggested.  I'm not exactly sure what should be removed from the demo tests to simplify them now like you mentioned.  
Also, as we discussed I added an `IsChecked` class to `query` to better be able to test trait change -> GUI change.

traitsui/testing/tester/query.py
```diff
@@ -39,3 +39,13 @@ class DisplayedText:
     Implementations should return a ``str``.
     """
     pass
+
+
+class IsChecked:
+    """ An object representing an interaction to obtain whether a checkbox
+    is checked or empty.
```
kitchoi, line 46: Not just checkbox can be have "isChecked" state. Other widgets like radio button can have that state. Perhaps instead of saying "checkbox", this should say "checkable widget (e.g. checkbox)"?

"empty" sounds a bit odd, but that could just be me. "is checked or not" seems good enough?

traitsui/testing/tester/query.py
```diff
@@ -39,3 +39,13 @@ class DisplayedText:
...
+
+    Implementations should return something which evaluates to True if checked
+    and False if not.
```
kitchoi, line 49: I'd suggest sticking with just `True` or `False`. Something which can be evaluated to true includes things like `[1]`, `"some-nonempty-string"`, ... we don't need to this flexible at this state. It is easier to go from strict requirement to a more lenient requirement later, whereas the reversed direction is not backward compatible.

traitsui/tests/editors/test_boolean_editor.py
kitchoi: >  I'm not exactly sure what should be removed from the demo tests to simplify them now like you mentioned.

I was just wondering if you'd have preferred the demo to be different if the test logic is moved here, because after all there are some duplications. If what is in the demo test now is what you'd have written regardless of what is in `test_boolean_editor`, then it is totally fine.

> Also, as we discussed I added an IsChecked class to query to better be able to test trait change -> GUI change.

Thanks!

traitsui/examples/demo/Standard_Editors/tests/test_BooleanEditor_simple_demo.py
```diff
@@ -0,0 +1,61 @@
...
+            self.assertEqual(displayed_count_changes, '2')
+            self.assertEqual(displayed_count_changes, str(demo.count_changes))
+
+            demo.my_boolean_trait = True
+            displayed = readonly.inspect(DisplayedText())
+            self.assertEqual(displayed, "True")
+
+            displayed_count_changes = count_changes.inspect(DisplayedText())
+            self.assertEqual(displayed_count_changes, '3')
+            self.assertEqual(displayed_count_changes, str(demo.count_changes))
```
kitchoi, line 55: Perhaps change this to demonstrate the use of `IsChecked`?

kitchoi (commented):
LGTM. Thank you! Some documentation related comments.

kitchoi (approved):
Thank you!

PR #1283, opened 2020-09-25
Doc: Expand module and package docstrings for traitsui.testing

kitchoi:
This PR
- Expands the module/package docstrings for `traitsui.testing` where they are missing.
- Adds `README.txt` as symbolic links are added so that these descriptions are prominent and visible on GitHub.  This should help developers navigate the package to know what contains what.

Note that the `*.txt` are recursively included in the source distribution. Including these symlinks should not cause any real issues so I did nothing about skipping them. Just in case, I still tested building and installing from source distribution on Windows and there are no issues.


traitsui/testing/__init__.py
```diff
@@ -0,0 +1,19 @@
...
+#  Thanks for using Enthought open source!
+#
+
+"""
+This package provide functionality to help testing GUI applications built
+using TraitsUI.
+
+The top-level ``api`` module exposes the public API for external projects.
+
+Private modules in this package serves TraitsUI only.
```
aaronayres35, line 18: nit: "serves" should be "serve"

aaronayres35 (approved):
Changes LGTM! Just one nit picky comment.  But also, one general question.  Could we only display the module docstring in the readme.txt files? / Do we need the copyright on every single one?  In that case I imagine we wouldn't be able to use just the simple symbolic links, which might be more of a headache.  It is also very possible that including the copyrights in the read me is desired, it just seemed unnecessary to me.  Feel free to disregard, and merge.

kitchoi:
Thanks!

> Could we only display the module docstring in the readme.txt files? / Do we need the copyright on every single one? In that case I imagine we wouldn't be able to use just the simple symbolic links, which might be more of a headache. It is also very possible that including the copyrights in the read me is desired, it just seemed unnecessary to me. Feel free to disregard, and merge.

I think we want copyright header in all files in general, especially if we shall introduce linting for the copyright header, the linter will complain about the init files not having copyright headers. Exceptions can be made but this does not seem worth the extra handling / skips. We can reassess in the future if this gets in the way, though!

PR #1284, opened 2020-09-25
Process GUI events before resolving location

kitchoi:
This PR causes `UIWrapper.locate` (testing tool) to process the GUI events prior to resolving locations on the GUI.

Use case:
- When a trait is changed to cause GUI elements to change (e.g. the available values in the enum is dynamically computed), those updates may have been deferred for the GUI event loop to process (using dispatch="ui" in on_trait_change or observe). In that case, we should locate a GUI element only after those events are processed such that the GUI state is consistent.

Locating a GUI element is similar in concept compared to `UIWrapper.inspect`. `inspect` also needs to process GUI events in order to ensure the GUI is in a consistent state prior to getting the GUI state for testing.

We should only need to process events prior to locating the GUI element. Locate should have no side effect to require processing events before `locate` returns.

This is motivated by attempts to solve #1145 (will be in a separate PR). But the change here is needed regardless of the solution to #1145.

### DIFF
--- a/traitsui/testing/tester/tests/test_ui_wrapper.py
+++ b/traitsui/testing/tester/tests/test_ui_wrapper.py
@@ -151,6 +151,8 @@ class TestUIWrapperInteractionRegistries(unittest.TestCase):
         )
 
 
+# Use of locate requires the GUI event loop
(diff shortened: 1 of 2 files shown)


aaronayres35 (approved):
LGTM

PR #1286, opened 2020-09-25
Fix AttributeError when a nested UI is disposed

### LINKED ISSUE #907: Crash when using UI.dispose() with a TableEditor nested in a ListEditor

### LINKED ISSUE #1145: UI disposal may cause AttributeError on None

kitchoi:
Closes #1145
Closes #907

The solution follows what was described in https://github.com/enthought/traitsui/issues/1145#issuecomment-692152541
The AttributeError issue mainly bites Qt code. In order to fix it, the `destroy_control` has to be called earlier than `UI.reset`.
Moving the call down should cause the added test to fail.

Following this fix, the `destroy_control` is ~main~ made asynchronous for wx. The added test for wx is not for the AttributeError but for the RuntimeError that would have resulted if `destroy_control` was not made asynchronous. 

This PR:
- Add a failing test for #1145 for Qt, fixed by moving `toolkit().destroy_control` up in `UI.finish`
- Add a test to demonstrate we need to make `destroy_control` asynchronous in wx, fixed by wrapping `DestroyControl` and `DestroyChildren` in `wx.CallAfter`

(Note that since the wx test suite has not been passing, I had to first skip all the tests that are failing on master before working on this. This allowed me to identify new failing tests, which should be fixed by #1284.)


traitsui/tests/test_ui.py
```diff
@@ -87,6 +94,7 @@ def test_reset_with_destroy_wx(self):
 
             ui.reset(destroy=True)
 
+            process_cascade_events()
```
aaronayres35, line 97: What I think is happening is that because additions were made to wx `destroy_children` and `destroy_control` to use `wx.CallAfter`, when `toolkit().destroy_children(self.control)` is called in ui.reset this wont occur until after events are processed.  Is that correct?
What this makes me curious about is that this seems to be changing the behavior of calling `ui.reset` on wx, but we don't expect that to cause any other issues correct?
You mentioned on #1145:
"Alternatively, if deferring deletion for wx should cause issues, we could add another destroy_control_later on Toolkit instead, but I guess that is unlikely to be necessary."
So I take it this has already been considered and you believe it shouldn't be a problem.

traitsui/tests/test_ui.py
```diff
@@ -242,3 +250,200 @@ def test_invalid_state(self):
...
+
+    class CustomWidget(QtGui.QWidget):
+
+        def __init__(self, editor, parent=None):
+            super(CustomWidget, self).__init__()
+            self._some_editor = editor
+
+        def sizeHint(self):
+            assert self._some_editor.factory is not None
+            return super().sizeHint()
```
aaronayres35, line 278: I see that when moving the call to `destroy_control` down, the test fails on this assertion, but where exactly is this `sizeHint` getting called?

aaronayres35 (approved):
Changes LGTM, just had some questions.  I also confirmed that many of the scenarios that showcased this symptom as listed on #1145 no longer occur when run on this branch.  It might be good to have one more set of eyes on this PR though, as I'm still not 100% confident in my complete understanding of what all the code here is doing.

corranwebster (approved):
LGTM

traitsui/tests/test_ui.py
corranwebster: It's likely being called from Qt C++ code: as things are destroyed various widgets still in the layout are going to get asked how big they think they should be (which is what `sizeHint` does) and if the state is not in a good state then it will fail.

traitsui/tests/test_ui.py
corranwebster: Yes, this is a change of behaviour for Wx, but it brings it closer in-line to what Qt does, so the toolkit-agnostic code should be fine with it.  It is possible that some particular Wx widgets could have problems introduced by this, but it's not likely - problems are _usually_ caused by things not being there when you expect them to be, rather than the other way around, so delaying destruction shouldn't cause issues.

traitsui/tests/test_ui.py
kitchoi: Yep, what Corran's said. I should add better comments for this test code though... doing that now...

kitchoi:
Thank you both for the review! Merging...

PR #1298, opened 2020-10-01
Rename _BaseSourceWithLocation to BaseSourceWithLocation

kitchoi:
This PR renames a class in a private module to have a public name, as the class is imported by other modules in the same package. Since the module itself is privately named (so is its container package), the class is still considered private from the end user.


kitchoi:
Oops, I failed to pull master :( Fixing.

aaronayres35 (approved):
LGTM

PR #1300, opened 2020-10-01
Make event processing optional in UITester

### LINKED ISSUE #1297: Add flag to turn off event processing in UI Wrapper
There may be situations in which we don't want event handling to be done by UI Wrapper as it may already be handled by something else.  For example, if using UI Tester together with ModalDialogTester from pyface, both will independently try to process events which may cause problems.  Optionally disabling event processing on UI Wrapper would address this.

kitchoi:
Closes #1297

This PR adds a ~"process_events"~ `auto_process_events` flag on `UITester` (propagated to `UIWrapper`) to skip GUI event processing if the users request so.

- User manual is updated to highlight the specific care required when using UITester with ModalDialogTester.  
- Add a test against InstanceEditor that launches a UI as a modal dialog. ModalDialogTester and UITester are used in that test, which demonstrate the issue in #1297 if the flag is set to true.
- The flag is set while instantiating `UITester`, because it is convenient to be able to set it once.
- For correctness, `UITester.create_ui` will also respect this flag.

*Alternative*
- Instead of setting the flag in the object constructor, we can instead pass it through `perform`, `inspect` and `locate`. However I find that a little laborious.

*Question*:
- ~`process_events` here is a boolean, and its name does clash with `pyface.api.GUI.process_events`. The latter is a callable. A different naming suggestion would be very welcome.~ Edited: New name is.... `auto_process_events`!

Note: I have checked compatibility between `UITester` and `GuiTestAssistant.event_loop_until_condition`. This pairing isn't affected. The modal dialog is a special one here because the function that launches the modal dialog typically blocks until the dialog is closed.

### DIFF
--- a/docs/source/traitsui_user_manual/testing.rst
+++ b/docs/source/traitsui_user_manual/testing.rst
@@ -352,7 +352,9 @@ Example::
     modal_dialog_tester = ModalDialogTester(callable_to_open_dialog)
     modal_dialog_tester.open_and_run(when_opened)
 
-Note that you can instantiate as many |UITester| objects as you need.
+In the above example, ``ui`` is an instance of |UI| that has been obtained
+elsewhere in the test. Note that you can instantiate as many |UITester| objects
+as you need.
 
 .. rubric:: Is UITester compatible with PyFace GuiTestAssistant?
 
--- a/docs/source/traitsui_user_manual/testing.rst
+++ b/docs/source/traitsui_user_manual/testing.rst
@@ -333,8 +333,26 @@ it with any testing framework (e.g. unittest, pytest).
 
 .. rubric:: Is UITester compatible with PyFace ModalDialogTester?
 
-Yes. For example, you can use |UITester| to launch a modal dialog, and use
+Yes, with some care.
+
+For example, you can use |UITester| to launch a modal dialog, and use
 ModelDialogTester to close it.
+
+But if you are attempting to modify or inspect GUI states using |UITester|
+while the dialog is opened, you should set the ``process_events`` attribute to
+false for those operations. Otherwise the ModalDialogTester and UITester
+will enter a race condition that blocks forever.
+
+Example::
+
+    def when_opened(modal_dialog_tester):
+        ui_tester = UITester(process_events=False)
+        ui_tester.find_by_id(ui, "button").perform(MouseClick())
+
+    modal_dialog_tester = ModalDialogTester(callable_to_open_dialog)
+    modal_dialog_tester.open_and_run(when_opened)
+
+Note that you can instantiate as many |UITester| objects as you need.
 
 .. rubric:: Is UITester compatible with PyFace GuiTestAssistant?
 
--- a/docs/source/traitsui_user_manual/testing.rst
+++ b/docs/source/traitsui_user_manual/testing.rst
@@ -341,7 +341,7 @@ ModelDialogTester to close it.
 But if you are attempting to modify or inspect GUI states using |UITester|
 while the dialog is opened, you should set the ``process_events`` attribute to
 false for those operations. Otherwise the ModalDialogTester and UITester
-will enter a race condition that blocks forever.
+will enter a deadlock that blocks forever.
 
 Example::
 
--- a/docs/source/traitsui_user_manual/testing.rst
+++ b/docs/source/traitsui_user_manual/testing.rst
@@ -339,14 +339,14 @@ For example, you can use |UITester| to launch a modal dialog, and use
 ModelDialogTester to close it.
 
 But if you are attempting to modify or inspect GUI states using |UITester|
-while the dialog is opened, you should set the ``process_events`` attribute to
-false for those operations. Otherwise the ModalDialogTester and UITester
-will enter a deadlock that blocks forever.
+while the dialog is opened, you should set the ``auto_process_events``
+attribute to false for those operations. Otherwise the ModalDialogTester and
+UITester will enter a deadlock that blocks forever.
 
 Example::
 
     def when_opened(modal_dialog_tester):
-        ui_tester = UITester(process_events=False)
+        ui_tester = UITester(auto_process_events=False)
         ui_tester.find_by_id(ui, "button").perform(MouseClick())
 
     modal_dialog_tester = ModalDialogTester(callable_to_open_dialog)
--- a/traitsui/testing/tester/tests/test_ui_wrapper.py
+++ b/traitsui/testing/tester/tests/test_ui_wrapper.py
@@ -475,10 +475,10 @@ class TestUIWrapperEventProcessed(unittest.TestCase, UnittestTools):
 
         wrapper = example_ui_wrapper(
             registries=[StubRegistry(handler=handler)],
-            process_events=False,
+            auto_process_events=False,
         )
 
-        # With process_events set to False, events are not automatically
+        # With auto_process_events set to False, events are not automatically
         # processed.
         wrapper.perform(None)
         self.addCleanup(process_cascade_events)
@@ -497,12 +497,12 @@ class TestUIWrapperEventProcessed(unittest.TestCase, UnittestTools):
 
         wrapper = example_ui_wrapper(
             registries=[StubRegistry(solver=solver)],
-            process_events=False,
+            auto_process_events=False,
         )
 
-        # With process_events set to False, events are not automatically
+        # With auto_process_events set to False, events are not automatically
         # processed.
         new_wrapper = wrapper.locate(None)
 
         self.assertEqual(side_effect.call_count, 0)
-        self.assertFalse(new_wrapper._process_events)
+        self.assertFalse(new_wrapper._auto_process_events)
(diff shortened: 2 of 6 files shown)


traitsui/testing/tester/ui_tester.py
```diff
@@ -38,6 +38,12 @@ class UITester:
         Time delay (in ms) in which actions by the tester are performed. Note
         it is propagated through to created child wrappers. The delay allows
         visual confirmation test code is working as desired. Defaults to 0.
+    process_events : bool, optional
```
kitchoi, line 41: A more unique name would be good... `do_process`, `auto_process`, ...?

docs/source/traitsui_user_manual/testing.rst
```diff
@@ -333,9 +333,27 @@ it with any testing framework (e.g. unittest, pytest).
...
+But if you are attempting to modify or inspect GUI states using |UITester|
+while the dialog is opened, you should set the ``process_events`` attribute to
+false for those operations. Otherwise the ModalDialogTester and UITester
+will enter a race condition that blocks forever.
+
+Example::
+
+    def when_opened(modal_dialog_tester):
+        ui_tester = UITester(process_events=False)
+        ui_tester.find_by_id(ui, "button").perform(MouseClick())
```
aaronayres35, line 350: Idk if it is necessary, but it might be confusing where this `ui` comes from seeing as we don't use `with tester.create_ui() as ui:` here. We could add a brief sentence explaining that you will need access to it via a different route (e.g. I found myself getting it from modal_dialog_tester.get_dialog_widget()._ui).

traitsui/testing/tester/ui_wrapper.py
```diff
@@ -277,7 +285,10 @@ def _perform_or_inspect(self, interaction):
                 supported.extend(e.supported)
                 continue
             else:
-                with _event_processed():
+                context = (
+                    _event_processed if self._process_events else _nullcontext
+                )
+                with context():
                     return handler(self, interaction)
```
aaronayres35, line 292: why exactly is `_nullcontext` needed?  Could we just do:
```
if self._process_events:
    with _event_processed:
        return handler(self, interaction)
else:
    return handler(self, interaction)
```

Not saying we should do that even if it works.  The current code is cleaner.  I just didn't know if it was functionally necessary or not.  What exactly does a context manager that simply `yield`s do exactly?

traitsui/testing/tester/tests/test_ui_wrapper.py
```diff
@@ -463,3 +463,43 @@ def handler(wrapper, action):
...
+        wrapper = example_ui_wrapper(
+            registries=[StubRegistry(handler=handler)],
+            process_events=False,
+        )
+
+        # With process_events set to False, events are not automatically
+        # processed.
+        wrapper.perform(None)
+
+        self.assertEqual(side_effect.call_count, 0)
```
aaronayres35, line 484: In the test for UITester, you say
```
# We still want the events (if any) to be processed at the end of
# the test.
self.addCleanup(process_cascade_events)
```
Is that not he case here as well, because when `wrapper.perform(None)` is called, `process_cascade_events` won't be, but `handler` still will be.

traitsui/testing/tester/tests/test_ui_wrapper.py
```diff
@@ -463,3 +463,43 @@ def handler(wrapper, action):
...
+            registries=[StubRegistry(solver=solver)],
+            process_events=False,
+        )
+
+        # With process_events set to False, events are not automatically
+        # processed.
+        new_wrapper = wrapper.locate(None)
+
+        self.assertEqual(side_effect.call_count, 0)
+        self.assertFalse(new_wrapper._process_events)
```
aaronayres35, line 505: same comment as above

traitsui/tests/editors/test_instance_editor.py
```diff
@@ -85,8 +90,46 @@ def test_simple_editor_resynch_editor(self):
...
+
+        with ui_tester.create_ui(obj, dict(view=view)) as ui:
+
+            def click_button():
+                ui_tester.find_by_name(ui, "inst").perform(MouseClick())
+
+            def when_opened(tester):
+                with tester.capture_error():
+                    try:
+                        dialog_ui = tester.get_dialog_widget()._ui
```
aaronayres35, line 116: seems like this is the "standard way" to do this?  See my comment on the documentation

aaronayres35 (approved):
LGTM, just a few comments/questions

traitsui/testing/tester/ui_tester.py
aaronayres35: Between those I like `auto_process` more.  Thoughts on `auto_process_events`? I think it is okay if it is similar to the `process_events` callable as they are related, but they shouldn't be identical.  I feel like events carries some information though, but if you disagree I'm happy with either.

kitchoi: `auto_process_events` sounds good to me.
The default value is the most common use case so I'm happy for the name to be longer so that it is explicit and clear what it does.

traitsui/testing/tester/tests/test_ui_wrapper.py
kitchoi: You are right that I should not leave the event behind at the end of the test. I was being lazy here because the `side_effect` here is a local mock object that won't interact with other tests so it does not matter if it is called later, though I did leave the object on the stack for longer than it has to be. For the other test on `create_ui`, GUI widgets are created and then scheduled for disposal so if I don't make sure events are processed at the end of the tests, those GUI widgets could interact with the next test (through stealing focus / overlapping windows / etc...).

traitsui/tests/editors/test_instance_editor.py
kitchoi: I am reluctant to mention this in the documentation because the `_ui` reference looks really dodgy (the fact that it is privately named gives away some hints too!). What we have is a cycle reference from UI to a GUI toolkit widget back to the UI. Unfortunately this does seem to be the only way, _right_now_, for getting the UI via the ModalDialogTester.

A lot of times when we write tests for code written without tests, we have to add questionable references and hooks just so that we can add those tests, but then the end goal is that these tests will then allow us to refactor the code safely so that eventually we don't need these hooks any more.

docs/source/traitsui_user_manual/testing.rst
```diff
@@ -333,9 +333,27 @@ it with any testing framework (e.g. unittest, pytest).
...
-Yes. For example, you can use |UITester| to launch a modal dialog, and use
+Yes, with some care.
+
+For example, you can use |UITester| to launch a modal dialog, and use
 ModelDialogTester to close it.
 
+But if you are attempting to modify or inspect GUI states using |UITester|
+while the dialog is opened, you should set the ``process_events`` attribute to
+false for those operations. Otherwise the ModalDialogTester and UITester
+will enter a race condition that blocks forever.
```
kitchoi, line 344: The situation should be better described as a deadlock rather than a race condition.

traitsui/tests/editors/test_instance_editor.py
kitchoi: More thoughts on how it should have been:
In this block:
https://github.com/enthought/traitsui/blob/1a83c8a3b691ab204cd89e41ed810a2e7d21c3fe/traitsui/qt4/instance_editor.py#L429-L431
The `UI` instance is created within `edit_traits` and if `kind` is modal, the modal dialog is open and it blocks the function from returning until the dialog is closed. So there is no way to get to the UI from there before the dialog is closed. Ideally, the actions for creating an instance of UI and showing the blocking dialog should be separated so that we can keep a reference to the UI before allowing it to show and block. Then we would have been able to use the UITester to get to this dialog UI from the editor, instead of going via the toolkit control.

The reference to `_ui` is assigned here.
https://github.com/enthought/traitsui/blob/1a83c8a3b691ab204cd89e41ed810a2e7d21c3fe/traitsui/qt4/ui_panel.py#L72
https://github.com/enthought/traitsui/blob/1a83c8a3b691ab204cd89e41ed810a2e7d21c3fe/traitsui/wx/ui_panel.py#L100
It is not clear to me yet whose audience this private attribute is intended for, but I would guess it is not a sanctioned hooks for purposes like testing.

traitsui/testing/tester/ui_wrapper.py
kitchoi: The code block you showed is equivalent indeed. A context manager that simply `yield`s is an no-op context manager. It is useful for situations where the context manager is optional (our exact use case here). This is why `nullcontext` is added to `contextlib` in Python, but since we are still supporting Python 3.6, I just duplicate the logic here as `_nullcontext`. Once we drop Python 3.6, it can be replaced with the one from the standard library.
See also https://bugs.python.org/issue10049

I'd prefer the noop context to repeating the code block . Here we have just one line, but in some cases, the code could be bigger. Also, if we had duplicated the code block, it may not be obvious to future developers that they need to be maintained to be exactly the same - it would be bad because it risks introducing bugs when the code block is changed in one branch and not another.

docs/source/traitsui_user_manual/testing.rst
kitchoi: I added a sentence under the example, hope that looks okay.

traitsui/testing/tester/tests/test_ui_wrapper.py
kitchoi: Done adding cleanups for processing events.

PR #1303, opened 2020-10-05
Fix Qt button click failing to cause other widgets to update

### LINKED ISSUE #928: Qt TextEditor doesn't always refresh
When the value is edited in the handler of a button fire event, `TextEditor` value isn't updated until the focus is set on the editor (clicking on `click_counter` field). The same doesn't seem to happen to the `readonly` editor, but does happen with `simple`, `custom` and `text` style editors.

``` python
class Foo(HasTraits):
    my_button_trait = Button('Click Me')
    click_counter = Int(0)

    def _my_button_trait_fired(self):
        self.click_counter += 1

    traits_view = View(
        'my_button_trait',
        'click_counter',
        buttons=['OK'],
    )
```

### LINKED ISSUE #1245: TabularEditor doesn't refresh instantly upon appending an element to a list trait contained in the trait being displayed
Hi there ! It seems that `TabularEditor` doesn't refresh instantly if the trait it renders contains a list trait to which we append elements, as shown on this video:

refres_append

The video has been made using the following code snippet on python 3.6.10, traits 6.1, traitsui 7.0 and pyqt5 (5.14):

```Python3
from traits.api import (
    Button, HasTraits, Instance, List, Property, Str
)
from traitsui.api import (
    TabularAdapter, TabularEditor, UItem, View
)


class Foo(HasTraits):

    mylist = List(Str)


class FooAdapter(TabularAdapter):

    columns = [
        ('Name', 'name')
    ]

    name_content = Property(Str)

    def _get_name_content(self):
        return ', '.join(self.item.mylist)


class Bar(HasTraits):

    foo_list = List(Instance(Foo))

    selected = Instance(Foo)

    button = Button('Add')

        return View(
            UItem('button'),
            UItem(
                "foo_list",
                editor=TabularEditor(
                    adapter=FooAdapter(),
                    selected='selected',
                ),
            ),
            resizable=True,
        )

    def _button_fired(self):
        if self.selected is not None:
            self.selected.mylist.append('new')


if __name__ == "__main__":
    win = Bar(
        foo_list=[
            Foo(mylist=['test1']),
            Foo(mylist=['test2']),
        ]
    )
    win.configure_traits()

```

kitchoi:
Closes #928 
Closes #1245 (after causing `TabularEditor.update` to fire, from this [comment](https://github.com/enthought/traitsui/issues/1245#issuecomment-696062894)).

The issue affects Mac OSX and Qt only. This [macOSX specific limitation on Qt documentation](https://doc.qt.io/qt-5/macos-issues.html#using-native-cocoa-panels) seems relevant.

This issue is not specific to the textbox, as I can reproduce the same behaviour with a different widget, using just QPushButton and QLabel.

<details>
    <summary> Same strange behaviour using Two QPushButton and QLabel </summary>

```
import sys
from PyQt5 import QtWidgets, QtCore

class MainWindow(QtWidgets.QWidget):
    def __init__(self, ):
        super().__init__()

        self.count = 1
        self.label = QtWidgets.QLabel(self)

        self.button = QtWidgets.QPushButton('button', self)
        self.button.clicked.connect(self.on_click)

        # This second button is important for reproducing the behaviour, and it is important that it is
        # added to the layout _after_ the QLabel
        self.button2 = QtWidgets.QPushButton('do nothing', self)

        vbox = QtWidgets.QVBoxLayout()
        vbox.addWidget(self.button)
        vbox.addWidget(self.label)
        vbox.addWidget(self.button2)
        self.setLayout(vbox)

    def on_click(self):
        self.count += 1
        self.label.setText(str(self.count))

if __name__ == "__main__":
	app = QtWidgets.QApplication(sys.argv)
	window = MainWindow()
	window.show()
	sys.exit(app.exec_())
```
</details>

While this strikes me as a Qt issue rather than a TraitsUI issue, downstream projects using TraitsUI have been working around the issue by adding `GUI.process_events()` in the change handler for their buttons. Manually driving GUI event processing in application code should be avoided whenever possible. This is already causing a lot of inconvenience for developers on OSX, therefore I think it is worth fixing/workaround in TraitsUI.

This PR fixes / workarounds the issue by connecting the QPushButton click signal to its slot as a queued connection, following the approach suggested in the [Qt documentation](https://doc.qt.io/qt-5/macos-issues.html#using-native-cocoa-panels) mentioned above. This should not affect Windows nor Linux in production environment.

However **this does have implications for test code that calls the QPushButton click directly**. If there are test code that does something like this:
```
button_editor = ui.get_editors("trait_name")[0]
with self.assertTraitChange(obj, "trait_name"):
    button_editor.control.click()
```
This PR will cause these test code to break. The invocation to `click` (which is `QPushButton.click`) previously would have caused the click slot to be invoked immediately (see https://doc.qt.io/qt-5/qt.html#ConnectionType-enum), and the expected trait change would have occurred even without manually asking GUI events to be processed in the test. This can be resolved by adding `GUI.process_events` after calling `click()`, or use `GuiTestAssistant.event_loop_until_condition` or, use `UITester` which causes the GUI events to be processed after every interaction by default.

No tests can be written for this PR as this has to do with internal event dispatching between Qt and OSX.

I should file an issue upstream too, but that is going to take me a while... :/

### DIFF
--- a/traitsui/examples/demo/Advanced/Auto_update_TabularEditor_demo.py
+++ b/traitsui/examples/demo/Advanced/Auto_update_TabularEditor_demo.py
@@ -26,7 +26,7 @@ long lists are used, since enabling this feature adds and removed Traits
 listeners to each item in the list.
 
 """
-# Issues related to the demo warning: enthought/traitsui#913,
+# Issues related to the demo warning:
 # enthought/traitsui#960
 
 from traits.api import HasTraits, Str, Float, List, Instance, Button
--- a/traitsui/examples/demo/Standard_Editors/ButtonEditor_simple_demo.py
+++ b/traitsui/examples/demo/Standard_Editors/ButtonEditor_simple_demo.py
@@ -25,10 +25,6 @@ class ButtonEditorDemo(HasTraits):
     def _my_button_trait_fired(self):
         self.click_counter += 1
 
-    # Currently there is some erroneous behavior with Qt5 and OSX causing
-    # the click_counter to not immediately increment when the button is
-    # clicked. For more deailts, see enthought/traitsui #913.
-
     # Demo view:
     traits_view = View(
         'my_button_trait',
--- a/traitsui/examples/demo/Standard_Editors/CSVListEditor_demo.py
+++ b/traitsui/examples/demo/Standard_Editors/CSVListEditor_demo.py
@@ -1,19 +1,10 @@
 """
-**WARNING**
-
-  This demo might not work as expected and some documented features might be
-  missing.
-
--------------------------------------------------------------------------------
-
 Demonstrate the CSVListEditor class.
 
 This editor allows the user to enter a *single* line of input text, containing
 comma-separated values (or another separator may be specified). Your program
 specifies an element Trait type of Int, Float, Str, Enum, or Range.
 """
-# Issue related to the demo warning: enthought/traitsui#913
-
 
 from traits.api import (
     HasTraits, List, Int, Float, Enum, Range, Str, Button, Property
--- a/traitsui/examples/demo/Standard_Editors/ColorEditor_demo.py
+++ b/traitsui/examples/demo/Standard_Editors/ColorEditor_demo.py
@@ -13,7 +13,7 @@ Implementation of a ColorEditor demo plugin for Traits UI demo program.
 
 This demo shows each of the four styles of the ColorEditor
 """
-# Issues related to the demo warning: enthought/traitsui#913,
+# Issues related to the demo warning:
 # enthought/traitsui#946
 
 
--- a/traitsui/qt4/button_editor.py
+++ b/traitsui/qt4/button_editor.py
@@ -69,7 +69,14 @@ class SimpleEditor(Editor):
             self.control.setAutoDefault(False)
 
         self.sync_value(self.factory.label_value, "label", "from")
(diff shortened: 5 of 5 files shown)


kitchoi:
Similar workaround might be needed in ImageEnumEditor too, and possibly more. Sigh.

rahulporuri (approved):
LGTM. I don't see any issues whatsoever on Windows but the fix itself makes sense.

kitchoi:
Before merging, I should open an issue for considering removing type setting once it is no longer needed and then refer to the issue in the comment.

kitchoi:
@aaronayres35 Since you have a mac, do you mind sanity checking if this resolves the issues on your machine as well please?

aaronayres35:
> @aaronayres35 Since you have a mac, do you mind sanity checking if this resolves the issues on your machine as well please?

I ran the examples given in the description of 928 and comment of 1245 and both were working as expected on this branch,  LGTM!

corranwebster:
Just a note here that the [linked Qt limitation](https://doc.qt.io/qt-5/macos-issues.html#using-native-cocoa-panels) is not relevant for this as it is talking about embedding/integrating native (ie. Objective-C/Cocoa) widgets into a Qt GUI.  This does not match our situation.

The fix may still be correct, however.

kitchoi:
Thank you all!

PR #1309, opened 2020-10-05
Doc: Add links two more demo GUI tests to the user manual

kitchoi:
This PR adds links in the user manual to the recently added demo tests for BooleanEditor and RangeEditor.

I removed one of the links to the ButtonEditor demo test as it has so much in common with the other example that there is little value in referencing both.


rahulporuri (approved):
LGTM. Haven't tested the built docs though.

kitchoi:
Ah yes, I have tested the built doc :)

PR #1310, opened 2020-10-05
Fix QDateTime deprecation warning

### LINKED ISSUE #1096: QDateTime deprecation warnings with PySide2 5.15
Running the test suite with PySide2 5.15.0 emits the following warning:
```
test_datetime_editor_qt_datetime_out_of_bound (traitsui.tests.editors.test_datetime_editor.TestDatetimeEditorQt) ... /Users/kchoi/Work/ETS/traitsui/traitsui/tests/editors/test_datetime_editor.py:246: DeprecationWarning: QDateTime.QDateTime(const QDate & arg__1) is deprecated
  q_datetime = QDateTime(QDate(datetime.MAXYEAR + 1, 1, 1))
```

aaronayres35:
fixes #1096 

This PR is simply a one line change to pass a null `QTime` object as the second argument to the `QDateTime` constructor, so that it is using a non-deprecated signature.

### DIFF
--- a/traitsui/tests/editors/test_datetime_editor.py
+++ b/traitsui/tests/editors/test_datetime_editor.py
@@ -252,7 +252,7 @@ class TestDatetimeEditorQt(BaseTestMixin, GuiTestAssistant, unittest.TestCase):
             # the user set the datetime on the Qt widget to a value
             # too large for Python
             from pyface.qt.QtCore import QDateTime, QDate, QTime
-            q_datetime = QDateTime(QDate(datetime.MAXYEAR + 1, 1, 1), QTime())
+            q_datetime = QDateTime(QDate(datetime.MAXYEAR + 1, 1, 1), QTime(0,0))
             editor.control.setDateTime(q_datetime)
 
             # Get the displayed value back.
--- a/traitsui/tests/editors/test_datetime_editor.py
+++ b/traitsui/tests/editors/test_datetime_editor.py
@@ -252,7 +252,9 @@ class TestDatetimeEditorQt(BaseTestMixin, GuiTestAssistant, unittest.TestCase):
             # the user set the datetime on the Qt widget to a value
             # too large for Python
             from pyface.qt.QtCore import QDateTime, QDate, QTime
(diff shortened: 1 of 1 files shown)


traitsui/tests/editors/test_datetime_editor.py
```diff
@@ -251,8 +251,8 @@ def test_datetime_editor_qt_datetime_out_of_bound(self):
                 self.launch_editor(instance, view) as editor:
             # the user set the datetime on the Qt widget to a value
             # too large for Python
-            from pyface.qt.QtCore import QDateTime, QDate
-            q_datetime = QDateTime(QDate(datetime.MAXYEAR + 1, 1, 1))
+            from pyface.qt.QtCore import QDateTime, QDate, QTime
+            q_datetime = QDateTime(QDate(datetime.MAXYEAR + 1, 1, 1), QTime())
```
kitchoi, line 255: `QTime()` without any argument causes it to use the local time, which varies from location to location, run to run. While I understand the value isn't really relevant here (because the date itself is large enough such that the value is already out-of-bound no matter what time is), typically we want the tests to run in a predictable manner wherever the machine is. Could you specify the hour, minute, second and millisec to 0 please?

aaronayres35: I was under the impression from the docs that `QTime()` constructed a Null time object (i.e. `isNull` returns `True` and `isValid` returns `False`).  And then `QDateTime` docs on the constructor say "If date is valid and time is not, the time will be set to midnight."
I am happy to explicitly set `QTime(0,0)` as it is clearer, but I think they should have the same effect?

kitchoi: Ah!  I think I was reading the output from a different test, another test is using `datetime.now()`, not this one. Sorry about that. I think you are right, and it is still better to be explicit than be implicit. It makes the test easier to reason about.

kitchoi (approved):
LGTM (flake8 is complaining on CI)

traitsui/tests/editors/test_datetime_editor.py
kitchoi: But feel free to revert that if you'd prefer a null time... in other tests the time isn't explicitly set either.

PR #1311, opened 2020-10-05
Fix QDesktopWidget.availableGeometry deprecation warning

### LINKED ISSUE #1114: Deprecation warning for QDesktopWidget.availableGeometry
Running from Python 3.8 and PySide2 5.15 produce the following warning on OSX (with a single screen):

```
>>> import traitsui.qt4.constants
/Users/kchoi/Work/ETS/venvs/etsdemo-test/lib/python3.8/site-packages/traitsui/qt4/constants.py:48: DeprecationWarning: QDesktopWidget.availableGeometry(int screen) const is deprecated
  _geom = QtGui.QApplication.desktop().availableGeometry()
```

Qt documentation suggests using `screens()` to get a list of screens: https://doc.qt.io/qt-5/qdesktopwidget-obsolete.html#availableGeometry

For reference, calling `QtGui.QApplication.screens()` on my machine produces this:
```
>>> from pyface.qt import QtGui
>>> QtGui.QApplication.screens()
[<PySide2.QtGui.QScreen(0x7fcd0bb24710, name="DELL U2719D") at 0x7fcd0f800a40>]
```

aaronayres35:
fixes #1114 

This PR simply replaces `QtGui.QApplication.desktop()` with `QtGui.QApplication.screens()[0]` so that `.availableGeometry` can be called on a QScreen object (where it is not a deprecated method) rather than on a `QDesktopWidget` object.  

My only concern is I do not know how the list of screens obtained by `.screens()` is ordered in the case when multiple screens are being used.  The documentation [here][https://doc.qt.io/qt-5/qguiapplication.html#screens] simply says "Returns a list of all the screens associated with the windowing system the application is connected to."

In the code we actually run 
```
_geom = QtGui.QApplication.screens()[0].availableGeometry()

screen_dx = _geom.width()
screen_dy = _geom.height()
```
and I confirmed that before and after this change the values of `screen_dx` and `screen_dy` remained the same. If however I change the 0 to a 1, the values change (I am using 2 different screens currently).  Thus, I believe the code is correct as it stands (correct in the sense it does what was done before).

### DIFF
--- a/traitsui/qt4/constants.py
+++ b/traitsui/qt4/constants.py
@@ -45,7 +45,7 @@ del _palette
 
 # Screen size values:
 
-_geom = QtGui.QApplication.desktop().availableGeometry()
+_geom = QtGui.QApplication.screens()[0].availableGeometry()
 
 screen_dx = _geom.width()
 screen_dy = _geom.height()
--- a/traitsui/qt4/constants.py
+++ b/traitsui/qt4/constants.py
@@ -15,7 +15,7 @@ editors and text editor factories.
 """
 
 
-from pyface.qt import QtGui
+from pyface.qt import QtCore, QtGui
 
 
 _palette = QtGui.QApplication.palette()
@@ -45,7 +45,10 @@ del _palette
 
 # Screen size values:
 
-_geom = QtGui.QApplication.screens()[0].availableGeometry()
+if QtCore.__version_info__ >= (5, 0):
+    _geom = QtGui.QApplication.screens()[0].availableGeometry()
+else:
+    _geom = QtGui.QApplication.desktop().availableGeometry()
 
 screen_dx = _geom.width()
 screen_dy = _geom.height()
--- a/traitsui/qt4/constants.py
+++ b/traitsui/qt4/constants.py
@@ -45,6 +45,8 @@ del _palette
 
 # Screen size values:
 
+# QDesktopWidget.availableGeometry(int screen) is deprecated and Qt docs
+# suggest using screens() instead, but screens in not available in qt4
 if QtCore.__version_info__ >= (5, 0):
     _geom = QtGui.QApplication.screens()[0].availableGeometry()
 else:


aaronayres35:
Appveyor CI is failing with:

```
ERROR: traitsui (unittest.loader._FailedTest)
  File "/Users/appveyor/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/traitsui/__init__.py", line 84, in load_tests
    toolkit()
  File "/Users/appveyor/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/traitsui/toolkit.py", line 113, in toolkit
    _toolkit = find_toolkit("traitsui.toolkits")
  File "/Users/appveyor/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/pyface/base_toolkit.py", line 269, in find_toolkit
    return import_toolkit(ETSConfig.toolkit, entry_point)
  File "/Users/appveyor/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/pyface/base_toolkit.py", line 233, in import_toolkit
    raise RuntimeError(msg)
RuntimeError: No traitsui.toolkits plugin could be loaded for qt4
Ran 1 test in 0.000s
```

aaronayres35:
> CI is failing with:
> 
> ```
> ======================================================================
> ERROR: traitsui (unittest.loader._FailedTest)
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/Users/appveyor/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/traitsui/__init__.py", line 84, in load_tests
>     toolkit()
>   File "/Users/appveyor/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/traitsui/toolkit.py", line 113, in toolkit
>     _toolkit = find_toolkit("traitsui.toolkits")
>   File "/Users/appveyor/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/pyface/base_toolkit.py", line 269, in find_toolkit
>     return import_toolkit(ETSConfig.toolkit, entry_point)
>   File "/Users/appveyor/.edm/envs/traitsui-test-3.6-pyqt/lib/python3.6/site-packages/pyface/base_toolkit.py", line 233, in import_toolkit
>     raise RuntimeError(msg)
> RuntimeError: No traitsui.toolkits plugin could be loaded for qt4
> ----------------------------------------------------------------------
> Ran 1 test in 0.000s
> ```

Not sure what is causing this, I will investigate further tomorrow morning.

traitsui/qt4/constants.py
```diff
@@ -45,7 +45,12 @@
 
 # Screen size values:
 
-_geom = QtGui.QApplication.desktop().availableGeometry()
+# QDesktopWidget.availableGeometry(int screen) is deprecated and Qt docs
+# suggest using screens() instead, but screens in not available in qt4
+if QtCore.__version_info__ >= (5, 0):
```
kitchoi, line 50: `__version_info__` from Qt is actually a tuple with three values. While we can do `>` and `<` comparison with a smaller tuple, `==` does not work that way. The safest thing to do here might just be to take the first element of `__version_info__` so that comparison is not sensitive to the length of `__version_info__`.
(Python `sys.version_info` is a tuple with 5 entries.)

kitchoi: Oh no, taking it back. This is okay.

kitchoi (approved):
LGTM

traitsui/qt4/constants.py
```diff
@@ -45,7 +45,12 @@
...
+    _geom = QtGui.QApplication.screens()[0].availableGeometry()
+else:
+    _geom = QtGui.QApplication.desktop().availableGeometry()
```
rahulporuri, line 53: Can you open an issue to track and remove this compatibililty fix? I should have caught this earlier.

corranwebster:
I only just saw this - this should use `ScreenMetrics. screen_width ` and `ScreenMetrics.screen_height` from Pyface rather than re-implementing the logic.

kitchoi:
I think `ScreenMetrics.screen_width` has the same issue using deprecated interface though.

kitchoi:
i.e. if we use ScreenMetrics from pyface, the issue to do with the deprecation warning needs to be reopened until pyface also fixes the issue.

kitchoi:
Opened https://github.com/enthought/pyface/issues/718

aaronayres35:
One minor thing to note: the `pyface.system_metrics` implementation of this uses `screenGeometry` rather than `availableGeometry`.  Per the Qt docs `availableGeometry` "Returns the available geometry of the screen with index screen. What is available will be subrect of screenGeometry() based on what the platform decides is available (for example excludes the dock and menu bar on macOS, or the task bar on Windows)." 
So it seems there will be some discrepancy between the two (it very well could be completely inconsequential).  

I opened PR https://github.com/enthought/pyface/pull/719 to address the deprecation warnings, but as far as https://github.com/enthought/traitsui/issues/1316 I am not sure if I should directly use what is in pyface as it will change behavior very slightly.  I could add to the `pyface.system_metrics` to have different methods for getting available screen height/width, or I could just modify the current implementation to return the available sizes rather than the full screen sizes, or I could just scrap the available screen in traitsui and just use directly what is in pyface now.

On wx, the logic in `traitsui.wx.constants` and `pyface.ui.wx.system_metrics` is exactly the same in which they both use the full screen (I don't think there is any comparable concept of an "available screen" on wx).  Based on this I am planning to just go ahead and replace the qt availableGeometry with the screenGeometry() used in pyface.

PR #1313, opened 2020-10-06
Maint: Make a test using datetime more deterministic

kitchoi:
This PR does a simple clean up on a test that is using a different date time every time the test is run.


aaronayres35 (approved):
LGTM!

PR #1318, opened 2020-10-07
Fix key sequence not updating view per keystroke on wx

### LINKED ISSUE #1200: UI Tester key sequence delay mismatch between Wx and Qt
The current Qt implementation adds the delay to each typed character, whereas on wx/OSX there is a delay and then all the text is written at once.  This is fine for the purposes of most tests, but when using the delay parameter, ideally a user wants to see interactions occurring as if it were a user interacting with the UI.

kitchoi:
Fix #1200

This fixes the view not updating when KeySequence (on wx TextCtrl) writes each of the key characters. The view is not updated because when the for loop is busy, wx has not got a chance to process pending events. Note that this only affects visuals when one tries to debug a test visually.

### DIFF
--- a/traitsui/testing/tester/_ui_tester_registry/wx/_interaction_helpers.py
+++ b/traitsui/testing/tester/_ui_tester_registry/wx/_interaction_helpers.py
@@ -368,6 +368,7 @@ def key_sequence_text_ctrl(control, interaction, delay):
     for char in interaction.sequence:
         wx.MilliSleep(delay)
         control.WriteText(char)
+        wx.SafeYield()
 
 
 def key_click_slider(control, interaction, delay):


aaronayres35 (approved):
LGTM! Thank you for fixing this!

I did notice something that is separate from this PR, but seems incorrect.  In order to verify this change was working visually I added a delay to the following test in `test_text_editor.py`:
```
def test_simple_auto_set_update_text(self):
        foo = Foo()
        view = get_view(style="simple", auto_set=True)
        tester = UITester(delay=1000)
        with tester.create_ui(foo, dict(view=view)) as ui:
            with self.assertTraitChanges(foo, "name", count=3):
                name_field = tester.find_by_name(ui, "name")
                name_field.perform(KeySequence("NEW"))
                # with auto-set the displayed name should match the name trait
            display_name = name_field.inspect(DisplayedText())
            self.assertEqual(foo.name, "NEW")
            self.assertEqual(display_name, foo.name)
```

It looked correct on both qt, and wx with a delay between each key press in the sequence, so indeed this change did seem to fix the wx issue of appearing as a long delay and then all keys displayed at once.  But, I noticed that the time to run this test was `Ran 1 test in 6.306s` on Qt and `Ran 1 test in 3.400s` on wx.  This suggests to me that we may be double delaying on Qt?  I haven't looked into it super closely to see why this might be happening, but after I investigate further I can open a separate issue if needed.

kitchoi:
> It looked correct on both qt, and wx with a delay between each key press in the sequence, so indeed this change did seem to fix the wx issue of appearing as a long delay and then all keys displayed at once. But, I noticed that the time to run this test was Ran 1 test in 6.306s on Qt and Ran 1 test in 3.400s on wx. This suggests to me that we may be double delaying on Qt? I haven't looked into it super closely to see why this might be happening, but after I investigate further I can open a separate issue if needed.

Well spotted. That does look strange. Feel free to open an issue to note the observation even without further investigation.

PR #1320, opened 2020-10-07
UI Tester - add IsEnabled query class and implementations for Button Editors

### LINKED ISSUE #1294: Add a way to inspect if something is enabled - UI Tester
Very similar to #1291.  Currently we could implicitly check if a button for example is enabled by trying to click it, but me may want to see if changing some GUI state changes the button being enabled without having to try clicking it.

aaronayres35:
Closes #1294 

This PR adds a new `IsEditable` query class for UI Tester, along with implementations for the standard traitsui `ButtonEditor` and the `ui_base` `ButtonEditor`.  It also adds very simple check to the currently existing tests for these editors using the new query class.

### DIFF
--- a/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/button_editor.py
+++ b/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/button_editor.py
@@ -32,7 +32,7 @@ def register(registry):
                 wrapper._target.control, wrapper.delay)),
         (DisplayedText,
             lambda wrapper, _: wrapper._target.control.text()),
-        (IsEnabled, lambda wrapper, _: wrapper._target.control.isEnabled())
+        (IsEnabled, lambda wrapper, _: wrapper._target.control.isEnabled()),
     ]
 
     for target_class in [SimpleEditor, CustomEditor]:
--- a/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/button_editor.py
+++ b/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/button_editor.py
@@ -8,7 +8,7 @@
 #
 #  Thanks for using Enthought open source!
 #
-from traitsui.testing.api import DisplayedText, MouseClick
+from traitsui.testing.api import DisplayedText, IsEnabled, MouseClick
 from traitsui.testing.tester._ui_tester_registry.qt4 import (
     _interaction_helpers
 )
@@ -31,7 +31,8 @@ def register(registry):
             lambda wrapper, _: _interaction_helpers.mouse_click_qwidget(
                 wrapper._target.control, wrapper.delay)),
         (DisplayedText,
-            lambda wrapper, _: wrapper._target.control.text())
+            lambda wrapper, _: wrapper._target.control.text()),
+        (IsEnabled, lambda wrapper, _: wrapper._target.control.isEnabled())
     ]
 
     for target_class in [SimpleEditor, CustomEditor]:
--- a/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/ui_base.py
+++ b/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/ui_base.py
@@ -9,7 +9,7 @@
 #  Thanks for using Enthought open source!
 #
 from traitsui.qt4.ui_base import ButtonEditor
-from traitsui.testing.api import DisplayedText, MouseClick
+from traitsui.testing.api import DisplayedText, IsEnabled, MouseClick
 from traitsui.testing.tester._ui_tester_registry.qt4 import (
     _interaction_helpers
 )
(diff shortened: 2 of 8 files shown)


traitsui/testing/tester/_ui_tester_registry/wx/_traitsui/button_editor.py
```diff
@@ -68,6 +68,12 @@ def register(registry):
         handler=lambda wrapper, _: wrapper._target.control.GetLabel()
     )
 
+    registry.register_interaction(
+        target_class=SimpleEditor,
+        interaction_class=IsEnabled,
+        handler=lambda wrapper, _: wrapper._target.control.IsEnabled()
```
corranwebster, line 74: The fact that you are repeating the same sort of code multiple times here (`lambda wrapper, _: wrapper._target.control.<something>()`) sort of implies that a utility function might be useful here.  Perhaps a something like:
```
def call_control_method(method, *args, **kwargs):
    def handler(wrapper, _):
        bound_method = getattr(wrapper._target.control, method)
        return bound_method(*args, **kwargs)
    return handler
```
Which you can the use as:
```
handler=call_control_method(`IsEnabled`)
```

This design is a lot like the `attrgetter` and `itemgetter` functions from the Python standard library.

corranwebster: There are other ways that similar results can be achieved, but they are more magical.

traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/button_editor.py
```diff
@@ -31,7 +31,8 @@ def register(registry):
             lambda wrapper, _: _interaction_helpers.mouse_click_qwidget(
                 wrapper._target.control, wrapper.delay)),
         (DisplayedText,
-            lambda wrapper, _: wrapper._target.control.text())
+            lambda wrapper, _: wrapper._target.control.text()),
+        (IsEnabled, lambda wrapper, _: wrapper._target.control.isEnabled())
```
rahulporuri, line 35:
```suggestion
        (IsEnabled, lambda wrapper, _: wrapper._target.control.isEnabled()),
```
nit-picky comment : leaving a trailing comma on the last item in a list ensures that we won't see a diff like this one the next time.

rahulporuri: comment applies here and elsewhere in the PR.

traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/ui_base.py
```diff
@@ -30,7 +30,8 @@ def register(registry):
             (lambda wrapper, _: _interaction_helpers.mouse_click_qwidget(
                 wrapper._target.control, wrapper.delay))),
         (DisplayedText,
-            lambda wrapper, _: wrapper._target.control.text())
+            lambda wrapper, _: wrapper._target.control.text()),
+        (IsEnabled, lambda wrapper, _: wrapper._target.control.isEnabled())
```
rahulporuri, line 34:
```suggestion
        (IsEnabled, lambda wrapper, _: wrapper._target.control.isEnabled()),
```

traitsui/testing/tester/_ui_tester_registry/wx/_traitsui/button_editor.py
kitchoi: Drive-by comment: If you want to go for the above, please choose a name that is more obviously about creating a function that causes side-effect (e.g. `create_handler_for_control_method`), rather than a name that looks like it was going to cause a side-effect.

aaronayres35: @kitchoi this seems like a useful thing to add to me, I am just not sure the best place for it to sit, as it could be used in other editor implementations for both toolkits.  It would make sense to me to go in a higher level `_interaction_helpers` module but that does not currently exist, and it feels incorrect to put the same function in both toolkit specific `_interaction_helpers` modules.  Any thoughts?  I would be happy to add a module in this PR for it to sit, or we could just open an issue (the scope of this change reaches beyond this PR) and address it in a separate PR.

kitchoi: > It would make sense to me to go in a higher level _interaction_helpers module but that does not currently exist, and it feels incorrect to put the same function in both toolkit specific _interaction_helpers modules. Any thoughts? I would be happy to add a module in this PR for it to sit

What you said about a new module makes sense if we are to add this function.
I think the function isn't doing a tons of things and I find the lambda function readable. Considering that it will have to live in a different file, if the new function is readable enough such that others reading the code does not have to find the function to understand what it is doing then it is worth the redirection.  
I will leave this to you whether to do it here.

kitchoi (approved):
LGTM.

See comment above about refactoring the creation of a handler that calls a method on the control. If it is implemented here, I am happy to review again, or we can open an issue or have it implemented in a separate PR like you said.

kitchoi:
Appveyor keeps thinking there is a new build started for this but I could not understand why. I tried triggering a rebuild and it would not budge. Another trick maybe to close and open this PR again.

aaronayres35:
> LGTM.
> 
> See comment above about refactoring the creation of a handler that calls a method on the control. If it is implemented here, I am happy to review again, or we can open an issue or have it implemented in a separate PR like you said.

I opened a separate issue #1328 regarding this.  I think it can be addressed in its own PR, and I will merge this PR as it stands now.

PR #1322, opened 2020-10-07
Replace screen metrics code with Pyface implementation

### LINKED ISSUE #1316: Remove duplicated screen metrics code
Pyface has (and has had for a long time) a class `pyface.screen_metrics.ScreenMetrics` that provides basic access to the screen size and standard application background colour.  TraitsUI has ad-hoc duplication of this code in `traitsui.qt.constants` and `traitsui.wx.constants`.

The TraitsUI code should just import and use the Pyface class for these values.

aaronayres35:
fixes #1316 

~~This PR removes the screen size from the `traitsui.qt4.constants.py` and `traitsui.wx.constants.py` files~~ (Edit: they have been added back in and use pyface's `SystemMetrics`.  They were added back as they had been public and it is possible they may be used elsewhere). Note pyface's `SystemMetrics`  is now used to get these values whenever they are needed (only a few places) within traitsui, rather than getting them from `constants.py`.  I.e. the constants which remain in the `constants.py` files are not used in traitsui itself, but could still potentially be being used elsewhere.   

Note that this will reintroduce deprecation warnings which were resolved by #1311 ~~until https://github.com/enthought/pyface/pull/719 is merged on pyface~~ (Edit: it is now merged) and a release is made so those changes are accessible in traitsui.

Additionally, see this comment: https://github.com/enthought/traitsui/pull/1311#issuecomment-705206248 for why this PR may very slightly change behavior.  If this change is something we wish to avoid I am more than happy to modify this PR as needed.

### DIFF
--- a/traitsui/qt4/constants.py
+++ b/traitsui/qt4/constants.py
@@ -16,8 +16,6 @@ editors and text editor factories.
 
 
 from pyface.qt import QtGui
-from pyface.api import SystemMetrics
-
 
 _palette = QtGui.QApplication.palette()
 
--- a/traitsui/qt4/constants.py
+++ b/traitsui/qt4/constants.py
@@ -43,8 +43,3 @@ EditableColor = _palette.color(QtGui.QPalette.Base)
 WindowColor = _palette.color(QtGui.QPalette.Window)
 
 del _palette
-
-# Screen size values:
-
-screen_dx = SystemMetrics()._get_screen_width()
-screen_dy = SystemMetrics()._get_screen_height()
--- a/traitsui/qt4/constants.py
+++ b/traitsui/qt4/constants.py
@@ -15,7 +15,8 @@ editors and text editor factories.
 """
 
 
-from pyface.qt import QtCore, QtGui
+from pyface.qt import QtGui
+from pyface.api import SystemMetrics
 
 
 _palette = QtGui.QApplication.palette()
@@ -45,14 +46,5 @@ del _palette
 
 # Screen size values:
 
-# QDesktopWidget.availableGeometry(int screen) is deprecated and Qt docs
-# suggest using screens() instead, but screens in not available in qt4
-if QtCore.__version_info__ >= (5, 0):
-    _geom = QtGui.QApplication.screens()[0].availableGeometry()
-else:
-    _geom = QtGui.QApplication.desktop().availableGeometry()
-
-screen_dx = _geom.width()
-screen_dy = _geom.height()
-
-del _geom
+screen_dx = SystemMetrics()._get_screen_width()
+screen_dy = SystemMetrics()._get_screen_height()
--- a/traitsui/qt4/constants.py
+++ b/traitsui/qt4/constants.py
@@ -17,6 +17,8 @@ editors and text editor factories.
 
 from pyface.qt import QtGui
 
+from pyface.api import SystemMetrics
+
 _palette = QtGui.QApplication.palette()
 
 # Default dialog title
@@ -41,3 +43,9 @@ EditableColor = _palette.color(QtGui.QPalette.Base)
 WindowColor = _palette.color(QtGui.QPalette.Window)
 
 del _palette
+
+#: Screen width
+screen_dx = SystemMetrics().screen_width
+
+#: Screen height
+screen_dy = SystemMetrics().screen_height
--- a/traitsui/qt4/toolkit.py
+++ b/traitsui/qt4/toolkit.py
@@ -32,12 +32,6 @@ from pyface.qt import QtCore, QtGui, qt_api
 from traitsui.toolkit import Toolkit
 
 
-#: Screen width
-screen_dx = SystemMetrics().screen_width
-
-#: Screen height
-screen_dy = SystemMetrics().screen_height
-
 # -------------------------------------------------------------------------
 #  Handles UI notification handler requests that occur on a thread other than
 #  the UI thread:
@@ -250,8 +244,8 @@ class GUIToolkit(Toolkit):
         if parent is None:
             px = 0
             py = 0
-            pdx = screen_dx
-            pdy = screen_dy
+            pdx = SystemMetrics().screen_width
+            pdy = SystemMetrics().screen_height
         else:
             pos = parent.pos()
             if int(parent.windowFlags()) & QtCore.Qt.Window == 0:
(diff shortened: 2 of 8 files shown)


traitsui/qt4/toolkit.py
```diff
@@ -26,10 +26,17 @@
...
 from traits.trait_notifiers import set_ui_handler
+from pyface.api import SystemMetrics
 from pyface.qt import QtCore, QtGui, qt_api
 
 from traitsui.toolkit import Toolkit
-from .constants import screen_dx, screen_dy
+
+
+#: Screen width
+screen_dx = SystemMetrics()._get_screen_width()
```
corranwebster, line 36: This is the getter for a property, and so shouldn't be called.  Just access `.screen_width`.
```suggestion
screen_dx = SystemMetrics().screen_width
```

corranwebster:
You'll need to make the same change throughout.

traitsui/qt4/constants.py
```diff
@@ -42,17 +41,3 @@
...
-
-# QDesktopWidget.availableGeometry(int screen) is deprecated and Qt docs
-# suggest using screens() instead, but screens in not available in qt4
-if QtCore.__version_info__ >= (5, 0):
-    _geom = QtGui.QApplication.screens()[0].availableGeometry()
-else:
-    _geom = QtGui.QApplication.desktop().availableGeometry()
-
-screen_dx = _geom.width()
-screen_dy = _geom.height()
```
kitchoi, line 56: With the module being publicly named, and the variable being equally public, I am not sure if there are no ones out there using these variables. @corranwebster are you sure we could remove them here?

traitsui/qt4/toolkit.py
```diff
@@ -26,10 +26,17 @@
...
+
+#: Screen height
+screen_dy = SystemMetrics()._get_screen_height()
```
kitchoi, line 39: I believe these new definitions are in response to this [comment](https://github.com/enthought/traitsui/issues/1316#issuecomment-704903499):

> A stretch goal would be to remove these values from constants.py and just use the Pyface class directly wherever it is needed (which I am guessing is only a few places).

The current move does not change when the screen width/height are evaluated. The dimensions are still evaluated at import time. 
I think what Corran had meant was to get the screen dimensions where they are used, e.g here:
https://github.com/enthought/traitsui/blob/32458481d7a38f3d4811020fcf1837ea1b8088c2/traitsui/wx/helper.py#L231
Such move is more likely to cause behavioural change:  currently the screen size is computed when the code is loaded (e.g. during `import traitsui.api`), the move to inline code will cause the screen size to be computed when it is needed, which could be much later, like when a widget is painted. This may well be a good and a correct thing to do, but is one that requires a bit more thoughts and evaluations.

@corranwebster Did I understand your comment correctly?

traitsui/qt4/constants.py
kitchoi: To err on the side of caution, shall we keep these constants here?

aaronayres35: I'm happy to leave them for now and just leave the issue open or open a new more specific issue to remove them from constants later

traitsui/qt4/toolkit.py
aaronayres35: Ah I see.  I had started to do this initially, but second guessed if it was the right thing.  What you described makes sense to me so I am happy to make that change now. But we can wait for Corran's response too if you prefer.

kitchoi: I am fairly sure that was what @corranwebster meant, blame it on me if I get it wrong then :)

kitchoi (approved):
LGTM.
For reference, the code being replaced is strictly speaking not a duplication because it was accessing `QScreen.availableGeometry` before and it is now accessing `QScreen.geometry`. See documentation [here](https://doc.qt.io/qt-5/qscreen.html#availableGeometry-prop):
 > availableGeometry : const QRect
>This property holds the screen's available geometry in pixels
>
> The available geometry is the geometry excluding window manager reserved areas such as task bars and system menus.
>
> Note, on X11 this will return the true available geometry only on systems with one monitor and if window manager has set _NET_WORKAREA atom. In all other cases this is equal to geometry(). This is a limitation in X11 window manager specification.

aaronayres35:
> LGTM.
> For reference, the code being replaced is strictly speaking not a duplication because it was accessing `QScreen.availableGeometry` before and it is now accessing `QScreen.geometry`. See documentation [here](https://doc.qt.io/qt-5/qscreen.html#availableGeometry-prop):

Yeah, I wanted to ask about that to be sure it is okay that this may be changing behavior slightly.  See https://github.com/enthought/traitsui/pull/1311#issuecomment-705206248 
Given your approval though, I am assuming that the change is okay?

kitchoi:
> Yeah, I wanted to ask about that to be sure it is okay that this may be changing behavior slightly. See #1311 (comment)
Given your approval though, I am assuming that the change is okay?

I am not sure the change is okay actually. I think I am approving on the basis that the code is correct for the purpose of #1316. I'd prefer not changing behaviour if I am not sure this is okay.

aaronayres35:
> > Yeah, I wanted to ask about that to be sure it is okay that this may be changing behavior slightly. See #1311 (comment)
> > Given your approval though, I am assuming that the change is okay?
> 
> I am not sure the change is okay actually. I think I am approving on the basis that the code is correct for the purpose of #1316. I'd prefer not changing behaviour if I am not sure this is okay.

I will hold off on merging this now then.  We can discuss at the standup

traitsui/qt4/toolkit.py
corranwebster: Yes, I was suggesting what Kit is suggesting above.  It's sort of orthogonal to the PR, but it will lead to slightly better behaviour if screen size changes.

kitchoi:
Retracting review approval so we don't accidentally merge this...

kitchoi:
On OSX, in a screen where I have my dock permanently visible at the bottom, the `screen_dy` is larger on this branch than on master.

On master:
```
>>> constants.screen_dx, constants.screen_dy
(2560, 1348)
```
Here:
```
>>> constants.screen_dx, constants.screen_dy
(2560, 1440)
```

This will therefore change the actual pixel use for `width` and `height` for relative number between 0-1.  For applications that use relative sizing consistently, the application will be consistently larger on this branch than before. For applications that use a mix of relative size and absolute size (e.g. `width=500`) and try to fiddle with the number to get almost pixel perfect alignment, this change could lead to misalignments.

I will try to come up with a small example...

corranwebster:
> For reference, the code being replaced is strictly speaking not a duplication because it was accessing `QScreen.availableGeometry` before and it is now accessing `QScreen.geometry`. See documentation [here](https://doc.qt.io/qt-5/qscreen.html#availableGeometry-prop):
> 

This is a significant change.  The Pyface code should be changed to use `availableGeometry` since that's almost certainly what we will want in every case (ie. how much room is available for a standard window to be shown).

kitchoi (approved):
https://github.com/enthought/pyface/pull/727 is now merged so the behaviour is preserved (for TraitsUI world)

PR #1325, opened 2020-10-08
Fix SetEditor on Qt and OSX not updating view after button clicks

### LINKED ISSUE #1305: Moving items in SetEditor does not update list / causes phantom view on OSX
Using this demo example: `examples/demo/Standard_Editors/SetEditor_demo.py` 
On OSX, if I move items from available to selected using the buttons, the list is not updated immediately. And when I try to click it, it is only partially updated.
Oct-05-2020 15-48-48

A treatment similar to #1303 can be applied to this line:
https://github.com/enthought/traitsui/blob/32458481d7a38f3d4811020fcf1837ea1b8088c2/traitsui/qt4/set_editor.py#L162
(i.e. adding `type=QtCore.Qt.QueuedConnect` seems to fix it)

Similar symptoms are seen in #928 and #1245. I have not tested this on Windows or Linux yet.

Platform: OSX 10.15.5
Python version: 3.6 and 3.8
Python/Qt wrapper: PySide2 (5.15.1) and PyQt5 (5.14.2)

kitchoi:
Closes #1305

The solution is exactly the same as #1303. Again, no tests can be written for this. 
I reproduced the issue on OSX 10.15.5 (with both PyQt5 and PySide2) and then verified that this change fixes the issue.

Like in #1303, this change has implications for tests that try to call `click` directly without causing the event loop to process pending events before checking for state change. Currently tests in TraitsUI are either using UITester that handles this automatically, or have already been driving the event loop processing manually. Therefore the tests here do not require changes.

### DIFF
--- a/traitsui/examples/demo/Standard_Editors/SetEditor_demo.py
+++ b/traitsui/examples/demo/Standard_Editors/SetEditor_demo.py
@@ -2,13 +2,6 @@
 #  License: BSD Style.
 
 """
-**WARNING**
-
-  This demo might not work as expected and some documented features might be
-  missing.
-
--------------------------------------------------------------------------------
-
 Implementation of a SetEditor demo plugin for the Traits UI demo program.
 
 The four tabs of this demo show variations on the interface as follows:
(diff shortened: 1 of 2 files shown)


kitchoi:
> I reproduced the issue on OSX 10.15.5 (with both PyQt5 and PySide2) and then verified that this change fixes the issue.

I should add that I have also confirmed that particular logic about connecting the `clicked` slot is already tested, e.g. if I comment the entire line, lots of tests fail.

rahulporuri (approved):
LGTM

PR #1326, opened 2020-10-08
Fix ImageEnumEditor button not causing other widget to updates on OSX and Qt

### LINKED ISSUE #1304: Clicking buttons in Qt Custom ImageEnumEditor causes weird repainting on OSX
For example, using the demo example `examples/demo/Standard_Editors/ImageEnumEditor_demo.py`
If I click an item in the "custom" editor, then I expect the other view to update, but what I saw, on OSX 10.15.5, is that the other view is updated when I click it, but with some kind of phantom view.
Oct-05-2020 15-34-07

I believe the same fix/workaround as in #1303 can be applied here:
https://github.com/enthought/traitsui/blob/806808f5ab9ab23e2f7fe5e5a874d5e19b200081/traitsui/qt4/enum_editor.py#L379
(i.e. add `type=QtCore.Qt.QueuedConnection` to the `connect` call).

Platform: OSX 10.15.5
Python version: 3.6 and 3.8
Python/Qt wrapper: PySide2 (5.15.1) and PyQt5 (5.14.2)

kitchoi:
Closes #1304 

The solution is exactly the same as #1303.
I reproduced the issue on OSX 10.15.5 (with both PyQt5 and PySide2) and then verified that this change fixes the issue.

Again, this has implications for tests that don't run the GUI event loop between clicking the button and checking for trait change.

In TraitsUI tests, we have tests using `UITester` for this logic:
https://github.com/enthought/traitsui/blob/a79d7f387bb8d697aff6134e9cdb2a0e81379719/traitsui/tests/editors/test_enum_editor.py#L378-L379
`perform` already does event processing, so the test is happy there.

We also have tests that have yet to be converted to use UITester, and it does the event processing:
https://github.com/enthought/traitsui/blob/964c0110112adc9dc35cf5c6f7f6eedf7ed28fc7/traitsui/tests/editors/test_image_enum_editor.py#L422-L425
So this test also does not require changes. Just to note, if I remove the `process_cascade_events` in this particular test, this PR would break the test.

### DIFF
--- a/traitsui/examples/demo/Standard_Editors/ImageEnumEditor_demo.py
+++ b/traitsui/examples/demo/Standard_Editors/ImageEnumEditor_demo.py
@@ -14,7 +14,7 @@ program.
 
 This demo shows each of the four styles of the ImageEnumEditor.
 """
-# Issues related to the demo warning: enthought/traitsui#913,
+# Issues related to the demo warning:
 # enthought/traitsui#947
 
 
--- a/traitsui/qt4/enum_editor.py
+++ b/traitsui/qt4/enum_editor.py
@@ -376,7 +376,11 @@ class RadioEditor(BaseEditor):
 
                     rb.setChecked(name == cur_name)
 
-                    rb.clicked.connect(self._mapper.map)
+                    # The connection type is set to workaround Qt5 + MacOSX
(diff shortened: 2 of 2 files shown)


rahulporuri (approved):
LGTM

PR #1327, opened 2020-10-08
Rename tester registry module to target_registry 

kitchoi:
The class defined by the module is called `TargetRegistry`. By convention we would have named the module `target_registry` as well. This also results in a module name that is more distinct and therefore easier to find / reference in code or in one's mind.


rahulporuri (approved):
LGTM. Merge once CI is happy.

rahulporuri:
martin fowler would have probably liked it if

- you created an empty module called target_registry first,
- import from registry in target_registry
- updated imports to target_registry
- finally, move the actual implementation to target_registry from registry
- and then remove registry.

kitchoi:
🙈  ~Pretend I had done all that and squashed/reordered all the commits~ No I haven't done those tiny steps.

PR #1354, opened 2020-10-12
Doc: Improve documentation on the UIWrapper._target being protected

kitchoi:
This PR improves the documentation to clarify that `UIWrapper._target` is a protected (or internal?) attribute that is neither public or private. While it is expected to be used, it should be used with discretion.

The attribute is expected to be used for extending the API (it is mentioned in the "Advanced Testing" section). It deliberately begs the questions "are you sure you want to use it"  / "could you do this differently" / "should you request the support to be implemented upstream". Whether it is appropriate to be used depends on the context, e.g. it may expose the implementation details that do not belong to the same project, or the implementation details may be maintained in the same place such that it is less of a problem.

The documentation here is also useful for developers maintaining the `UIWrapper`: this particular attribute should NOT be renamed lightly once the API is published.

### DIFF
--- a/docs/source/traitsui_user_manual/testing.rst
+++ b/docs/source/traitsui_user_manual/testing.rst
@@ -380,8 +380,8 @@ The API allows extension such that
    code are not maintained together, tests may be subject to breakages caused
    by internal changes of the UI editors. Certain attributes described in this
    document are classified as protected, to be used for extending the testing
-   API. Developers should evaluate based on context whether using these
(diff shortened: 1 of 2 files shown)


rahulporuri (approved):
LGTM

kitchoi:
It seems that Appveyor has been randomly aborting jobs with no log output.
Screenshot 2020-10-12 at 11 43 22

Screenshot 2020-10-12 at 11 43 28

PR #1360, opened 2020-10-13
Update pyface version being run on CI

kitchoi:
[Targeting 7.1 maintenance branch]

This PR updates this CI source requirement such that Pyface maint/7.1 branch is used in tests.


rahulporuri:
looks like you'll need to update travis/appveyor config to run tests on the maint/7.1 branch.

kitchoi:
Ah, forgot about that. Good point. Thanks.

PR #1363, opened 2020-10-14
 Add 'Notes on Upgrading' to the Changelog for 7.1

kitchoi:
[Targeting maint/7.1]

This PR adds a section in the changelog to remind users about the potential impact on tests due to the fix/workaround to Qt PushButton (and that now they may try to remove their workarounds).


rahulporuri (approved):
LGTM. This will need to be backported to master, right?

kitchoi:
> This will need to be backported to master, right?

Yep, it is one of the post release steps.

PR #1366, opened 2020-10-15
Update setup.py to allow prerelease version

### LINKED ISSUE #1357: Update version computation logic for release candidates
It would be nice if the release candidate can say in its version that it is a release candidate, e.g. 
```
>>> from traitsui import __version__
>>> __version__
"7.1.0rc1"
```
Currently any version apart from the release has a suffix of "dev{num}".

kitchoi:
Closes #1357

This PR is very similar to that one in pyface: https://github.com/enthought/pyface/pull/747

The logic to do with computing git revisions have not changed (the code is the same as that in Pyface, but different from that in Traits).

Main changes

- The computation of version now includes a prerelease segment.
- `version` is now the same as `full_version`. In `traits.__init__` we have been using `full_version` already so this does not change behaviour. The documentation has been getting only the major and minor version from the string, so this also has no impact on that.
- _version.py is always updated in the development environment when setup.py is run. Previously it is only updated if the file was not found.


rahulporuri:
`ARCHIVE_COMMIT_HASH` won't work as expected if the `.gitattributes` file doesn't exist in the git source repository/directory - https://github.com/enthought/traits/blob/master/.gitattributes. See original PR that added this code - https://github.com/enthought/traits/pull/526

kitchoi:
> ARCHIVE_COMMIT_HASH won't work as expected if the .gitattributes file doesn't exist in the git source repository/directory

Thank you! The archive version logic is the one that I really struggle to write test for, so it was untested in my set of tests (see https://github.com/enthought/pyface/pull/747). I should know that wherever I am taking a leap of faith, this is whether it is going to fail 😓

rahulporuri:
also, bad on me because I originally reviewed these PRs when they were added to traits and i forgot about that when i reviewed the pyface pr 😓

rahulporuri (approved):
LGTM. I haven't tested this locally because I presume this is just a copy-paste of the pyface changes.

PR #1367, opened 2020-10-15
Add IsEnabled to the testing api docstring

kitchoi:
This PR adds a reference to the new `IsEnabled` query in the api module docstring.

### DIFF
--- a/traitsui/testing/api.py
+++ b/traitsui/testing/api.py
@@ -32,6 +32,7 @@ Interactions (for getting GUI states)
 
 - :class:`~.DisplayedText`
 - :class:`~.IsChecked`
+- :class:`~.IsEnabled`
 - :class:`~.SelectedText`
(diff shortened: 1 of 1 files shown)


aaronayres35 (approved):
LGTM!  Completely orthogonal to this PR, this change just made me think of it, but I believe we don't currently have a `test_api.py` to make sure all the imports succeed like we do in other packages.  Should we add one?

kitchoi:
>  I believe we don't currently have a test_api.py to make sure all the imports succeed like we do in other packages. Should we add one?

Yep it would make sense to have one!

aaronayres35:
> > I believe we don't currently have a test_api.py to make sure all the imports succeed like we do in other packages. Should we add one?
> 
> Yep it would make sense to have one!

Done see #1370

PR #1368, opened 2020-10-15
Refresh developer command for generating API documentation

### LINKED ISSUE #996: Don't document tests in API docs.
See for example https://docs.enthought.com/traitsui/api/traitsui.tests.html. Similar to issue in pyface https://github.com/enthought/pyface/issues/571

kitchoi:
This PR
- Remove all except one `*.rst` in `docs/source/api` folder.
- The original `docs/source/api/index.rst` is kept to provide the cross reference back to User Manual. The TOC path is modified.
- Update the `docs` command in `etstool.py` to include `sphinx-apidoc` step. Generated files go into a folder called "autogenerated" inside `docs/source/api`, which is then skipped in `.gitignore`.
- Drive-by update to the docstring of `etstool.py` on how to create documentation.
- Closes #996 by skipping `tests` packages in the apidoc command.

I have compared the newly generated files with the existing ones, and I could NOT find evidence of manual curation apart from the ones for `traitsui.testing`. The ones for the testing package were created manually because back then I just assumed committed API source mean they are manually maintained. So this does change how the API documentation looks for the testing package somewhat (aka the TOC is a bit more verbose), but not a lot.

This is how it used to look:
Screenshot 2020-10-15 at 18 08 34

This is how it looks on this branch (I generated this on master, hence the "7.2" heading):
Screenshot 2020-10-15 at 18 09 19


aaronayres35 (approved):
LGTM!
One thing though: I still see load_tests, like we were seeing in pyface.  Would we like to get rid of that for traitsui as well?
Screen Shot 2020-10-15 at 12 53 57 PM
I imagine we could do effectively the same fix as is done https://github.com/enthought/pyface/pull/762/files ?

kitchoi:
> One thing though: I still see load_tests, like we were seeing in pyface. Would we like to get rid of that for traitsui as well?
> I imagine we could do effectively the same fix as is done https://github.com/enthought/pyface/pull/762/files ?

Yep!

PR #1369, opened 2020-10-15
Allow docs command to be run on Windows

### LINKED ISSUE #997: "docs" etstool command doesn't work on Windows
Similar to https://github.com/enthought/pyface/issues/564. This is because of our use of `make` instead of using `-m sphinx ...`.

https://github.com/enthought/traitsui/blob/0a4a6f418d01f4d6e24a84eb5995202217c8e66c/etstool.py#L334-L336

kitchoi:
The current `docs` command in `etstool.py` cannot be reliably run on Windows as reported by @rahulporuri 
While I could not reproduce, we can avoid the issue by using the sphinx command directly.

The make files are kept, I believe some developers still prefer to use them(?). It does risk the two going out-of-sync though.

Closes #997

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -376,12 +376,15 @@ def docs(runtime, toolkit, environment):
     click.echo('Done installing documentation tools')
 
     os.chdir('docs')
-    commands = [
-        "edm run -e {environment} -- make html",
-    ]
+    command = (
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM 🙏

rahulporuri:
fixes #997

PR #1370, opened 2020-10-15
Add a test for traitsui.testing.api

aaronayres35:
This PR adds simple import tests for the `traitsui.testing.api`.  I broke the tests up into sections to try to make it easier to ensure the test is up to date with the api moving forward.  

In doing so I also thought it would make sense to reorganize the ordering of the groups of imports in the api module to match the docstring (and have the tests be written in the same order).  In doing so I realized that this actually caused some tests to fail as importing UITester ultimately led to trying to import things from `traitsui.testing.api` whose imports happened after UI Tester.  To avoid this, I simply had those problem cases not import from the api.  (I'm not sure if this was the correct approach, but I thought it made sense).

### DIFF
--- a/traitsui/testing/api.py
+++ b/traitsui/testing/api.py
@@ -57,6 +57,9 @@ Exceptions
 - :class:`~.TesterError`
 """
 
+# Note: Imports are ordered to match the docstring, not in the typical
+# alphabetical order
+
 # Tester
 from .tester.ui_tester import UITester
 
--- a/traitsui/testing/api.py
+++ b/traitsui/testing/api.py
@@ -57,19 +57,25 @@ Exceptions
 - :class:`~.TesterError`
 """
 
+# Tester
+from .tester.ui_tester import UITester
+
+# Interactions (for changing GUI states)
 from .tester.command import (
     MouseClick,
     KeyClick,
     KeySequence
 )
 
-from .tester.exceptions import (
-    Disabled,
-    InteractionNotSupported,
-    LocationNotSupported,
-    TesterError
+# Interactions (for getting GUI states)
+from .tester.query import (
+    DisplayedText,
+    IsChecked,
+    IsEnabled,
+    SelectedText
 )
 
+# Locations (for locating GUI elements)
 from .tester.locator import (
     Index,
     TargetById,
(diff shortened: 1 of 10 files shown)


traitsui/testing/api.py
```diff
@@ -78,13 +84,13 @@
...
-    DisplayedText,
-    IsChecked,
-    IsEnabled,
-    SelectedText
+# Exceptions
+from .tester.exceptions import (
+    Disabled,
+    InteractionNotSupported,
+    LocationNotSupported,
+    TesterError
```
kitchoi, line 95: On the one hand, I agree that having the import order matching the docstring helps detecting missing entries in the docstring, so it is useful. On the other hand, it goes against the usual convention of alphabetically ordering imports.

If you'd like to keep the new ordering here, because this is out of the ordinary of alphabetical order, it might be worth a comment to explain the grouping matches the docstring. In any case, the new comments like `# Exceptions` are good!

traitsui/testing/tester/_ui_tester_registry/_traitsui_ui.py
```diff
@@ -9,7 +9,7 @@
 #  Thanks for using Enthought open source!
 #
 
-from traitsui.testing.api import TargetById, TargetByName
+from traitsui.testing.tester.locator import TargetById, TargetByName
```
kitchoi, line 12: I think this makes more sense for the internal non-test code NOT to depend on `api` . So this change is good.

The audience of `api` is the user of the library. It makes discovery of public API easier and it encapsulates internal structure from the users. In order to do so, it brings together all public API the library has to offer, i.e. it _depends_ on the library internal code.  For the library internal code to depend on its own `api` again, cycle dependency is created and cycle imports would be inevitable.

It is fine for test code to import from `api` though (e.g. `test_button_editor`), because test is a component isolated from the software being tested, so it is allowed to depend on `api` like the user would do. In fact, it is better for tests to import from `api` as it exercises the user's context better.

I think previously this has worked based on chance 😓 . The import that has the most dependency is `ui_tester` (it is the top-level thing). It was imported late in `api` by virtue of alphabetical order. The cycle dependency is there but cycle imports were not triggered because Python import caches imported modules.

kitchoi (approved):
LGTM

traitsui/testing/tester/_ui_tester_registry/_traitsui_ui.py
aaronayres35: I removed a few more API imports on this PR that were being used by internal non-test code.  All of the editor implementations currently import from API - I think I will replace these in a separate PR.

kitchoi:
Strictly speaking this does not need to be backported to 7.1, but it would be good to keep the maintenance branch closer to the development branch so that future backports of bugfixes would be easier. We have good coverage for all these imports so this is safe enough.

PR #1274, opened 2020-09-24
Update UI Tester demo examples to import from traitsui.testing.api

aaronayres35:
This fixes the fifth bullet of #1173 

It simply changes the import statements in the current tests of examples in `examples/demo` to import from the now populated `traitsui.testing.api`

PR #1371, opened 2020-10-15
Update UI Tester demo example for ListEditor demo to import from traitsui.testing.api

aaronayres35:
I believe the PR to introduce this test was not merged before #1274 and there was probably a copy paste error.  
This PR simply uses the testing api for imports and removes an old comment.


kitchoi (approved):
Well spotted! LGTM

PR #1374, opened 2020-10-16
Add "update" click command to etstool.py

rahulporuri:
This commit adds the "update" click command to the etstool.py utility.

This command has been lifted from the etstool.py utility in the pyface github repository - with one minor change - which was to remove the `edm_option` which doesn't exist in the traitsui etstool.py utility.

xref https://github.com/enthought/pyface/blob/9f130977ee2546fa2655473cf37d8f42e0f880ef/etstool.py#L360-L373

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -255,6 +255,21 @@ def install(runtime, toolkit, environment, editable, source):
 
 
 @cli.command()
+@click.option("--runtime", default="3.6", help="Python version to use")
+@click.option("--toolkit", default="pyqt", help="Toolkit and API to use")
+@click.option("--environment", default=None, help="EDM environment to use")
+def update(runtime, toolkit, environment):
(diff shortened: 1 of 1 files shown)


etstool.py
```diff
@@ -254,6 +254,21 @@ def install(runtime, toolkit, environment, editable, source):
     click.echo('Done install')
 
 
+@cli.command()
+@click.option("--runtime", default="3.6", help="Python version to use")
+@click.option("--toolkit", default="pyqt", help="Toolkit and API to use")
+@click.option("--environment", default=None, help="EDM environment to use")
+def update(runtime, toolkit, environment):
```
kitchoi, line 261: Now this `update` is defined twice... Flake8 would have caught this but this file is not currently being flaked.

rahulporuri:
apparently i am blind. sorry for the noise. closing.

PR #1376, opened 2020-10-16
Skip load_tests in API documentation

### LINKED ISSUE #1375: Skip load_tests in the API documentation
Currently the API documentation will include `load_tests`.
load-tests-screenshot

This function is invoked by `unittest discover` and is mostly of the time, not relevant for users of traitsui.
It would be nicer if it is hidden in the public API documentation.

Similar issue: https://github.com/enthought/pyface/issues/757

kitchoi:
Closes #1375 

Similar to https://github.com/enthought/pyface/pull/762, this PR adds a skip logic in autodoc so that `load_tests` does not appear in the API documentation.

Maybe a more scalable and explicit solution is to have a decorator purely for documentation purpose, and the decorator adds a private attribute that autodoc-skip-member will use to skip the member. So one does something like this:
```
@skip_api_doc
def load_tests(...):
    ...
```

With the current PR, if one actually wants to document a function called `load_tests`, they are going to have a harder time figuring out why it is not visible on the API documentation...
On the other hand, this might well be an overkill. Just an idea.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -189,7 +189,7 @@ except ImportError as exc:
 # #html_style = 'default.css'
 # #html_theme = 'classic'
 
-# Useful aliases to avoid repeating long URLs. 
+# Useful aliases to avoid repeating long URLs.
 extlinks = {'github-demo': (
     'https://github.com/enthought/traitsui/tree/master/traitsui/examples/demo/%s',
     'github-demo')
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -189,7 +189,7 @@
 # #html_style = 'default.css'
 # #html_theme = 'classic'
 
-# Useful aliases to avoid repeating long URLs. 
+# Useful aliases to avoid repeating long URLs.
```
kitchoi, line 192: This was my IDE doing... as I was about to revert this, I figured that this is probably a good drive-by, so I kept it.

docs/source/conf.py
```diff
@@ -238,3 +238,13 @@
     'traits': ('http://docs.enthought.com/traits', None),
     'pyface': ('http://docs.enthought.com/pyface', None),
 }
+
+
+def autodoc_skip_member(app, what, name, obj, skip, options):
+    # Skip load_tests targeting unittest discover
+    is_load_tests = what == "module" and name == "load_tests"
```
kitchoi, line 245: The `what == "module"` part was not there in the pyface PR. This additional check is more strict: it excludes class method named "load_tests" or exceptions named "load_tests"... which aren't part of unittest discover protocol.

rahulporuri (approved):
LGTM. Tested this locally and it works as expected.

Issue #1364, opened 2020-10-15
FileEditor dialog inconsistently opens with PyQt5 5.15 on macOS


yoda-vid:
I have found the `FileEditor` dialog window to sporadically fail to open starting with PyQ5 5.15 on macOS. Sometimes the dialog will open on the first try, other times I will need to click it up to 20 times to open the dialog. The behavior is also inconsistent with consecutive button presses. A file dialog opened through PyFace does not have this issue.

Testing:
I ran the File Editor demo as follows and clicked on the folder button in the `File` widget:
```
python traitsui/examples/demo/Standard_Editors/File_Dialog/File_Open.py
```

**Edit (2020-10-19)**: I posted the incorrect demo, which should be:
```
python traitsui/examples/demo/Standard_Editors/FileEditor_demo.py
```

For comparison with PyFace, I ran its Python Editor demo and clicked on the `Open` button:
```
python examples/python_editor.py
```

Environment:
- Create a barren Venv environment with Python 3.6
- Pip install traitsui and PyQt5 (current version is 5.15.1)
- Test as above
- Pip uninstall PyQt5 and install PyQt5==5.14.2
- Test as above

I encountered this behavior in macOS 10.14 and 10.15 but did not have any issues on Windows 10 (there I used the Conda traitsui package). When testing on prior versions of PyQt5 (eg 5.14.2), the dialog opens consistently.

kitchoi:
Hi @yoda-vid, thank you for your report!

I am testing this on macOS 10.15.5, and this is what I did:
```
$ python3.8 -m venv my_env
$ source my_env/bin/activate
(my_env) $ pip install traitsui pygments pyqt5==5.15.1
(my_env) $ python traitsui/examples/demo/Standard_Editors/File_Dialog/File_Open.py
```

I tried to press the button many times, rerunning the script, but I could not reproduce the issue: the file dialog opens reliably.
I also tried installing `pyqt5==5.14.2`, and I could not reproduce.
Perhaps there is something I miss?

yoda-vid:
Thanks for testing, @kitchoi !

I realized that I posted the incorrect demo, which should be the `FileEditor_demo` instead. I followed your approach except ran this example:

```
python traitsui/examples/demo/Standard_Editors/FileEditor_demo.py
```

[...]

PR #1383, opened 2020-10-19
Fix Qt IconButton with a clickable area too small

kitchoi:
Closes https://github.com/enthought/traitsui/issues/1364

On OSX, the clickable area of the icon button is too small: you have to click dead center with a few pixel accuracy to actually click the button. This explains why sometimes the dialog can be launched and sometimes it cannot.

This issue affects the simple FileEditor, RangeEditor (the buttons for incrementing and decrementing the value), and the ListEditor (the button for opening a menu for adding or removing an item).

This PR fixes the issue by reimplementing the sizeHint of the `IconButton` (inheriting QPushButton) to return a larger size and remove the `setMaximumSize` line which set the button size to be too small.  The sizeHint implemented here follows a logic similar to that of `QToolButton` (see source [here](https://code.woboq.org/qt5/qtbase/src/widgets/widgets/qtoolbutton.cpp.html)). 

On master, see the little grey area when I click the button:
Oct-19-2020 12-19-00

On this branch, the clickable area is bigger (as shown by the larger grey area when I press the button):
Oct-19-2020 12-19-27

### DIFF
--- a/traitsui/qt4/helper.py
+++ b/traitsui/qt4/helper.py
@@ -179,6 +179,13 @@ class IconButton(QtGui.QPushButton):
         self.clicked.connect(slot)
 
     def sizeHint(self):
+        """ Reimplement sizeHint to return a recommended button size based on
+        the size of the icon.
+
+        Returns
+        -------
+        size : QtCore.QSize
+        """
         # We want the button to have a size similar to the icon.
         # Using the size computed for a tool button gives a desirable size.
         option = QtGui.QStyleOptionButton()
--- a/traitsui/qt4/helper.py
+++ b/traitsui/qt4/helper.py
@@ -173,12 +173,20 @@ class IconButton(QtGui.QPushButton):
 
         # Configure the button.
         self.setIcon(ico)
-        self.setMaximumSize(ico_sz, ico_sz)
         self.setFlat(True)
         self.setFocusPolicy(QtCore.Qt.NoFocus)
 
         self.clicked.connect(slot)
 
+    def sizeHint(self):
+        # We want the button to have a size similar to the icon.
+        # Using the size computed for a tool button gives a desirable size.
+        option = QtGui.QStyleOptionButton()
+        self.initStyleOption(option)
+        size = self.style().sizeFromContents(
+            QtGui.QStyle.CT_ToolButton, option, option.iconSize
+        )
+        return size
 
 # ------------------------------------------------------------------------
 # Text Rendering helpers
--- a/traitsui/qt4/helper.py
+++ b/traitsui/qt4/helper.py
@@ -186,6 +186,8 @@ class IconButton(QtGui.QPushButton):
         -------
         size : QtCore.QSize
         """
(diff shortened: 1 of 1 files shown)


traitsui/qt4/helper.py
```diff
@@ -173,12 +173,20 @@ def __init__(self, icon, slot):
...
 
+    def sizeHint(self):
+        # We want the button to have a size similar to the icon.
+        # Using the size computed for a tool button gives a desirable size.
+        option = QtGui.QStyleOptionButton()
+        self.initStyleOption(option)
+        size = self.style().sizeFromContents(
+            QtGui.QStyle.CT_ToolButton, option, option.iconSize
+        )
+        return size
```
kitchoi, line 189: The added `sizeHint` method is covered by test: if I try to return anything else that is not an instance of QSize, the test suite would fail. However any assertion on the returned size is going to be tricky: it varies from system to system.

kitchoi:
For the record, the issue does not seem to affect Windows or Linux. The button is able to register the click even if I click close to the edge of the icon.

On Windows, when I press and hold the button, this is what I saw on master (using FileEditor_demo.py):
master_button
And this is what I saw on this branch:
fixed_button

traitsui/qt4/helper.py
```diff
@@ -173,12 +173,20 @@ def __init__(self, icon, slot):
...
         self.setIcon(ico)
-        self.setMaximumSize(ico_sz, ico_sz)
         self.setFlat(True)
         self.setFocusPolicy(QtCore.Qt.NoFocus)
 
         self.clicked.connect(slot)
 
+    def sizeHint(self):
+        # We want the button to have a size similar to the icon.
+        # Using the size computed for a tool button gives a desirable size.
```
aaronayres35, line 183: Could we add a docstring for this method?  The block comment covers the majority of the important information but having a docstring would be good for consistency

aaronayres35 (approved):
LGTM!  Just one comment about adding a docstring.  Also I was able to reproduce the results you showed on OSX.

traitsui/qt4/helper.py
```diff
@@ -173,12 +173,27 @@ def __init__(self, icon, slot):
...
 
+    def sizeHint(self):
+        """ Reimplement sizeHint to return a recommended button size based on
+        the size of the icon.
+
+        Returns
+        -------
+        size : QtCore.QSize
+        """
+        # We want the button to have a size similar to the icon.
```
kitchoi, line 189: I should call `ensurePolished()` here for good measure: https://doc.qt.io/qt-5/qwidget.html#ensurePolished
This was also called in the original implementation. https://code.woboq.org/qt5/qtbase/src/widgets/widgets/qpushbutton.cpp.html#_ZNK11QPushButton8sizeHintEv

traitsui/qt4/helper.py
kitchoi: Yep. Done!

Issue #1024, opened 2020-07-17
Remove deprecated traitsui.image submodule


rahulporuri:
The `traitsui.image` submodule got moved to `pyface.image` in the PR https://github.com/enthought/traitsui/pull/304 and https://github.com/enthought/pyface/pull/221 . The changes have been public since versions 6.0.0 of traitsui and pyface. I think it's time we can remove the deprecated `traitsui.image` submodule.

https://github.com/enthought/traitsui/blob/6fccc3966e89a8082aace9623b84a80a74c46f3e/traitsui/image/image.py#L1-L10

Note that setup.py will also need to be updated if `traitsui.image` is removed - 
https://github.com/enthought/traitsui/blob/6fccc3966e89a8082aace9623b84a80a74c46f3e/setup.py#L182

rahulporuri:
@corranwebster can we get rid of this now or should it stay in master for longer (until after 7.1.0)?

kitchoi:
Hmmm I can find this import in Mayavi: https://github.com/enthought/mayavi/blob/9c128635943e28513776dec68492ffd0268c28d1/mayavi/tools/data_wizards/data_source_wizard.py#L12

kitchoi:
FTR: It is difficult to spot deprecation warning emitted at import time as they are not very visible. By the time one runs the test suite and have unittest runner set the warning filter, the deprecation warning is already emitted and no one gets to see them. To counter that, I have `export PYTHONWARNINGS='default'` in my `~/.bash_profile`. 😂  Here in TraitsUI's `etstool.py` the filter is also set: https://github.com/enthought/traitsui/blob/f9dfbdce42c96a5959156e8cbe30317a5bc61fd6/etstool.py#L312

kitchoi:
Opened issue on Mayavi: https://github.com/enthought/mayavi/issues/971

kitchoi:
Given `traitsui.image` is merely providing aliases, it costs little to keep it (unless `pyface.image` also needs to be moved, but I can't see how that would be the case). An alternative would be to say in the warning message more definitely that imports from `traitsui.image` are no longer maintained and developers importing from this module are doing so at their own risk.

PR #1384, opened 2020-10-19
Update deprecation warnings for removing traitsui.image

kitchoi:
Related to #1024
This PR updates the deprecation warning message with more up-to-date information.

### DIFF
--- a/traitsui/image/image.py
+++ b/traitsui/image/image.py
@@ -5,6 +5,6 @@ from warnings import warn
 warn(
     DeprecationWarning(
         "traitsui.image has been moved to pyface.image and will be removed in "
-        "TraitsUI version 6.0"
+        "TraitsUI version 7.2"
     )
 )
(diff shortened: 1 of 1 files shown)


traitsui/image/image.py
```diff
@@ -5,6 +5,6 @@
 warn(
     DeprecationWarning(
         "traitsui.image has been moved to pyface.image and will be removed in "
-        "TraitsUI version 6.0"
+        "TraitsUI version 7.2"
```
aaronayres35, line 8: Should we still mention it was originally intended to be removed in 6.0 and now will be removed in 7.2?  This implies it has been on its way out for quite some time.  It may not matter, but I thought I remembered that from the stand up.

aaronayres35 (approved):
LGTM! Just one comment

traitsui/image/image.py
kitchoi: Sure!

rahulporuri (approved):
Thanks 🙏 LGTM

PR #1386, opened 2020-10-20
Doc: Restore API documentation URLs

kitchoi:
In #1368, the files generated by `sphinx-apidoc` are stored in a separate folder `api/autogenerated` so that they are kept separate from the manually curated `api/index.rst`, allowing the folder to be ignored by git and be reset by `etstool.py` easily.

However, this causes the URLs for the API references to change. If these links have been referenced externally, they will become dead links.

This PR restores the URLs. The manually curated `api.index.rst` is still kept, at a cost:
 The `.gitignore` is slightly less trivial. The `etstool.py docs` command also won't reset the folder. Developers should do `git clean` instead in case their `docs/source/api` folder is contaminated.


rahulporuri (approved):
LGTM. Haven't tested this locally though.

PR #1391, opened 2020-10-20
Rename version folders for built documentation

kitchoi:
[Targeting gh-pages]

This PR renames the `release-X.Y.Z` folders to just `X.Y`.

This will break any external full URL referencing _historical_ releases. However, links to historical releases have not been publicized nor referenced in the documentation, and hence it is unlikely they have been used. (If they are really really need, it is easy to resurrect them again).

Having the folder named `MAJOR.MINOR.MICRO` forces developers building the documentation to create a new folder for bugfix releases, e.g. the build for release 7.0.1 cannot override the build for release 7.0.0 and has to be a separate folder.  It is unconventional to maintain separate documentation builds down to the micro versions, as there should be no API changes for bugfix releases. For example, Python documentations are hosted for MAJOR.MINOR versions only.


rahulporuri (approved):
LGTM. Served/checked the docs locally and things work as expected (i could access the 5.1, 6.0, 6.1, 7.0 and 7.1 folders)

PR #1392, opened 2020-10-26
Dont install pyface from git source

### LINKED ISSUE #1382: Dont install pyface from source after 7.1.0 release
Once pyface 7.1.0 is released (currently underway), it would be nice to stop installing `pyface` from source - and instead use the released version of pyface.

See https://github.com/enthought/traitsui/blob/master/ci-src-requirements.txt and https://github.com/enthought/traitsui/blob/cfa7781413486a33754527364d41b58444e577a3/etstool.py#L112-L116

rahulporuri:
fixes #1382 

This PR updates CI/etstool to install pyface using edm instead of installing it from git source.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -117,6 +117,7 @@ dependencies = {
 
 # NOTE : pyface is always installed from source
 source_dependencies = {
+    "pyface": "git+http://github.com/enthought/pyface.git#egg=pyface",
     "traits": "git+http://github.com/enthought/traits.git#egg=traits",
 }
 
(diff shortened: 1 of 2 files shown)


kitchoi:
Same as https://github.com/enthought/pyface/pull/787
I shall let @corranwebster review this.

etstool.py
```diff
@@ -111,8 +111,8 @@
 
 # Required runtime dependencies. Should match install_requires in setup.py
 dependencies = {
+    "pyface",
     "traits" + TRAITS_VERSION_REQUIRES,
-    # pyface is installed from source via ci-src-requirements.txt
 }
 
 # NOTE : pyface is always installed from source
```
kitchoi, line 118: The source_dependencies still needs to be updated so that cron job is run against pyface master?

rahulporuri: yep. nice catch. corrected.

kitchoi (approved):
Offline discussion: @corranwebster confirmed to go ahead with installing released Pyface for TraitsUI CI, but he also said we may need to switch back to installing Pyface master for some tasks.

PR #1393, opened 2020-10-26
Remove unnecessary use of f-string.

rahulporuri:
In this specific case, there was nothing that needed to be formatted in the string so we can just use a normal string.

This was reported as a flake8 failure in the CI job - https://travis-ci.org/github/enthought/traitsui/jobs/738921895, PR #1392 .


kitchoi (approved):
LGTM

PR #1395, opened 2020-11-05
Run pip install with no deps

rahulporuri:
When `etstool.py install` is run in CI, because we are not passing `--no-deps`, `pyface` gets installed once via edm and a second time via pip/setup.

This should not usually be a problem as ideally pip would try to install the same version that edm installed - but this cron job begs to differ - https://travis-ci.com/github/enthought/traitsui/jobs/428569793. For reasons that are not entirely clear, edm installed pyfave version 6.1.2 and then running `pip install .` installed version 7.1.0. After that, removing the edm package worked as expected but trying to install pyface from source crashed because pyface was already installed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
     - env: RUNTIME=3.6 TOOLKITS="null"
     - env: RUNTIME=3.6 TOOLKITS="pyqt pyqt5 pyside2"
     - env: RUNTIME=3.6 TOOLKITS="wx"
-    - env: RUNTIME=3.6 TOOLKITS="pyqt5" TRAITS_REQUIRES="^=6.0"
+    - env: RUNTIME=3.6 TOOLKITS="pyqt5"
   allow_failures:
     - env: RUNTIME=3.6 TOOLKITS="wx"
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ matrix:
     - env: RUNTIME=3.6 TOOLKITS="null"
     - env: RUNTIME=3.6 TOOLKITS="pyqt pyqt5 pyside2"
     - env: RUNTIME=3.6 TOOLKITS="wx"
-    - env: RUNTIME=3.6 TOOLKITS="pyqt5"
+    - env: RUNTIME=3.6 TOOLKITS="pyqt5" TRAITS_REQUIRES="^=6.0"
   allow_failures:
     - env: RUNTIME=3.6 TOOLKITS="wx"
   fast_finish: true
--- a/etstool.py
+++ b/etstool.py
@@ -221,10 +221,13 @@ def install(runtime, toolkit, environment, editable, source):
         | ci_dependencies
     )
 
-    install_traitsui = "edm run -e {environment} -- pip install "
+    install_traitsui = (
+        "edm run -e {environment} -- "
+        "pip install --no-deps --force-reinstall "
+    )
     if editable:
         install_traitsui += "--editable "
-    install_traitsui += ". --no-deps"
+    install_traitsui += "."
 
     # edm commands to setup the development environment
     if sys.platform == 'linux':
--- a/etstool.py
+++ b/etstool.py
@@ -224,7 +224,7 @@ def install(runtime, toolkit, environment, editable, source):
     install_traitsui = "edm run -e {environment} -- pip install "
     if editable:
         install_traitsui += "--editable "
-    install_traitsui += "."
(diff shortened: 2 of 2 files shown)


etstool.py
```diff
@@ -224,7 +224,7 @@ def install(runtime, toolkit, environment, editable, source):
     install_traitsui = "edm run -e {environment} -- pip install "
     if editable:
         install_traitsui += "--editable "
-    install_traitsui += "."
+    install_traitsui += ". --no-deps"
```
kitchoi, line 227: Nit: It might read better if this `--no-deps` is added after `pip install ` on line 224.

I am tempted to add `--force-reinstall` here just to be safe so we don't recreate what happens in enable where the CI is run against released traitsui code rather than the local source.

.travis.yml
```diff
@@ -30,7 +30,7 @@ matrix:
     - env: RUNTIME=3.6 TOOLKITS="null"
     - env: RUNTIME=3.6 TOOLKITS="pyqt pyqt5 pyside2"
     - env: RUNTIME=3.6 TOOLKITS="wx"
-    - env: RUNTIME=3.6 TOOLKITS="pyqt5" TRAITS_REQUIRES="^=6.0"
```
kitchoi, line 33: Could this be a separate PR please? It would be good if we see more obviously at what point we stopped testing against Traits 6.0. It needs to be prominent. Also if we are shuffling things around between pyface, envisage and apptools, whereas the next release of envisage and apptools should still support Traits 6.0, we may not be able to drop Traits 6.0 support on Pyface this soon.

kitchoi: Oops, this is in traitsui. Anyway, same comment applies here too as envisage and apptools depend on traitsui.

rahulporuri: reverted.

etstool.py
rahulporuri: moved `--no-deps` and added `--force-reinstall`

kitchoi (approved):
LGTM

rahulporuri:
> Oops, looks like CI is still not happy.

I'm not sure what the best solution here is. Because we pin traits to v 6.0 in the pyqt5+ linux travis ci job, we get traitsui and pyface < v 7.0 -> and when we force reinstall traitsui from source without depenencies, triatsui is reinstalled but that doesn't bring in the necessary new version of pyface. We pinned triatsui >= 7.0 to need pyface >= 7.0.

One way to solve this is to specifically allow a newer version of pyface when installing packages - which is acceptable because pyface >= 7.0 still supports traits v 6.0.

etstool.py
```diff
@@ -232,8 +239,20 @@ def install(runtime, toolkit, environment, editable, source):
...
+    # of dependencies e.g. traitsui/pyface < 7.0, which is a problem for
+    # traitsui master because it requires pyface >= 7.0.
+    # So, for now, if an older version of traits is required, we explicitly
+    # allow newer versions of pyface.
+    # This constraint modifier can be removed when we stop supporting traits
+    # 6.0.
+    if TRAITS_VERSION_REQUIRES:
+        constraint_modifiers = CONSTRAINT_MODIFIERS
+    else:
+        constraint_modifiers = ""
```
kitchoi, line 252: Nit on readability: Could we move this if-else logic together with where `CONSTRAINT_MODIFIERS` is defined? Or just define the constraint modifier here. It is confusing to see `CONSTRAINT_MODIFIERS` being defined as a constant at the top while it is actually a value depending on `TRAITS_VERSION_REQUIRES`.  The `constraint_modifiers` also have different types in these two branches. String happens to be an iterable in Python, but logically, it is a scalar value.

Something like this?
```
CONSTRAINT_MODIFIERS = {"--allow-newer pyface"} if TRAITS_VERSION_REQUIRES else set()
```

rahulporuri:
sigh. that didn't work as expected. Allowing newer versions of pyface led to edm choosing pyface 7.0.0 and not pyface 7.1.0, which is what triatsui master relies on.

kitchoi:
We now have a scheduled workflow on GitHub Actions to test against Traits and Pyface master. The job for Traits 6.0 simply won't interfere. The last run triggered by `git push` is here: https://github.com/enthought/traitsui/runs/1463016248?check_suite_focus=true
(Logs do not last forever, here is the saved log: [cron.txt](https://github.com/enthought/traitsui/files/5607454/cron.txt)

This PR shows how awkward it is to maintain independent workflows in one configuration.
Shall we close this PR and continue with migrating the workflows over to GitHub Actions?

rahulporuri:
closing this PR as we are will continue to migrate workflows to github actions ci.

PR #1397, opened 2020-11-13
Support -h for getting help

rahulporuri:
Currently, `python -m etstool --help` works for getting help but `python -m etstool -h` does not. This breaks user expectations when working with a command line application. This PR makes `-h` work as expected.

(For @achabotl )


aaronayres35 (approved):
LGTM, I tested manually and it worked as expected

PR #1406, opened 2020-11-19
Enable the scrollable trait of a Group in the Qt backend

### LINKED ISSUE #1403: Scrollable groups not implemented in Qt backend
They should probably be implemented somewhere in this section: https://github.com/enthought/traitsui/blob/770e94346e06dc70909f125ba7237edce90872da/traitsui/qt4/ui_panel.py#L582-L608

The fix would likely involve creating a scrollable sub-panel, and then adding to that.

The option is implemented in the WxPython backend.

corranwebster:
Fixes #1403 

This only attempts to add scrolling to vertical, horizontal and grid groups, as it doesn't make sense for tabbed, split or vfold groups.

Currently no tests: I could hack one together, but it would be good to know if this can be integrated into the new testing framework easily.

### DIFF
--- a/traitsui/qt4/tests/test_ui_panel.py
+++ b/traitsui/qt4/tests/test_ui_panel.py
@@ -172,11 +172,14 @@ class TestPanelLayout(unittest.TestCase):
         example = ScrollableGroupExample()
 
         ui = example.edit_traits(view=scrollable_group_view)
-        mainwindow = ui.control.layout().itemAt(0).widget()
-        scroll_area = mainwindow.centralWidget()
-        self.assertIsInstance(scroll_area, QtGui.QScrollArea)
-        content = scroll_area.widget()
-        self.assertIsInstance(content, QtGui.QWidget)
+        try:
+            mainwindow = ui.control.layout().itemAt(0).widget()
+            scroll_area = mainwindow.centralWidget()
+            self.assertIsInstance(scroll_area, QtGui.QScrollArea)
+            content = scroll_area.widget()
+            self.assertIsInstance(content, QtGui.QWidget)
+        finally:
+            ui.dispose()
 
     def test_scrollable_group_box(self):
         from pyface.qt import QtGui
@@ -184,11 +187,14 @@ class TestPanelLayout(unittest.TestCase):
         example = ScrollableGroupExample()
 
         ui = example.edit_traits(view=scrollable_group_box_view)
-        mainwindow = ui.control.layout().itemAt(0).widget()
-        scroll_area = mainwindow.centralWidget()
-        self.assertIsInstance(scroll_area, QtGui.QScrollArea)
-        group_box = scroll_area.widget()
-        self.assertIsInstance(group_box, QtGui.QGroupBox)
+        try:
+            mainwindow = ui.control.layout().itemAt(0).widget()
+            scroll_area = mainwindow.centralWidget()
+            self.assertIsInstance(scroll_area, QtGui.QScrollArea)
+            group_box = scroll_area.widget()
+            self.assertIsInstance(group_box, QtGui.QGroupBox)
+        finally:
+            ui.dispose()
 
     def test_scrollable_labelled_group(self):
         from pyface.qt import QtGui
@@ -196,18 +202,24 @@ class TestPanelLayout(unittest.TestCase):
         example = ScrollableGroupExample()
 
         ui = example.edit_traits(view=scrollable_labelled_group_view)
-        mainwindow = ui.control.layout().itemAt(0).widget()
-        scroll_area = mainwindow.centralWidget()
-        self.assertIsInstance(scroll_area, QtGui.QScrollArea)
-        group_box = scroll_area.widget()
-        self.assertIsInstance(group_box, QtGui.QWidget)
+        try:
+            mainwindow = ui.control.layout().itemAt(0).widget()
+            scroll_area = mainwindow.centralWidget()
+            self.assertIsInstance(scroll_area, QtGui.QScrollArea)
+            group_box = scroll_area.widget()
+            self.assertIsInstance(group_box, QtGui.QWidget)
+        finally:
+            ui.dispose()
 
     def test_non_scrollable_group_typical(self):
         from pyface.qt import QtGui
 
         example = ScrollableGroupExample()
 
-        ui = example.edit_traits(view=non_scrollable_group_view)
-        mainwindow = ui.control.layout().itemAt(0).widget()
-        content = mainwindow.centralWidget()
-        self.assertIsInstance(content, QtGui.QWidget)
+        try:
+            ui = example.edit_traits(view=non_scrollable_group_view)
+            mainwindow = ui.control.layout().itemAt(0).widget()
+            content = mainwindow.centralWidget()
+            self.assertIsInstance(content, QtGui.QWidget)
+        finally:
+            ui.dispose()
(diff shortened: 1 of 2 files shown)


kitchoi:
> Currently no tests: I could hack one together, but it would be good to know if this can be integrated into the new testing framework easily.

I am guessing what you want to be testing is the final Qt widget structure, so the GUI event loop does not even need to start and it is probably not something of interest to downstream projects' tests. Using the new testing library would probably be an overkill, but it can work.

corranwebster:
> I am guessing what you want to be testing is the final Qt widget structure

Yes; ideally Wx as well since it should work there also, but the structure will be different of course.  The event loop probably isn't needed, as this is just about layout.

Having helpers for testing layout would potentially be useful for other projects, but in most cases projects should be able to inspect the structure of the `View` and trust that TraitsUI does the right thing from that.

It sounds like I am leaning towards a really targeted test of `_GroupPanel` inside `traitsui.qt4.tests.*`, which is less than ideal because it is testing private implementation details, but it is the cleanest way to test the addition does what it is supposed to.

corranwebster:
Looks like the tests need to be cleverer about finding the widgets.

traitsui/qt4/tests/test_ui_panel.py
```diff
@@ -110,3 +161,53 @@ def test_subpanel_no_toolbar_nor_button_in_widget(self):
...
+@requires_toolkit([ToolkitName.qt])
+class TestPanelLayout(unittest.TestCase):
+
+    def test_scrollable_group_typical(self):
+        from pyface.qt import QtGui
+
+        example = ScrollableGroupExample()
+
+        ui = example.edit_traits(view=scrollable_group_view)
+        mainwindow = ui.control.layout().itemAt(0).widget()
```
kitchoi, line 175: Minor: If instead we do `mainwindow, = ui.control.findChildren(QtGui.QMainWindow)`, this may expose the test to less implementation details while still asserting the expected behaviour?

traitsui/qt4/ui_panel.py
```diff
@@ -596,6 +596,21 @@ def __init__(self, group, ui, suppress_label=False):
                 # Create an editor.
                 self._setup_editor(group, GroupEditor(control=outer))
 
+            if group.scrollable:
+                # ensure we have a widget rather than a layout for the scroll area
```
kitchoi, line 600: Nit: Local flake8 🤖  Line too long

traitsui/qt4/tests/test_ui_panel.py
```diff
@@ -110,3 +161,53 @@ def test_subpanel_no_toolbar_nor_button_in_widget(self):
...
+        from pyface.qt import QtGui
+
+        example = ScrollableGroupExample()
+
+        ui = example.edit_traits(view=scrollable_labelled_group_view)
+        mainwindow = ui.control.layout().itemAt(0).widget()
+        scroll_area = mainwindow.centralWidget()
+        self.assertIsInstance(scroll_area, QtGui.QScrollArea)
+        group_box = scroll_area.widget()
+        self.assertIsInstance(group_box, QtGui.QWidget)
```
kitchoi, line 203: The returned widget is not actually a `QGroupBox` so the variable name is a bit misleading (as I was thinking I could do `assertIsInstance(group_box, QtGui.QGroupBox)` instead, but I can't.

Not sure if this is an acceptable amount of implementation details being used to check whether the label is within the scrollable area:
```
        label = scroll_area.findChild(QtGui.QLabel)
        self.assertIn("Scrollable View", label.text())
```

(I remember segfault from ModalDialogTester checking for `text()`. In the context of `ModelDialogTester`, the GUI event loop is running and such query 
 is made from when the event is being processed... Here the GUI event loop is not running and making these searches from the main thread seems safer than the setup we have in ModelDialogTester.)

kitchoi (approved):
LGTM. Minor suggestions for the tests.

traitsui/qt4/tests/test_ui_panel.py
```diff
@@ -110,3 +161,53 @@ def test_subpanel_no_toolbar_nor_button_in_widget(self):
...
+
+@requires_toolkit([ToolkitName.qt])
+class TestPanelLayout(unittest.TestCase):
+
+    def test_scrollable_group_typical(self):
+        from pyface.qt import QtGui
+
+        example = ScrollableGroupExample()
+
+        ui = example.edit_traits(view=scrollable_group_view)
```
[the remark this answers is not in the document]
corranwebster: Yes, I had added this as you were reviewing (via `try...except...` rather than `addCleanup`, but equivalent.

traitsui/qt4/tests/test_ui_panel.py
```diff
@@ -110,3 +161,53 @@ def test_subpanel_no_toolbar_nor_button_in_widget(self):
...
+        mainwindow = ui.control.layout().itemAt(0).widget()
+        scroll_area = mainwindow.centralWidget()
+        self.assertIsInstance(scroll_area, QtGui.QScrollArea)
+        content = scroll_area.widget()
+        self.assertIsInstance(content, QtGui.QWidget)
```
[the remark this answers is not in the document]
corranwebster: No, it's actually supposed to be a `QWidget` precisely, so the test is wrong - I guess `assertEquals(type(content), QtGui.QWidget)` is reasonable.

traitsui/qt4/tests/test_ui_panel.py
corranwebster: While it's a bit verbose, I'm OK with this - it's saying find the first (and it turns out, only) item in the control's layout, and get its widget, which is a reasonably straightforward description of what we're looking for.

The most likely change to affect code would be to make the top-level control a `QMainWindow` subclass instead of a sticky `Dialog` in which case this test would be broken anyway.

traitsui/qt4/ui_panel.py
corranwebster: Need to look into my VS Code setup - it didn't flag this 😞

traitsui/qt4/tests/test_ui_panel.py
corranwebster: I don't think I want to test for where the label occurs - it can't be on the outside because then we wouldn't get a `QScrollArea` where we would expect it.  Testing the inserted label is really a test for the group `label` trait, not the `scrollable` trait, and so getting beyond the scope of this PR.

traitsui/qt4/tests/test_ui_panel.py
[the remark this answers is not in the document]
kitchoi: For my understanding (and possibly other developers who may have to look into this test assertion again), why is it supposed to be a `QWidget` **precisely**?

kitchoi (approved):
Still LGTM.

(Not sure I understand the equals-widget part in the test as in what it means. It would well just be me, but it might be good to get some hints for future developers working with these tests.)

corranwebster:
The code we're testing is creating a bare `QWidget` (which is not an ABC) to contain stuff at the points where we are testing for it.  So the test is just saying we expect the class of the object to be precisely `QWidget` and not any subclass of it.

kitchoi:
> The code we're testing is creating a bare QWidget (which is not an ABC) to contain stuff at the points where we are testing for it. So the test is just saying we expect the class of the object to be precisely QWidget and not any subclass of it.

I see, it indicates there is a container of sub-widgets. Thank you!

PR #1407, opened 2020-11-19
Install pyface source dependencies

rahulporuri:
This PR explicitly installs the newer pyface dependencies when pyface is installed from source. This should fix the cron jobs.
This is necessary until a new release of pyface is available because the two new dependencies won't be installed in the environment otherwise.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -115,7 +115,9 @@ dependencies = {
     "traits" + TRAITS_VERSION_REQUIRES,
 }
 
-# NOTE : pyface is always installed from source
+# The order matters here, traits needs to be installed after pyface because
+# installing pyface from source will bring in traits as a dependency from
+# pypi.
 source_dependencies = {
     "pyface": "git+http://github.com/enthought/pyface.git#egg=pyface",
     "traits": "git+http://github.com/enthought/traits.git#egg=traits",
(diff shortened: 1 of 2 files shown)


etstool.py
```diff
@@ -264,20 +266,55 @@ def install(runtime, toolkit, environment, editable, source):
...
+
+    # Create the empty devenv
+    commands = ["edm environments create {environment} --force --version={runtime}"]
+    execute(commands, parameters)
+
+    install_additional_dependencies = [
+        "edm run -e {environment} -- pip install {additional_dependencies}",
+    ]
+    parameters["additional_dependencies"] = " ".join(ci_dependencies | {"sphinx", "enthought-sphinx-theme"})
+    execute(install_additional_dependencies, parameters)
```
kitchoi, line 288: I am getting this error:
```
Command '['edm', 'run', '-e', 'traitsui-test-3.6-null', '--', 'pip', 'install', 'coverage enthought-sphinx-theme flake8 sphinx']' returned non-zero exit status 1.
```

Apparently the parameters are applied _after_ the command is split by space.
I have always found this `execute` unintuitive for _me_ to understand...

etstool.py
```diff
@@ -264,20 +266,55 @@ def install(runtime, toolkit, environment, editable, source):
...
+    parameters = get_parameters(runtime, toolkit, environment)
+
+    # Create the empty devenv
+    commands = ["edm environments create {environment} --force --version={runtime}"]
+    execute(commands, parameters)
+
+    install_additional_dependencies = [
+        "edm run -e {environment} -- pip install {additional_dependencies}",
+    ]
+    parameters["additional_dependencies"] = " ".join(ci_dependencies | {"sphinx", "enthought-sphinx-theme"})
```
kitchoi, line 287: I guess it is meant to be used with string concatenation after all.
```suggestion
    additional_dependencies = " ".join(ci_dependencies | doc_dependencies)
    install_additional_dependencies = [
        "edm run -e {environment} -- pip install " + additional_dependencies,
    ]
```

etstool.py
```diff
@@ -264,20 +266,55 @@ def install(runtime, toolkit, environment, editable, source):
...
+    execute(install_additional_dependencies, parameters)
+
+    # Install the traitsui package and its dependencies
+    extras = {"editors", "test"}
+    if toolkit not in ["null", "wx"]:
+        extras.add(toolkit)
+    parameters["extras_flags"] = "[{}]".format(",".join(extras))
+    install_traitsui = [
+        "edm run -e {environment} -- pip install .{extras_flags}"
+    ]
```
kitchoi, line 297: This block works because `extras_flags` does not contain space... but for consistency with the suggested changes above, then this should be changed to:
```suggestion
    extras_flags = "[{}]".format(",".join(extras))
    install_traitsui = [
        "edm run -e {environment} -- pip install ." + extras_flags
    ]
```

etstool.py
```diff
@@ -264,20 +266,55 @@ def install(runtime, toolkit, environment, editable, source):
...
+    execute(commands, parameters)
+
+    install_additional_dependencies = [
+        "edm run -e {environment} -- pip install {additional_dependencies}",
+    ]
+    parameters["additional_dependencies"] = " ".join(ci_dependencies | {"sphinx", "enthought-sphinx-theme"})
+    execute(install_additional_dependencies, parameters)
+
+    # Install the traitsui package and its dependencies
+    extras = {"editors", "test"}
```
kitchoi, line 291: Since we are not going to test the examples in the cron job (hence skipping the examples dependencies here), we need to be able to skip the following in the `test` command (see line 361 in this file, which I could not comment directly):
```
"edm run -e {environment} -- python -W default -m unittest discover -v {integrationtests}"
```
Which means... we might need yet another command / line in `.travis.yml` for the test command to be different from per-PR build.  Maybe it is good to split out the command for running the integration tests anyway.
GitHub actions should make these setups cleaner.

etstool.py
```diff
@@ -264,20 +266,55 @@ def install(runtime, toolkit, environment, editable, source):
...
+    if toolkit not in ["null", "wx"]:
```
kitchoi, line 292: Apparently `pip install pyqt` could not resolve to a valid version, perhaps Python 3.6 is too new for it, so we might still need to use EDM for that particular build matrix...

rahulporuri:
closing now that we have cron jobs running on github actions ci.

PR #1413, opened 2020-11-26
Update the image license files

rahulporuri:
This PR makes major changes to the image license files.

- explicitly "Enthought"-licensed content has been removed from all license files in the repository because the license files state that

> "Unless stated in this file, icons are the work of Enthought, and are released under a 3 clause BSD license."

- remove license files which are empty
- consolidate individual image license files into the top-level `image_LICENSE.txt` file
- add a missing `image_LICENSE_OOo.txt` license file because we include images which are licensed OOo in the traitsui distribution.


image_LICENSE.txt
```diff
@@ -22,46 +22,22 @@ docs/source/tutorials/images:
               mpl_figure_editor.png | Enthought
                   traits_thread.png | Enthought
 
+integrationtests/ui/images:
```
kitchoi, line 25: `integrationtests` is not distributed, right?

kitchoi: Hmmm, it is, which is surprising to me. That's orthogonal then.

kitchoi (dismissed):
LGTM

image_LICENSE.txt
rahulporuri: yeah but i'm not sure if we should have attribution for *all* images in the github repository or attribution only for the images which are distributedvia traitsui.

rahulporuri: either way, we can actually remove them because at the top of the license file, we see the following

> "Unless stated in this file, icons are the work of Enthought, and are released under a 3 clause BSD license."

I went ahead and remove the section. @kitchoi do you think I should add it back?

kitchoi:
> either way, we can actually remove them because at the top of the license file, we see the following
>> "Unless stated in this file, icons are the work of Enthought, and are released under a 3 clause BSD license."
> I went ahead and remove the section. @kitchoi do you think I should add it back?

Good point. Perhaps _all_ items attributed to Enthought can be omitted for consistencies? Then we would only need to worry about ones that are externally sourced.

kitchoi (approved):
LGTM!

rahulporuri:
@kitchoi sorry for the last minute change but i discovered one more redundant image license file when i did one more round of self review.

kitchoi:
Nice. Still LGTM!

PR #1414, opened 2020-11-26
Remove codecov config file

rahulporuri:
This PR removes the codecov config file because we dont use codecov anymore.


kitchoi (approved):
LGTM

Issue #1416, opened 2020-11-27
Move CI to GitHub Actions


kitchoi:
This issue is for migrating our CI workflow to GitHub Actions such that Travis/Appveyor are removed from our CI workflows.

What we currently have on Travis + Appveyor:

TraitsUI Package tests+ Demo example tests on
- PyQt4 + Ubuntu + Python 3.6
- PyQt5 + Ubuntu + Python 3.6
- PySide2 + Ubuntu + Python 3.6
- Null + Ubuntu + Python 3.6
- wx + Ubuntu + Python 3.6 (allowed to fail)
- PyQt4 + MacOS + Python 3.6
- PyQt5 + MacOS + Python 3.6
- PySide2 + MacOS + Python 3.6
- wx + MacOS + Python 3.6 (allowed to fail)
- PyQt4 + Windows + Python 3.6
- PyQt5 + Windows + Python 3.6
- PySide2 + Windows + Python 3.6
- wx + Windows + Python 3.6 (allowed to fail)
- PyQt5 + Traits 6.0 + Ubuntu + Python 3.6

Styling checks on
- Null + Ubuntu + Python 3.6

Cron jobs run all the above with Traits and Pyface sources from their development branch (which is awkward for the job targeting Traits 6.0).

~All the above~ The Ubuntu jobs apply to packages tests for the ets-demo distribution (however I don't think the cron is actually installing traits/pyface/traitsui from source)

With GitHub Actions, we have more flexibility in configuring workflows, so we may not need to do an exact one-to-one migration.

Note also that GitHub Actions currently does not support allow-failure settings (more specifically, retaining the red tick for the job without causing the entire workflow to fail), so the ones that are allowed to fail should either be made to pass and or be removed from CI.

PR #1417, opened 2020-11-27
Reduce CI matrix on Appveyor

kitchoi:
Part of #1416

This PR removes part of the test matrix on Appveyor, which has been the bottleneck of per-PR CI build.

We do lose one test job from this change. The (allow-to-fail) job with wxPython and OSX is now gone.
The other removed jobs are now covered on GitHub Actions.

The setup on OSX with wxPython has not been successful as I ran into this issue: https://stackoverflow.com/questions/48531006/wxpython-this-program-needs-access-to-the-screen 
On the other hand, TraitsUI test suite for wxPython + OSX has been failing.

### DIFF
--- a/.github/workflows/minimal-deps.yml
+++ b/.github/workflows/minimal-deps.yml
@@ -8,7 +8,10 @@ on: pull_request
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    strategy:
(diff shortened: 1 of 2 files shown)


mdickinson (approved):
LGTM

PR #1422, opened 2020-12-01
Fix changelog section title for consistency

kitchoi:
[Targeting maint/7.1]

When I build the documentation, I realize that the section title "Bugfix release" in the changelog is really an odd one out. This PR fixes the title to be more consistent.


rahulporuri (approved):
LGTM. Should've caught that myself.

Issue #1419, opened 2020-11-30
Bug fix release 7.1.1


kitchoi:
Several bug fixes have gone into master since 7.1.0 was released. It might be good to have a bug fix release 7.1.1.
See PRs which may be backported: https://github.com/enthought/traitsui/pulls?q=is%3Apr+is%3Aclosed+label%3A%22need+backport+to+7.1%22

Pre-release
---
- [x] Check and label PRs that need to be backported to 7.1
- [x] Create a new branch from `maint/7.1`, cherry pick PRs labeled with 'need backport to 7.1' to it. e.g. `get cherry-pick -x [merge-commit-hash]`, update changelog, then open a PR against `maint/7.1`. (#1421)
- [x] Once this PR is merged, remove the labels from the backported PRs
- [x] Test against other ETS packages and other traitsui-using projects
- [x] Check MANIFEST and requirements are still up to date.
- [x] Test building the documentation

Release
-------
- [x] Create a new branch from the `maint/7.1` branch. (#1423)
        - Set release to `7.1.1`, and set `IS_RELEASED` is true; commit
        - Install from source distribution and run tests again. `git clean -ffxd`, `python setup.py sdist`, `pip install dist/traitsui-version.tar.gz[test]`, `cd anywhere-not-the-traitsui-repo-root`, `python -m unittest discover traitsui`
        - Open a PR targeting `maint/7.1`
- [x] Once the previous PR is merged, create another PR targeting `maint/7.1` (#1424)
        - Bump the version to `7.1.2` and set `IS_RELEASED` to false; commit
- [x] Tag (annotated!) the release commit (where `IS_RELEASED` is true), and push the tag (e.g. `git tag -a -m "Release <release version number>" <release version number> <commit-hash>`, `git push origin <release version number>`)
- [x] Make PR targeting `gh-pages` branch: Generate documentation and copy the content to the branch. Verify that the resulting docs/index.html looks good. (#1425)
- [x] Upload to PyPI
- [x] Test the PyPI package
- [x] Announcement

Post-release
-------------
- [x] Package update for `enthought/free` repository (for EDM)
- [x] Backport release note and change log to master (#1426)

kitchoi:
The release is now available on PyPI and EDM. Announcement has been made. Closing.

PR #1423, opened 2020-12-01
TraitsUI Release 7.1.1

kitchoi:
For #1419

I will open a separate PR to flip IS_RELEASED to false and bump the version after this one is merged.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ MAJOR = 7
 MINOR = 1
 MICRO = 1
 PRERELEASE = ""
-IS_RELEASED = False
+IS_RELEASED = True
 
 # If this file is part of a Git export (for example created with "git archive",
 # or downloaded from GitHub), ARCHIVE_COMMIT_HASH gives the full hash of the


rahulporuri (approved):
LGTM. ~I haven't run any tests locally.~ Ran the testsuite locally on windows with python 3.6 and pyqt5 and there are no issues.

kitchoi:
For the record, I ran the test with the source distribution in a Python 3.9 environment, and this is what I did with this branch:
```
git clean -ffxd
rm -rf ../py39-venv/traitsui-env/
python3.9 -m venv ../py39-venv/traitsui-env
source ../py39-venv/traitsui-env/bin/activate
pip install -U pip setuptools wheel
python setup.py sdist
pip install dist/traitsui-7.1.1.tar.gz 
cd
python -m unittest discover traitsui
cd -
pip install dist/traitsui-7.1.1.tar.gz[test]
cd
python -m unittest discover traitsui
pip install pyside2
pip install pygments
python -m unittest discover traitsui
```
All tests pass (with some expected skips).
So I tested with both the null and qt toolkits. Since the test suite is known to fail on wx + OSX, I did not test with wx.

PR #1427, opened 2020-12-01
Fix cron job not installing additional dependencies

kitchoi:
The cron job on Travis is currently failing with this:
```
======================================================================
ERROR: traitsui.testing.tests.test_gui (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: traitsui.testing.tests.test_gui
Traceback (most recent call last):
  File "/home/travis/.edm/envs/traitsui-test-3.6-null/lib/python3.6/unittest/loader.py", line 428, in _find_test_path
    module = self._get_module_from_name(name)
  File "/home/travis/.edm/envs/traitsui-test-3.6-null/lib/python3.6/unittest/loader.py", line 369, in _get_module_from_name
    __import__(name)
  File "/home/travis/.edm/envs/traitsui-test-3.6-null/lib/python3.6/site-packages/traitsui/testing/tests/test_gui.py", line 17, in <module>
    from pyface.api import GUI
  File "/home/travis/.edm/envs/traitsui-test-3.6-null/lib/python3.6/site-packages/pyface/api.py", line 24, in <module>
    from .gui_application import GUIApplication
  File "/home/travis/.edm/envs/traitsui-test-3.6-null/lib/python3.6/site-packages/pyface/gui_application.py", line 39, in <module>
    from .i_splash_screen import ISplashScreen
  File "/home/travis/.edm/envs/traitsui-test-3.6-null/lib/python3.6/site-packages/pyface/i_splash_screen.py", line 20, in <module>
    from pyface.image_resource import ImageResource
  File "/home/travis/.edm/envs/traitsui-test-3.6-null/lib/python3.6/site-packages/pyface/image_resource.py", line 19, in <module>
    ImageResource = toolkit_object("image_resource:ImageResource")
  File "/home/travis/.edm/envs/traitsui-test-3.6-null/lib/python3.6/site-packages/pyface/base_toolkit.py", line 152, in __call__
    module = import_module(mname, package)
  File "/home/travis/.edm/envs/traitsui-test-3.6-null/lib/python3.6/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "/home/travis/.edm/envs/traitsui-test-3.6-null/lib/python3.6/site-packages/pyface/ui/null/image_resource.py", line 19, in <module>
    from pyface.i_image_resource import IImageResource, MImageResource
[...]

PR #1430, opened 2020-12-02
Remove job on Travis for testing against Traits 6.0

kitchoi:
This is motivated by the difficulties in fixing the cron job on Travis CI (see attempts in #1427 and #1395). By removing the dependency requirement given to EDM for pulling Traits 6.0, we would be able to fix the cron job on Travis more easily.

A job has been added on GitHub Actions for testing against Traits 6.0 (see #1415), so we are still testing against Traits 6.0.  However note that to conserve resources, the job on GitHub Actions is currently scheduled to be run per week, rather than per-PR.  We can change that event trigger setting too.


kitchoi:
After this change, I think I can resurrect either #1427 and #1395 to fix the cron job on Travis.

mdickinson (approved):
LGTM

PR #1433, opened 2020-12-07
Use correct source_path in example

### LINKED ISSUE #1432: Incorrect search_path in example
As mentioned in https://github.com/enthought/pyface/issues/830#issuecomment-738954788 and the subsequent comments on that issue, the search_path in `Image_editor_demo` refers to a non existent path in traits.  This leads to the image never being able to be located, and without https://github.com/enthought/pyface/pull/831 actually, causes the corresponding intergration test to fail on the latest pyface.

The path should be updated to: `search_path = [join(dirname(traitsui.api.__file__), '..', 'examples', 'demo', 'Extras')]`, and note to do this, `traitsui.api` must be imported directly.

aaronayres35:
fixes #1432 

This PR changes the search path the be the correct path to the directory that contains the image folder containing the images of interest.

### DIFF
--- a/traitsui/examples/demo/Extras/Image_editor_demo.py
+++ b/traitsui/examples/demo/Extras/Image_editor_demo.py
@@ -6,16 +6,12 @@ A simple demonstration of how to use the ImageEditor to add a graphic element
 to a Traits UI View.
 """
 
-import traits
-
 # Imports:
 from os.path \
     import join, dirname
 
 from traits.api \
     import HasTraits, Str
-
-import traitsui.api
 
 from traitsui.api \
     import View, VGroup, Item, ImageEditor
--- a/traitsui/examples/demo/Extras/Image_editor_demo.py
+++ b/traitsui/examples/demo/Extras/Image_editor_demo.py
@@ -25,8 +25,8 @@ from pyface.image_resource \
 
 # Constants:
 
-# Necessary because of the dynamic way in which the demos are loaded:
-search_path = [join(dirname(__file__))]
+# The images folder is in the same folder as this file:
+search_path = [dirname(__file__)]
 
 # Define the demo class:
 
--- a/traitsui/examples/demo/Extras/Image_editor_demo.py
+++ b/traitsui/examples/demo/Extras/Image_editor_demo.py
@@ -15,11 +15,10 @@ from os.path \
 from traits.api \
     import HasTraits, Str
 
-from traitsui.api \
-    import View, VGroup, Item
+import traitsui.api
 
 from traitsui.api \
-    import ImageEditor
+    import View, VGroup, Item, ImageEditor
 
 from pyface.image_resource \
     import ImageResource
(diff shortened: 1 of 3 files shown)


traitsui/examples/demo/Extras/Image_editor_demo.py
```diff
@@ -15,20 +15,18 @@
...
 from traitsui.api \
-    import ImageEditor
+    import View, VGroup, Item, ImageEditor
 
 from pyface.image_resource \
     import ImageResource
 
 # Constants:
 
 # Necessary because of the dynamic way in which the demos are loaded:
```
kitchoi, line 28: If I run the script from the command line, even just `search_path = []` would work, but that's using some default magic (with inspection on where the code is running from etc.). I believe this is the magic:
https://github.com/enthought/traits/blob/a0d0a70331ec202eb33b6439570175c869fe353c/traits/trait_base.py#L256-L305

That would not work when this example is run from within the demo application. `join(dirname(__file__))` remove the assumption and dependency on the magic: It is more explicit, which I think is a good thing to do anyway. Now the comment makes it look worse than this actually is.

What do you think about rewording the comment to something along the line of `# The images folder is in the same folder as this file`?

traitsui/examples/demo/Extras/Image_editor_demo.py
```diff
@@ -15,20 +15,18 @@
...
-search_path = [join(dirname(traits.api.__file__),
-                    '..', '..', 'examples', 'demo', 'Extras')]
+search_path = [join(dirname(__file__))]
```
kitchoi, line 29: The `join` is actually redundant now that there are no fragments to join. Sorry about that. `[dirname(__file__)]` would do.

kitchoi (commented):
May I also suggest adding a simple test to prevent this missing file link from happening again?
Something like this.
```
import os
import runpy
import unittest

#: Filename of the demo script
FILENAME = "Image_editor_demo.py"

#: Path of the demo script
DEMO_PATH = os.path.join(os.path.dirname(__file__), "..", FILENAME)


class TestImageEditorDemo(unittest.TestCase):

    def test_image_path_exists(self):
        search_path, = runpy.run_path(DEMO_PATH)["search_path"]
        self.assertTrue(os.path.exists(search_path))

```

traitsui/examples/demo/Extras/tests/test_Image_editor_demo.py
```diff
@@ -0,0 +1,22 @@
+import os
```
kitchoi, line 1: The emptiness of at the top of the file is a bit disturbing... It needs a docstring, and possibly a copyright header.

traitsui/examples/demo/Extras/Image_editor_demo.py
```diff
@@ -15,20 +15,18 @@
 from traits.api \
     import HasTraits, Str
 
-from traitsui.api \
-    import View, VGroup, Item
+import traitsui.api
```
kitchoi, line 18: This added line is not needed any more (neither is the `import traits` at the top?)

kitchoi (approved):

kitchoi (approved):
Still LGTM. Thank you!

PR #1436, opened 2020-12-10
Drop Python 3.5 support in TraitsUI

### LINKED ISSUE #1435: Declare dropping Python 3.5 support
CI against Python 3.5 has been removed (https://github.com/enthought/traitsui/pull/615) and Python 3.5 has passed end-of-life.

This issue is for officially declare TraitsUI stops supporting Python 3.5, which includes updating setup.py's classifiers metadata, and possibly the `python_requires` metadata. This will make package installer avoid the next release of traitsui and fall back to the latest release that does not have this restriction.

Related: https://github.com/enthought/pyface/issues/845

kitchoi:
Closes #1435

This PR removes Python 3.5 from etstool.py (it is not used by CI), and most importantly, updated setup.py so that package installer will no longer pick up the next release in a Python <=3.5 environment.

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -51,7 +51,7 @@ using::
 
     python etstool.py test_all
 
-Currently supported runtime values are ``3.5`` and ``3.6``, and currently
+Currently supported runtime values are ``3.6``, and currently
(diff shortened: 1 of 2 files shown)


aaronayres35 (approved):
LGTM

PR #1437, opened 2020-12-10
Declare Python 3.5 not supported for ets-demo

kitchoi:
Similar to #1436 but for etsdemo.

etsdemo is an application and is a separate distribution, but in any case, we have not tested the application with Python 3.5 or lower.


aaronayres35 (approved):
LGTM

Issue #1428, opened 2020-12-01
etsdemo 'Run' button does not execute script in __main__ block


kitchoi:
If one has a script that contains a block like this:
```
if __name__ == "__main__":
    ...
```
Trying to run this script from the `etsdemo` application by pressing the 'Run' button will not run that block, which is unexpected.

To reproduce, create a file like this:
```
if __name__ == "__main__":
    print("Hello")
```
Then run this:
```
from etsdemo.main import main
main(
    [
        {
            "version": 1,
            "name": "Some name",
            "root": ".",  # or a path to a directory that contains the above test file.
        },
    ]
)
```

That block is not run when one presses the 'Run' button.
![Dec-01-2020 18-03-35](https://user-images.githubusercontent.com/3673984/100778952-91291d80-33ff-11eb-9039-10147935b0e1.gif)

But if one changes the "__main__" to "___main___" (three underscores), then it would run:
![Dec-01-2020 18-01-11](https://user-images.githubusercontent.com/3673984/100778706-3ee7fc80-33ff-11eb-9325-79328b3d5141.gif)

From the snapshot of an old version of the application, this `___main___` is not a typo: It looks deliberate.
https://github.com/enthought/traitsui/blob/96f45ff9c51ff55b5724cbed2361b933f468c183/enthought/traits/ui/extras/demo.py

Looks like the `run_code` method is serving two purposes:
(1) When the file is clicked, it loads the file to find this specially named variable 'demo'. 
(2) When the Run button is clicked, it runs the file too.

For the first purpose, we don't want the block in `if __name__ == "__main__"` to run. For the second purpose, we do.

corranwebster:
Note also the commented out code here: https://github.com/enthought/traitsui/blob/96f45ff9c51ff55b5724cbed2361b933f468c183/enthought/traits/ui/extras/demo.py#L150-L151

Thoughts:

- using `___main___` is bad because then examples can't be run stand-alone from the command-line
- changing `sys.modules['__main__'].__file__` a few lines later is highly dubious (and will potentially have weird impacts - e.g. this used by traits.etsconfig to guess the name of the current app and then after that the directories where to put configuration, etc...)
- perhaps the convention should be to have an optional `main()` function, and `if __name__ == '__main__':` just calls it?

kitchoi:
> using ___main___ is bad because then examples can't be run stand-alone from the command-line

[...]

PR #1444, opened 2020-12-15
Add CLI commands to ets-demo/etstool.py for generating and uploading application bundles

### LINKED ISSUE #1379: [ETS-Demo] Consider shipping an EDM/EAM application after traitsui/pyface 7.1.0
With TraitsUI 7.1.0, the etsdemo application will be populated with TraitsUI examples. It might therefore be a good idea to create and ship an EDM/EAM application, similar to "mayavi_demo", for the ets-demo application.

aaronayres35:
closes #1379

This PR adds the CLI commands `generate_bundles` and `upload_bundles` to the `ets-demo` `etstool.py` file.  I have tested that creating an edm environment from the bundles and calling `etsdemo` launches the application and things work as expected.  Once we have decided on / have access to a repository we can test the upload.  In order to do so, we will need to add an `edm.yaml` file, however with `enthought/free` which we plan to upload to in the end, that configuration file is not needed. 

IIUC we will need to upload one bundle per platform, so all 3 bundles are generated.


ets-demo/etstool.py
```diff
@@ -311,6 +336,62 @@ def flake8(runtime, toolkit, environment):
...
+    """
+
+    for platform in PLATFORMS:
+        bundle_dir = os.path.join("bundle", platform)
+        if not os.path.exists(bundle_dir):
+            os.makedirs(bundle_dir)
+
+        bundle_file = os.path.join(
+            bundle_dir,
+            f"etsdemo_{platform.replace('-', '_')}-{APP_BUNDLE_VERSION}.bundle"
```
aaronayres35, line 352: per the hatcher docs: "The filename of the bundle is expected to follow the <name-with-underscores>-<version>-<build>.bundle template." 

However, IIUC we will need to upload a bundle for every platform we wish to support? 
"Exported bundles are platform-specific, and cannot be used to reproduce a similar environment on a different platform, e.g. you cannot use an osx-x86_64 bundle to create an environment on a Linux machine."

ets-demo/etstool.py
```diff
@@ -311,6 +336,62 @@ def flake8(runtime, toolkit, environment):
...
+        )
+        if os.path.exists(bundle_file):
+            print(bundle_file)
+            BUNDLES.append((bundle_file, platform))
+
+    for bundle, platform in BUNDLES:
+        cmd = "hatcher bundles upload enthought ets {platform} {bundle}"
```
aaronayres35, line 391: The docs say that the templaate for this command is `hatcher bundles upload <organization-name> <repository-name> <platform> <filename>`.  I'm unsure if `<filename>` includes the path or not.  
The code currently has `{bundle}` which is for example `bundle/rh7-x86_64/etsdemo_rh7_x86_64-1.0-1.bundle`.  I think this will likely need fixing, but I don't know if the hatcher command is supposed to be run from the sam directory as the bundle files?

ets-demo/etstool.py
```diff
@@ -311,6 +336,62 @@ def flake8(runtime, toolkit, environment):
...
+            bundle_file,
+        ] + sorted(full_app_dependencies)
+        try:
+            subprocess.check_call(edm_command)
+        except subprocess.CalledProcessError:
+            click.echo("Failed to generate bundle {}".format(bundle_file))
+
+
+@cli.command(name="upload-bundles")
+@click.option('--repository', default=DEFAULT_REPOSITORY)
```
aaronayres35, line 376: More machinery will be needed to use non `enthought/free` repository.  "If the bundle is uploaded to any repository other than enthought/free, the edm yaml config file needs to be updated with a new bundle_repositories entry, pointing to the repository containing the bundle file."

ets-demo/etstool.py
```diff
@@ -127,6 +130,21 @@
 
 doc_dependencies = {}
 
+full_app_dependencies = [
+    'eam',
+    'traitsui',
+    'etsdemo',
+    'envisage',
```
aaronayres35, line 137: 'envisage' currently has examples/demos contributed to etsdemo, but because of #1428, the examples effectively do not run currently.  
We should either wait until this is resolved to upload the bundles, or not include `envisage` when we first upload the bundles.

aaronayres35:
After the latest commit, I ran
```
python etstool.py generate-bundles
edm environments import NEW_ENV -f bundle/osx-x86_64/etsdemo_osx_x86_64-1.0-1.bundle --force
edm shell -e NEW_ENV
etsdemo
```
as a sanity check, and in the environment created from the bundle, `etsdemo` worked as expected.

ets-demo/etstool.py
```diff
@@ -127,6 +130,23 @@
...
+    "apptools",
+    "chaco",
+    "h5py",
+    "numpy",
+    "pandas",
+    "pygments",
+    "pytables",
+    "pyqt5"
```
aaronayres35, line 145: We need to have a single specified toolkit.  `pyqt5` may ultimately cause problems do to `QtWebKit` issues.  See  https://github.com/enthought/enable/pull/494#issuecomment-745473789
We may want `pyside2` instead?   I'm not sure, but we will need to settle on a single toolkit.

ets-demo/etstool.py
aaronayres35: I removed envisage

ets-demo/etstool.py
rahulporuri: > However, IIUC we will need to upload a bundle for every platform we wish to support?

That is correct but they don't need to have unique names. When the upload happens, the server parses the incoming file and understands what platform the bundle is used for.

So, we can remove the platform names from the bundle files.

ets-demo/etstool.py
rahulporuri: We are running the `etstool.py` script - which then uploads the bundles via the command - and the `bundle` cmdline argument needs to be a full path.

ets-demo/etstool.py
rahulporuri: This won't be a problem as we only intend to upload to `enthought/free` - and all packages we rely on will be available via `enthought/free`

ets-demo/etstool.py
rahulporuri: For now, we have more stability with pyqt5 than pyside2, if im not wrong. We can always replace the underlying toolkit in the future so for now pyqt5 is fine.

rahulporuri:
> edm environments import NEW_ENV -f bundle/osx-x86_64/etsdemo_osx_x86_64-1.0-1.bundle --force

The right command to test the bundles with is - `edm app install etsdemo-test-2 -f .\bundle\win-x86_64\etsdemo_win_x86_64-1.0-1.bundle`

And then, open EDM GUI
- check that ETS Demo is listed
- launch ETS Demo and check that the application is actually launched

ets-demo/etstool.py
```diff
@@ -311,6 +336,64 @@ def flake8(runtime, toolkit, environment):
...
+        bundle_file = os.path.join(
+            bundle_dir,
+            f"etsdemo_{platform.replace('-', '_')}-{APP_BUNDLE_VERSION}.bundle"
+        )
+        if os.path.exists(bundle_file):
+            BUNDLES.append((bundle_file, platform))
+
+    for bundle, platform in BUNDLES:
+        cmd = "hatcher bundles upload enthought {repository} {platform} {bundle}"
+        execute(
```
rahulporuri, line 391: the use of `subprocess.check_call` vs `execute` when generate vs update bundles is slightly jarring. Is there a reason why we can't just use `subprocess.check_call` here? Is it more complicated/convoluted?

ets-demo/etstool.py
```diff
@@ -311,6 +339,72 @@ def flake8(runtime, toolkit, environment):
...
+@cli.command(name="upload-bundles")
+@click.option('--repository', default=DEFAULT_REPOSITORY)
+def upload_bundles(repository):
+    """ Upload the generated bundles """
+    BUNDLES = []
+    for platform in PLATFORMS:
+        bundle_dir = os.path.join("bundle", platform)
+        bundle_file = os.path.join(
+            bundle_dir,
+            f"etsdemo_{platform.replace('-', '_')}-{APP_BUNDLE_VERSION}.bundle"
```
rahulporuri, line 387: The generated bundles don't have this name format anymore. It'll be useful to also refactor the bundle name into a module-level constant that is used both when generating and uploading the bundles.

aaronayres35:
@rahulporuri 
I have this PR still needing review, not sure when we want to try and push this through/upload the application bundles.  

I figure with #1466, and some potential upcoming etsdemo changes, this can at least hold off until after the spring planning meeting

rahulporuri (approved):
LGTM. This PR has been stuck here for a long time. Let's push this ahead as-is and make changes/fixes as necessary when we need to.

PR #1445, opened 2020-12-16
Add links to API docs in a etsdemo examples for traitsui standard editors

rahulporuri:
This PR adds ReST links to API docs for editors in the contributed etsdemo examples for traitsui standard editors.

Note : I originally assumed that when clicked on the link, the user would simply be redirected to their default browser and the link API docs link would be opened in the browser. Interestingly, i'm not sure how but the API doc is rendered in the middle pane of the etsdemo application. This leads to the slightly awkward behavior where the user cant go back to read the module docstring/description if they open the link - unless they view a different example.

The user can still right-click on the link to copy the link and open it in a browser of their choice.

api-docs-in-etsdemo

PR #1446, opened 2020-12-16
Open links externally instead of in the html editor in etsdemo application

rahulporuri:
This PR updates the etsdemo application such that clicking on links in the description will be opened in the default browser instead of in the html editor itself. This behavior was observed in #1445 and was recommended as the right user experience - https://github.com/enthought/traitsui/pull/1445#issuecomment-746182226

### DIFF
--- a/ets-demo/etsdemo/app.py
+++ b/ets-demo/etsdemo/app.py
@@ -873,8 +873,6 @@ demo_file_view = View(
             editor=HTMLEditor(
                 format_text=True,
                 base_url_name='base_url',
-                # Open links externally, in the default browser, instead of
-                # in the editor.
                 open_externally=True,
             ),
         ),
--- a/ets-demo/etsdemo/app.py
+++ b/ets-demo/etsdemo/app.py
@@ -873,6 +873,9 @@ demo_file_view = View(
             editor=HTMLEditor(
                 format_text=True,
                 base_url_name='base_url',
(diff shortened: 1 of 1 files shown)


ets-demo/etsdemo/app.py
```diff
@@ -873,6 +873,9 @@ def get_children(self):
             editor=HTMLEditor(
                 format_text=True,
                 base_url_name='base_url',
+                # Open links externally, in the default browser, instead of
+                # in the editor.
```
kitchoi, line 877: I guess the attribute name is _almost_ expressive enough to not require a separate comment. `open_link_externally` would have been a better name, but it is harder to change a published public API. Would an inline `open_externally=True,    # for links` be informative enough without hurting readability?

kitchoi (approved):
LGTM. Minor nitpick.

ets-demo/etsdemo/app.py
corranwebster: I'd just drop the comment.  `HTMLEditor` is reasonably well documented so it can be looked up if there is a question about what it does.

kitchoi:
My bad I did not manually test this with my usual PySide2-from-PyPI exercise.
Thanks to @aaronayres35 testing this. We might need to revert this because this switch makes the entire description screen blank.
Screenshot 2020-12-16 at 16 55 56

Issue #1448, opened 2020-12-16
Use of API from QtWebKit that is not available in QtWebEngine causing problems 


aaronayres35:
Issue motivated from error observed in this comment: https://github.com/enthought/traitsui/pull/1445#pullrequestreview-553865200

Basically, trying to run etsdemo with the changes from #1445 and #1446 fails because we end up trying to use `setLinkDelegationPolicy` from the QtWebKit API that is not available with QtWebEngine

See https://github.com/enthought/traitsui/pull/1445#issuecomment-746601596 as well

PR #1449, opened 2020-12-16
Revert "Open links externally instead of in the html editor in etsdemo application"

kitchoi:
Reverts enthought/traitsui#1446

Unfortunately this does not work well with PySide2 (with QtWebEngine) due to #1448.

### DIFF
--- a/ets-demo/etsdemo/app.py
+++ b/ets-demo/etsdemo/app.py
@@ -873,7 +873,6 @@ demo_file_view = View(
             editor=HTMLEditor(
                 format_text=True,
                 base_url_name='base_url',
-                open_externally=True,
             ),
         ),
         VSplit(


rahulporuri (approved):
LGTM

corranwebster:
This is the wrong fix - open links externally is the correct behaviour, since we don't have the full browser navigation available in a simple html view.  The bug is not with the demo app, but with the way that we are hacking in WebEngine support in Pyface; and this issue will affect every web view which uses this option.

Short-term fix is to have the HTML editor detect which API call is available and use the right one.

The right fix is to spend the effort to get Pyface's web support into a sane state (it's becoming increasingly clear that WebEngine _isn't_ a drop-in replacement for WebKit) and then have the Qt HTMLEditor do some sort of mapping of APIs (or present a lower-level common API in Pyface or something).

kitchoi:
> This is the wrong fix - open links externally is the correct behaviour, since we don't have the full browser navigation available in a simple html view. The bug is not with the demo app, but with the way that we are hacking in WebEngine support in Pyface; and this issue will affect every web view which uses this option.
>
> Short-term fix is to have the HTML editor detect which API call is available and use the right one.
>
> The right fix is to spend the effort to get Pyface's web support into a sane state (it's becoming increasingly clear that WebEngine isn't a drop-in replacement for WebKit) and then have the Qt HTMLEditor do some sort of mapping of APIs (or present a lower-level common API in Pyface or something).

Agreed that the HTML editor needs to be fixed. This PR does not attempt to close or fix #1448.
However IMO, the behaviour about opening a link is really really minor in comparison with the lost of the entire view and crashing of the etsdemo application. We can resurrect this flag once HTML editor is fixed and released.

I will leave this to @rahulporuri and @corranwebster whether to merge.

kitchoi:
For reference, #1450 is blocked by this PR, and any attempt to add tests for meaningfully test the etsdemo view will be blocked by this error as well.

kitchoi:
> I will leave this to @rahulporuri and @corranwebster whether to merge.

Actually, it might be best to flip the option :D 
If this should not be merged, please go ahead close this PR. Otherwise I will merge this later in my next free cycle to avoid PRs going stale and unresolvable.

corranwebster:
I have a draft PR that should do the quick hacky fix (but is completely untested, and I don't have time to finish it properly).

corranwebster:
Closing in favour of #1451 or something along those lines.

PR #1456, opened 2020-12-21
Hide demo tab when the demo is None

### LINKED ISSUE #1454: Remove demo tab from the etsdemo application
The demo tab works well when the content is like the examples from TraitsUI where there is always a GUI that can be run as a subpanel within the etsdemo.

However the UX is awkward for when the application is used against examples where there isn't a GUI, e.g. examples from traits.

The existence of the demo tab also adds a lot of complexity in the handling of the `if __name__ == "__main__"` block: When the run button is run and if there is a variable named `demo` or `popup`, the application skips the `__name__ == "__main__"` block because it wants to refresh the demo tab instead (see https://github.com/enthought/traitsui/issues/1428)

corranwebster:
This adds an alternative group in demo file view for when demo is None.

Unfortunately we can't just hide the "Demo" tab, because dynamic hiding doesn't work.

This should essentially fix #1454

### DIFF
--- a/ets-demo/etsdemo/app.py
+++ b/ets-demo/etsdemo/app.py
@@ -887,27 +887,49 @@ demo_file_view = View(
                     visible_when="source is not None",
                 ),
             ),
-            Tabbed(
-                UItem(
-                    "demo",
-                    style="custom",
-                    resizable=True,
-                ),
-                Item(
-                    "log",
-                    style="readonly",
-                    editor=CodeEditor(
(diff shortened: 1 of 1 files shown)


aaronayres35 (approved):
LGTM! (although code is a little clunky from tabs not working with `visible_when` but thats orthogonal)
I tested with the envisage examples that don't have a demo GUI component and everything looked correct (and I verified things still look the same when there is a demo GUI).

PR #1457, opened 2020-12-22
Add help aliases to etsdemo etstool module

rahulporuri:
This PR makes `-h` an alias to `--help` in the etstool module in ets-demo distribution.

### DIFF
--- a/ets-demo/etstool.py
+++ b/ets-demo/etstool.py
@@ -143,9 +143,9 @@ environment_vars = {
     'null': {'ETS_TOOLKIT': 'null'},
 }
 
-CONTEXT_SETTINGS = dict(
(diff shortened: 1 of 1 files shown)


ets-demo/etstool.py
```diff
@@ -143,8 +143,12 @@
     'null': {'ETS_TOOLKIT': 'null'},
 }
 
+CONTEXT_SETTINGS = dict(
```
aaronayres35, line 146: NIT: Use a dictionary literal here instead

aaronayres35 (approved):
LGTM! 
I tested locally and it works as expected.  Just one super nit picky comment, feel free to ignore.

Issue #1464, opened 2021-01-05
Qt HTMLEditor does not open link when HTML/user gesture opening links externally


kitchoi:
Qt WebKit/WebEngine supports opening a link in a new window when the HTML author instructs it to do so with `target='_blank'`:
```
<a href='https://example.com' target='_blank'>Target to _blank Link</a>
```
The user can also elect to open a link in a new window by selecting "Open Link in New Window" from the context menu. 

Unexpected behaviour: HTMLEditor does not open the link at all for the above gestures.
Expected behaviour: HTMLEditor will honour the above gestures and allow the link to be open in a new window. What is a "new window" depends on the context. But given the existing behaviour of the `open_externally` flag on the HTMLEditor is to open a link in the user's machine default web browser, I'd expect that new window to mean the same.

To demonstrate with this example code:

<details>
    <summary> Code snippet </summary>

```

from traits.api import HasTraits, HTML
from traitsui.api import UItem, View, HTMLEditor


html = """
    <h1><a href='https://example.com'>Target to default Link</a></h1>
    <h1><a href='https://example.com' target='_blank'>Target to _blank Link</a></h1>

"""

class HTMLEditorDemo(HasTraits):

    my_html_trait = HTML(html)

    traits_view = View(
        UItem(
            'my_html_trait',
            editor=HTMLEditor(format_text=True)
        ),
        title='HTMLEditor',
        buttons=['OK'],
        width=800,
        height=600,
        resizable=True
    )


# Create the demo:
demo = HTMLEditorDemo()

# Run the demo (if invoked from the command line):
if __name__ == '__main__':
    demo.configure_traits()

```

</details>

This is the current behaviour (video is created using PySide2 + WebEngine, but PyQt5 + WebKit behaves in pretty much the same way except the "Back" option is not available from the context menu):

https://user-images.githubusercontent.com/3673984/103654983-89105080-4f5e-11eb-9182-1dac2d38a9e6.mp4

If I add `open_externally=True` to `HTMLEditor`, like this:
```
        UItem(
            'my_html_trait',
            editor=HTMLEditor(format_text=True, open_externally=True)
        ),
```

Then all that changes is that the link is open in the default web browser when I click the link without `target='_blank'`, gestures to explicitly open the link in a new window still results in the link not being opened at all:

https://user-images.githubusercontent.com/3673984/103655265-e5737000-4f5e-11eb-98c0-bf48e5559c31.mp4

kitchoi:
[...]

PR #1465, opened 2021-01-05
Refactor and extend tests for Qt HTMLEditor handling of opening links

kitchoi:
Motivated by #1464, this PR extends the tests for testing how the HTMLEditor opens `<a>` links with/without `target` set, along with the `open_externally` flag.

The tests rely on Qt webkit/webengine being able render CSS so that the link can fill up the entire page and the tests can simply click anywhere to trigger the link.

Only test code is changed. Two of the tests are effectively tests with "expected failures" to illustrate the issue described in #1464.

Note the draft implementation in https://github.com/enthought/traitsui/issues/1464#issuecomment-754662151 requires the existing test be modified as the slot to `linkClicked` (WebKit only) is removed there. The draft implementation passes the new tests (after replacing expected failures with expected successes).

### DIFF
--- a/traitsui/tests/editors/test_html_editor.py
+++ b/traitsui/tests/editors/test_html_editor.py
@@ -124,7 +124,7 @@ def wait_for_qt_signal(qt_signal, timeout):
         raise RuntimeError("Timeout waiting for signal.")
 
 
-def qt_allow_page_to_load(page, step_timeout=0.1):
+def qt_allow_page_to_load(page, step_timeout=0.5):
     """ Allow QWebPage/QWebEnginePage to finish loading.
 
     Out of context, this function does not know if the page has started
--- a/traitsui/tests/editors/test_html_editor.py
+++ b/traitsui/tests/editors/test_html_editor.py
@@ -16,10 +16,12 @@ from traitsui.api import HTMLEditor, Item, View
 from traitsui.tests._tools import (
     BaseTestMixin,
     create_ui,
+    is_qt,
     requires_toolkit,
     reraise_exceptions,
     ToolkitName,
 )
+from traitsui.testing.api import MouseClick, TargetRegistry, UITester
 
 
 class HTMLModel(HasTraits):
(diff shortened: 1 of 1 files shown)


kitchoi:
Hmmm, it is strange that one of the tests would hang on Qt4, and I am unable to reproduce locally. Marking as WIP for now.

kitchoi:
> Hmmm, it is strange that one of the tests would hang on Qt4, and I am unable to reproduce locally. 

It wasn't reproducible on OSX but reproducible on Linux. It is resolved now.

traitsui/tests/editors/test_html_editor.py
```diff
@@ -74,32 +255,129 @@ def test_base_url_changed(self):
...
-    def test_open_externally_qt(self, open_new_mock):
-        from pyface.qt import QtCore, QtWebKit
+    def test_open_internal_link(self):
+        # this test requires Qt because it relies on the link filling up
+        # the entire page through the use of CSS, which isn't supported
+        # by Wx.
+        model = HTMLModel(content="""
+        <html>
+            <a
+              href='/#'
```
aaronayres35, line 265: ~why is `href='/#' here, but in the `test_open_internal_link_externally` test it is `href='test://testing'`?~

EDIT: I think I understand now

traitsui/tests/editors/test_html_editor.py
```diff
@@ -74,32 +255,129 @@ def test_base_url_changed(self):
             model.model_base_url = "/new_dir"
 
     @requires_toolkit([ToolkitName.qt])
-    @mock.patch('webbrowser.open_new')
-    def test_open_externally_qt(self, open_new_mock):
-        from pyface.qt import QtCore, QtWebKit
+    def test_open_internal_link(self):
+        # this test requires Qt because it relies on the link filling up
+        # the entire page through the use of CSS, which isn't supported
+        # by Wx.
```
aaronayres35, line 261: This explains the use of `style='display:block; width: 100%; height: 100%'` in the tests, and the implementation of `qt_mouse_click_web_view` which "Performs a mouse click at the center of the web view".  I'm assuming this was done this way because finding the location of the link itself in the page to be clicked is too difficult.  If it were possible though these tests (or ones similar with possibly different expected behaviors) could be done with wx correct?

aaronayres35 (approved):
LGTM!
Just a couple questions, but no suggested changes

traitsui/tests/editors/test_html_editor.py
```diff
@@ -30,26 +32,205 @@ class HTMLModel(HasTraits):
...
+    -------
+    html : str
+    """
+    qt_allow_view_to_load(page)
+    if _is_webkit_page(page):
+        return page.mainFrame().toHtml()
+
+    content = []
+    page.toHtml(content.append)
+    qt_allow_view_to_load(page)
```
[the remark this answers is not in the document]
kitchoi: This is because `QWebEnginePage.toHtml` is asynchronous. https://doc.qt.io/qt-5/qwebenginepage.html#toHtml
If this `qt_allow_view_to_load` is removed here, then the content returned will be empty.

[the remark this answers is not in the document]
aaronayres35: I think I see now, thanks

traitsui/tests/editors/test_html_editor.py
kitchoi: > I'm assuming this was done this way because finding the location of the link itself in the page to be clicked is too difficult. I'm assuming this was done this way because finding the location of the link itself in the page to be clicked is too difficult.

I believe it is possible. Both Qt and Wx have methods for finding text in the HTML view and having them selected. I think what happened here is that I took the shorter path so we at least have some tests to work with for #1464. The thinking goes: modify/add tests for refactoring, refactor, add failing tests for changing behaviour, change the behaviour, then loop again. It is possible that eventually we will have modified these tests a number of times and we will end up with test support code that more powerful and useful. But until then, we can iterate and improve these test support code while keeping them private here.

Issue #1455, opened 2020-12-18
Temporarily modify sys.path when executing example file in etsdemo


kitchoi:
Context: Often demo examples are written assuming they are run from the command line. Sometimes these examples require components outside of that file, e.g. [here](https://github.com/enthought/enable/blob/6cf525cbeefdf34dc6eac5e54d7a73f755ea79b1/examples/enable/canvas_demo.py#L10), [here](https://github.com/enthought/envisage/blob/4e2c07a0aded7360639e379b99b6f05e4dd92210/envisage/examples/demo/plugins/tasks/run_attractor.py#L11-L12). However, when these examples are run from within the etsdemo, the `sys.path` is set for running the `etsdemo` application, and these imports would normally fail because they assume the local directory being on the path.

This inconsistency adds cognitive load to developers maintaining the examples: Developers need to keep in mind that the path is not what they typically expect.

The proposed solution would be to temporarily prepend to the `sys.path` the path of the directory containing the example file, similar to the behaviour when one runs the file from the command line. 

This requires knowing exactly when the run context starts and ends, as the `sys.path` is a global state and it must be reverted back to how it was when the run context for an example terminates.

(Removing the demo tab will determining the life-time of the run context a lot easier, see #1454)

corranwebster:
Thinking about this, local imports from examples are inherently problematic.

For example, let's say we have two related examples, something like this:
```
example_1/
    main.py
    helper.py
example_2/
    main.py
    helper.py
```
In each case assume `main.py` includes an `import helper` statement.

Then even if we mess with `sys.path` to make the imports work, we have a problem that whichever is run first will cause `sys.modules["helper"]` to hold the contents of the local `helper.py`, and then any subsequent `import helper` statements will get _that_ `helper`, not the intended local one.

Messing around with `sys.modules` if even more fraught than messing around with `sys.path`.

The only way to prevent this sort of problem is to run the example in a different interpreter (for simplicity, in a separate process, but 3.9 has subinterpreters which could also solve the problem).

kitchoi:
Acknowledging that the double import trap is indeed a problem for some examples if they are run after another, the implementation of subprocess is not cheap, and it brings a whole loads of other restrictions. See #1466 for a rough comparison. A lot of the existing features in the demo application become difficult to achieve with subprocess.

Subprocess by itself does not solve the problem with Python path, but it does change how we should set the python path (i.e. instead of mutating sys.path as a list, set the PYTHONPATH environment variable). 

[...]

Issue #1124, opened 2020-08-10
Code in demo application should be readonly


rahulporuri:
At the moment, the code shown in the ETS Demo application can be edited by the user. I don't think we allow/save user edits to the code so to remove any ambuigity, the code should be readonly.

![code-readonly](https://user-images.githubusercontent.com/1926457/89761775-ac81bd00-dade-11ea-8905-2a7813ace97b.gif)

kitchoi:
I think the idea is that one can edit the code and press the "Run" button to see the new behaviour caused by the edit.
Is that functionality not working?

rahulporuri:
> I think the idea is that one can edit the code and press the "Run" button to see the new behaviour caused by the edit.
Is that functionality not working?

Ahh. That makes sense. It does work but it wasn't obvious that I could update the code and see the results by manually clicking "Run".

kitchoi:
Agreed some way of indicating the run button can be clicked to test the new code would be a good UX enhancement.

Currently the Run button is always enabled, and I think there is value in keeping its enabled even if the code has not changed (e.g. for running the same thing twice).

Maybe a small instruction next to the button or changing the button text from "Run" to "Re-run" would be a good hint.

rahulporuri:
I updated the issue description to reflect the change.

kitchoi:
Thanks!

kitchoi:
@corranwebster  This UX issue can also be fixed by removing the Demo tab so that the function of Run is more obvious.
At the moment, with the demo UI being visible, it is not obvious that one needs to press the Run button to refresh it. 
The awkwardness is more apparent considering an example that opens a popup:

https://user-images.githubusercontent.com/3673984/103523839-6c97e980-4e74-11eb-9f19-25cce7836f5e.mp4

After editing the code, one needs to press two buttons: "Run", and then "Run demo..." for the new code to be used.
It is tempting to think pressing "Run demo..." is enough, but it isn't.

corranwebster:
There is a potential refactor that could resolve some of these issues.  Following #1456 you can now hide the demo tab by having the `demo` variable be `None` (or not set).

This code https://github.com/enthought/traitsui/blob/5a8f5d4f80ccadb0cdd94ccdb28183ca35dbb432/ets-demo/etsdemo/app.py#L474-L482
[...]

PR #1466, opened 2021-01-06
[POC] Update how code is run in etsdemo

kitchoi:
This PR is intended for scoping efforts and for planning. It is not intended to be merged. When its purpose to proof a concept is fulfilled, it should be closed without merging.

It is marked as draft, but that is because it is not meant to be merged. It can be reviewed now. Note that for efficiency, there exists shortcuts and omissions, allowed purely because this won't be merged.

Motivations
---
- Running example code that does not advertise a magically named `demo` or `popup` is awkward. If the code contains an `if __name__ == '__main__'` block. That block is not run. See #1428
- Showing GUI created from any example as an embedded UI in the "demo" tab may have contributed towards the UX issue where it is not obvious to the user that they had to press the "Run" button to refresh. The UX for when it is a popup is even more obscure (#1124)
- The specially named `demo` or `popup` or  `modal_popup` are too obscure for developers to remember/use when they write/maintain an example script.
- A lot of our examples are written and maintained in repositories assuming they are run as scripts, and some of them assume a Python path including the script folder and import modules locally. Those imports don't work in the etsdemo application because the assumption on the Python path is not fulfilled (#1455)

What is in this POC?
----
- The "Demo" tab for showing an embedded GUI is removed. Pressing the Run button behaves more or less like running the example as a script.
- Extract code running logic and provides two implementations for comparing running the code in the same interpreter as the demo application and running the code in a subprocess.

Local VS Subprocess
----
See the class docstrings of `LocalCodeRunner` and `SubprocessCodeRunner` to find their limitations and tradeoffs.

I did not complete the implementations for reading STDOUT/STDERR for the subprocess approach, because it will take a lot more time: I assume it will be the usual messy approach where one creates a separate thread to read the stream, put new content to a queue, and have the main thread to consume from the queue, and integrate that as part of the GUI event loop. (similar to the behaviour achieved by `asyncio.create_subprocess_shell`, but integrating asyncio and Qt would be another matter.

Try it yourself
----
The demo application on master does not run well with Wx, so the POC is assuming Qt backend.

- To test the difference between running things locally and running things in a subprocess, switch between the two runner implementations in this block in `app.py`:
```
demo_file_handler = DemoFileHandler(
    #runner=SubprocessCodeRunner(),
    runner=LocalCodeRunner(),
)
```
- To test with examples that are already installed in your environment, run `etsdemo` from the command line. 
- To test running other examples that launches a GUI built from a different framework, try testing Pyface examples (e.g. [this](https://github.com/enthought/pyface/blob/7.1.0/examples/data_view/column_example.py)):
```
$ python -c "from etsdemo.main import main; main([{'name': 'pyface', 'version': 1, 'root': '../../pyface/examples'}])"
```
(Change `'../../pyface/examples'` to whatever local path Pyface's examples folder is on your computer)
- Try pressing the "Run" button a second time without closing GUI opened from an earlier "Run".  Expect the previously opened GUI is closed.
- Try switching examples without closing GUI opened from the previous example. Expect the previously opened GUI is closed.
- Try an example that refers to local image resources (e.g. this [example](https://github.com/enthought/traitsui/blob/7.1.0/traitsui/examples/demo/Extras/Image_editor_demo.py), or this [example](https://github.com/enthought/pyface/blob/7.1.0/examples/data_view/column_example.py)) 
- Try deliberately mess up the example locally (e.g. introduce syntax error) and run it.
- Try running an example that does not involve GUI (e.g. this [example](https://github.com/enthought/envisage/blob/5.0.0/envisage/examples/demo/MOTD/run.py)).
- Try running two examples (e.g. [this](https://github.com/enthought/pyface/blob/7.1.0/examples/tasks/advanced/run.py), then [this](https://github.com/enthought/pyface/blob/7.1.0/examples/tasks/basic/run.py)) : one that contains a local import which happens to have a name clash with a local import in another. This is the peril of [double import trap](http://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-double-import-trap) which will either have to be accepted or worked around.

### DIFF
--- a/ets-demo/etsdemo/app.py
+++ b/ets-demo/etsdemo/app.py
@@ -11,8 +11,10 @@
 """ A Traits UI demo that borrows heavily from the design of the wxPython demo.
 """
 
+import abc
 import contextlib
 import glob
+import io
 from io import StringIO
 import operator
 import os
@@ -29,10 +31,12 @@ from os.path import (
     splitext,
 )
 import sys
+import subprocess
+import tempfile
 import token
 import tokenize
 import traceback
-
+import types
 
 from configobj import ConfigObj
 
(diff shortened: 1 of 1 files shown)


kitchoi:
This PR has served its purpose for planning. Closing.

PR #1467, opened 2021-01-06
Fix/Update copyright headers in traitsui

rahulporuri:
The recommended source code header policy at the moment is the following - 

```
# Enthought product code
#
# (C) Copyright 2004-2021 Enthought, Inc., Austin, TX
# All rights reserved.
#
# This file and its contents are confidential information and NOT open source.
# Distribution is prohibited.
```

This PR updates the copyright headers for all source code in the traitsui package and in the integrationtests directory. Note that copyright has been updated to start from 2004. This is because I noticed a number of source code modules were written by David C Morill in 2004.

For lack of a better way, this was all done manually. Each source code module in the codebase was manually opened and the copyright header was checked/updated.

**NOTE**: There are number of modules which are Riverbank Copyright headers. For those modules, there exist now two headers - the original header with the Riverbank Copyright and the new Enthought header.

Note that this PR does not update the source code headers in the etsdemo distribution. That will be done in a separate PR. Additionally, flake8 has not been setup at the moment - which will be done in a separate PR.


aaronayres35:
The header in the PR description does not match the changes made in the PR.  Could we update the PR description?

integrationtests/ui/code_editor_test.py
```diff
@@ -1,15 +1,14 @@
...
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
 #
-#  (c) Copyright 2005 by Enthought, Inc.
-#  License: BSD Style.
-#
-#-------------------------------------------------------------------------
+# Thanks for using Enthought open source!
+
+""" Test using a KeyBindings object with the traits Codeditor."""
```
aaronayres35, line 11: Nit: Typo, should be CodeEditor

rahulporuri:
> The header in the PR description does not match the changes made in the PR. Could we update the PR description?

I'm not entirely sure what the mismatch is. Can you elaborate?

The PR fixes/updates the copyright headers in the codebase.

traitsui/qt4/clipboard.py
```diff
@@ -1,3 +1,13 @@
+# (C) Copyright 2008-2021 Enthought, Inc., Austin, TX
```
aaronayres35, line 1: I'm assuming it was intentional that many of these are left as 2008-2021 not 2004-2021?

aaronayres35 (approved):
Changes LGTM

traitsui/qt4/clipboard.py
rahulporuri: Yes, this was intentional. I updated the PR description to reflect this - some of the source code in traitsui has Riverbank Copyright headers. We should not remove those headers - and we can claim copyright to those source code modules from the year after which Riverbank copyright.

Issue #1399, opened 2020-11-16
Doc: Expand Testing TraitsUI Application Section with how exceptions from GUI are handled


kitchoi:
Currently the "Testing TraitsUI Application Section" in the user manual talks little (or none at all) regarding how exceptions from GUI event loops are handled.

It is worth mentioning that where UITester is running the event loop, it also captures the exceptions for them to be reraised to fail a test.

Issue #1394, opened 2020-10-26
QLineEdit.setText() does not work in UI tests


kozmaz87:
Given a situation when one needs to make sure that the Qt widgets work at the end of a system test may want to call setText on the Qt widget now conveniently located under 

    field = ui_tester.find_by_name(editor_app, "t_field")
    field._target.control.setText('hello')

This does not update the gui nor does the inspect return hello afterwards with auto_process_events on. I assume lineedit integration probably works processing key events and focus shifts to trigger events which I am to investigate next but setText does not seem to trigger a value changed event of any kind or so it seems.

UPDATE: PyCharm's debugger seems to cause this. Simple pytest run makes things work. So does a process_cascade_events in debugger context.

aaronayres35:
Hi @kozmaz87, may I ask what the `field._target` object is?  Currently the UI Tester supports some common interactions on some of the more frequently used traitsui editors.  If `field._target` is one of the currently supported editors, I would recommend importing `from traitsui.testing.api import KeyClick, KeySequence` and then replacing the line `field._target.control.setText('hello')` with
```
field.perform(KeySequence("hello"))
field.perform(KeyClick("Enter"))
```

By using the `perform` method on the `field` object (which is an instance of UIWrapper), the event handling logic is handled automatically by the internals of UIWrapper.  

If `field._target` is not one of the currently supported editors, (or it is but that editor does not support a command you want to perform such as `KeySequence`) I would recommend adding the implementation yourself in a Local Registry, and creating a `UITester` object with that registry as an argument for construction.  For details on how to do this, see [here](https://docs.enthought.com/traitsui/traitsui_user_manual/testing.html#add-support-for-performing-a-user-interaction).  This way you could still use the `perform` method and ensure that all the event handling is handled.  Otherwise you may find yourself needing to do things like calling `process_cascade_events` as you mentioned.  

I hope this helps!  Please let us know if you are still having problems

kozmaz87:
I started out by modifying the button click example from the demos. It is a text_editor.SimpleEditor, which is supported and stuff now works. It turned out to be pydev debugger that pycharm uses interfered with event processing. This is not the first time pydev caught me out. It does not particularly play nice with the type of 'multithreading' the testing framework seems to be doing under the hood. I can oly theorize pydev and the test framework exploit the same principles and they cancel each other out trying to work around the GIL. normal test run without debugger works just fine. 

[...]

Issue #1418, opened 2020-11-30
Create an interface out of TargetRegistry to support different target types dynamically


kitchoi:
This issue proposes a plan to refactor the current `traitsui.testing.tester.target_registry` in order to support a wider range of UI targets in testing. One immediate use case is to be able to support `isVisible` and `isEnabled` query on any Qt/wx widgets.

This is partly motivated by #1291, and partly motivated by the observation that both `QWidget` and `wx.Window` allow one to query whether a widget is visible / enabled. And a lot of UI objects in TraitsUI hold such widget in the `control` attribute.

Currently the `TargetRegistry` uses static typing, e.g.:
```
TargetRegistry.get_handler(target_class, interaction_class)
```
The `target_class` is a subclass of type. These `get_*` methods are mainly consumed by an instance of `UIWrapper` which holds an instance of the target_class.
 
If instead we change it to accept an instance, e.g.:
```
TargetRegistry.get_handler(target, interaction_class)
```
and then lift these `get_*` methods to an interface (could be implemented as an ABC, and could be kept internal for now until there is an external need for it to be made public):
```
AbstractTargetRegistry.get_handler(target, interaction_class)
```
Then we can use dynamic typing approach for #1291.

The current `TargetRegistry` can be one implementation of this ABC, and we can _add_ another implementation for supporting `isVisible` and `isEnabled` for Qt/wx widgets:

```
class QtControlRegistry:
    def get_handler(self, target, interaction_class):
        if hasattr(target, "control") and isinstance(target.control, QtGui.QWidget):
            # return a callable that returns isEnable() if the interaction_class refers to the IsEnabled query
            # return a callable that returns isVisible() if the interaction_class refers to the query on visibility
```

Note: The `TargetRegistry.get_handler` is only called by `UIWrapper`. The method should be considered internal to the _package_ where both `TargetRegistry` and `UIWrapper` belong. It may be renamed with a leading underscore or the docstring should clarify its intended audience. At the moment, I doubt there is a use case for actually using this method outside of the `traitsui.testing.tester` package.

kitchoi:
Closing via #1490

PR #1482, opened 2021-01-15
Refactor traitsui.testing documentation

kitchoi:
This refactoring is motivated by #1399 and #1394: Instead of squeezing in a discussion of GUI event loop and error handling in the single-page documentation, the testing documentation will benefit from a refactoring to provide better homes for them. 

This PR refactors the documentation following the pattern advocated in https://documentation.divio.com/. The resulting documentation provides obvious space for developers to add new content, and for users to find content.

This PR aims at refactoring the documentation with the current content, and add as little new content as possible. After this PR, we can add content for #1399 and #1394 more easily.

Unlike refactoring code, refactoring technical documentation sometimes require a bit more rewriting, otherwise the text would be awkward to read or seem out-of-place.

~Marking as WIP as I'd like to do a bit more self-checking on grammar (really not my expertise) and spelling (there are tools for it). But the general layout is pretty much here.~ Edited: Done, ready for review.


docs/source/conf.py
```diff
@@ -240,8 +240,9 @@
 
 # Intersphinx configuration
 intersphinx_mapping = {
-    'traits': ('http://docs.enthought.com/traits', None),
     'pyface': ('http://docs.enthought.com/pyface', None),
+    "python": ("https://docs.python.org/3", None),
+    'traits': ('http://docs.enthought.com/traits', None),
```
kitchoi, line 245: This change was motivated by a cross reference to unittest which is then removed, I will revert changes to this file.

docs/source/traitsui_user_manual/testing/discussions/compatibility_pyface_testing.rst
```diff
@@ -0,0 +1,67 @@
+Compatibility with Pyface test utilities
+========================================
+
+|UITester| is intended to be compatible with Pyface's |ModalDialogTester|,
```
aaronayres35, line 4: The `|ModalDialogTester|` is not working as a link to the pyface docs for me in the built documentation.

I ran:
```
python etstool.py install
python etstool.py docs
python -m http.server
```

docs/source/traitsui_user_manual/testing/discussions/compatibility_pyface_testing.rst
```diff
@@ -0,0 +1,67 @@
...
+Using UITester and GuiTestAssistant
+-----------------------------------
+
+|GuiTestAssistant| is a more general tool dealing with GUI processing in tests.
+|UITester|, on the other hand, is a more specific tool for testing GUI
+components managed by TraitsUI. The two can be used together in tests.
+|GuiTestAssistant| has been around before |UITester| is introduced.
+
+Since various methods on |UIWrapper| (such as |UIWrapper.perform| and
+|UIWrapper.inspect|) automatically request GUI events to be processed. Where
```
aaronayres35, line 63: This sentence seems to end too early

docs/source/traitsui_user_manual/testing/discussions/working_of_extensions.rst
```diff
@@ -0,0 +1,40 @@
...
+|UITester|. When |UIWrapper.perform| or |UIWrapper.inspect| is called, the
+first registry that can support the given target and interaction will be used.
+Likewise, |UIWrapper.locate| follows the same rule for the given target and
+location. See :ref:`testing-target-interaction-location` for an explanation on
+these three types of objects.
+
+This is how |UITester| supports testing on TraitsUI editor: It always extends
+the given list of registries with more built-in registries that know how to
+interact with TraitsUI editors, and let |UIWrapper| do most of the work.
+
```
aaronayres35, line 39: We could also mention that this allows for overriding the default behavior.  This is implied by the content here but we may want to be explicit?  
Or we very well may not.  Overriding default behavior may be something we want to discourage.  I am not sure.  e.g. I can't think of a direct use case, but it is possible a user may have a different idea of what `DisplayedText()` should be or something?

docs/source/traitsui_user_manual/testing/discussions/working_of_extensions.rst
```diff
@@ -0,0 +1,40 @@
...
+This list of registries is passed on to all |UIWrapper| created from the
+|UITester|. When |UIWrapper.perform| or |UIWrapper.inspect| is called, the
+first registry that can support the given target and interaction will be used.
+Likewise, |UIWrapper.locate| follows the same rule for the given target and
+location. See :ref:`testing-target-interaction-location` for an explanation on
+these three types of objects.
+
+This is how |UITester| supports testing on TraitsUI editor: It always extends
+the given list of registries with more built-in registries that know how to
```
aaronayres35, line 37: Is this strictly speaking correct?  I believe for TraitsUI editor implementations provided by default they are all provided by the single default registry.  So wee are not extending the list of registries for each editor we support, we are adding to the things we register on the default registry.  Although both would work fine

docs/source/traitsui_user_manual/testing/references/examples.rst
```diff
@@ -0,0 +1,21 @@
...
+:ref:`TraitsUI\'s own demos<traitsui-demo>`.
+
+Editors
+-------
+- :github-demo:`BooleanEditor <Standard_Editors/tests/test_BooleanEditor_simple_demo.py>`
+- :github-demo:`ButtonEditor <Standard_Editors/tests/test_ButtonEditor_simple_demo.py>`
+- :github-demo:`CheckListEditor <Standard_Editors/tests/test_CheckListEditor_simple_demo.py>`
+- :github-demo:`EnumEditor <Standard_Editors/tests/test_EnumEditor_demo.py>`
+- :github-demo:`InstanceEditor <Standard_Editors/tests/test_InstanceEditor_demo.py>`
+- :github-demo:`ListEditor <Advanced/tests/test_List_editor_notebook_selection_demo.py>`
```
aaronayres35, line 15: Note there is also https://github.com/enthought/traitsui/blob/master/traitsui/examples/demo/Standard_Editors/tests/test_ListEditor_demo.py 
But the List example you have linked is probably the better example to refer to

docs/source/traitsui_user_manual/testing/substitutions.rst
```diff
@@ -0,0 +1,38 @@
...
+.. |UIWrapper| replace:: :class:`~traitsui.testing.tester.ui_wrapper.UIWrapper`
+.. |UIWrapper.find_by_id| replace:: :func:`~traitsui.testing.tester.ui_wrapper.UIWrapper.find_by_id`
+.. |UIWrapper.find_by_name| replace:: :func:`~traitsui.testing.tester.ui_wrapper.UIWrapper.find_by_name`
+.. |UIWrapper.help| replace:: :func:`~traitsui.testing.tester.ui_wrapper.UIWrapper.help`
+.. |UIWrapper.inspect| replace:: :func:`~traitsui.testing.tester.ui_wrapper.UIWrapper.inspect`
+.. |UIWrapper.locate| replace:: :func:`~traitsui.testing.tester.ui_wrapper.UIWrapper.locate`
+.. |UIWrapper.perform| replace:: :func:`~traitsui.testing.tester.ui_wrapper.UIWrapper.perform`
+.. |UIWrapper._target| replace:: :attr:`~traitsui.testing.tester.ui_wrapper.UIWrapper._target`
+
+.. |ModalDialogTester| replace:: :class:`~pyface.ui.qt4.util.modal_dialog_tester.ModalDialogTester`
```
aaronayres35, line 37: Aha, I am guessing this did not link to docs because it is a toolkit specific class

docs/source/traitsui_user_manual/testing.rst
```diff
@@ -4,656 +4,54 @@
...
-.. |TargetById| replace:: :class:`~traitsui.testing.tester.locator.TargetById`
-.. |TargetByName| replace:: :class:`~traitsui.testing.tester.locator.TargetByName`
+.. toctree::
+   :maxdepth: 1
 
-.. |TargetRegistry| replace:: :class:`~traitsui.testing.tester.target_registry.TargetRegistry`
-.. |TargetRegistry.register_interaction| replace:: :func:`~traitsui.testing.tester.target_registry.TargetRegistry.register_interaction`
-.. |TargetRegistry.register_location| replace:: :class:`~traitsui.testing.tester.target_registry.TargetRegistry.register_location`
+   testing/references/examples
+   API references <../api/traitsui.testing>
```
aaronayres35, line 55: nit: I believe in other Ets documentation we just say "API Reference" rather than references

aaronayres35 (approved):
Mostly LGTM, just a few comments and nit picks

docs/source/traitsui_user_manual/testing/discussions/compatibility_pyface_testing.rst
kitchoi: Thank you for checking, and sorry I should have mentioned that this is expected. Pyface documentation does not include toolkit specific objects.

I kept the cross reference here despite the missing link because there is meant to be a cross reference.  
When that pyface issue is fixed (if ever), the hope is that the cross link here should work (when we generate the documentation again).

docs/source/traitsui_user_manual/testing/discussions/working_of_extensions.rst
kitchoi: I was future-proofing this sentence. The UITester can add more registries than just one. I did not think it would be interpreted as one-registry-per-editor though.

(Motivation, for https://github.com/enthought/traitsui/issues/1418, I was thinking the `_get_default_registry` function will be changed to return a list, rather than just one registry.)

Any suggestion to reword this? Otherwise I might be keeping this as is...

docs/source/traitsui_user_manual/testing/discussions/working_of_extensions.rst
kitchoi: Yeah I agree there is an obvious use case and so we may not want to encourage it.  In the original documentation, there was a section about overriding the default behaviour, but really the motivation there was to explain the machinery about the registry priority.  Since that is reworded here, I think I was going with "this is implied here" so if someone really really wanted to override the default behaviour.

docs/source/traitsui_user_manual/testing/howtos/add_new_location.rst
```diff
@@ -0,0 +1,88 @@
...
+
+Suppose we have a custom UI editor that contains some buttons. The objective of
+a test is to click a specific button with a given label. We will therefore need
+to locate the button with the given label before a mouse click can be
+performed.
+
+The test code we wanted to achieve looks like this::
+
+    container = UITester().find_by_id(ui, "some_container")
+    button_wrapper = container.locate(NamedButton("OK"))
```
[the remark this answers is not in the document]
kitchoi: Good point!

docs/source/traitsui_user_manual/testing/references/examples.rst
kitchoi: I did not notice that. Thank you. I think they are both good and should be included. The standard one probably should go first.

docs/source/traitsui_user_manual/testing/substitutions.rst
kitchoi: Yep.

docs/source/traitsui_user_manual/testing/discussions/compatibility_pyface_testing.rst
kitchoi: Thank you for spotting this.

docs/source/traitsui_user_manual/testing/discussions/working_of_extensions.rst
aaronayres35: That makes sense to me.  This is good as is in that case 👍

docs/source/traitsui_user_manual/testing/discussions/working_of_extensions.rst
aaronayres35: maybe just remove the word "always"? But given the potential change coming from addressing 1418 I am perfectly happy to leave this as it is here.

kitchoi: Removing 'always' sounds good to me!

docs/source/traitsui_user_manual/testing.rst
```diff
@@ -52,6 +52,6 @@ Technical References
    :maxdepth: 1
 
    testing/references/examples
-   API references <../api/traitsui.testing>
+   API References <../api/traitsui.testing>
```
aaronayres35, line 55: Ah I meant make it singular ("Reference") as I think that is what traits etc. call it.  However, tbh I'm not really sure what is grammatically correct / sounds better, it was simply a comment on consistency

aaronayres35 (approved):
Still LGTM, just the one lingering nit about Reference vs References.

docs/source/traitsui_user_manual/testing.rst
kitchoi: Oops, I overlooked that. Yes you are right. Let's go with what is already there.

kitchoi: Argh... The "Technical References" should be "Technical Reference" too. Will fix in a separate PR... (very tempted to make it drive-by in the next one)

aaronayres35: Ah my mistake I missed that, I think a drive by is fine

PR #1485, opened 2021-01-18
Add a section in the documentation about debugging GUI tests

### LINKED ISSUE #1394: QLineEdit.setText() does not work in UI tests

kitchoi:
Closes #1394

This PR adds a section in the testing documentation to mention the fact that debugging activities could interfere with GUI tests. The only `traitsui.testing` relevant part is the `delay` parameter. The rest is in fact general to running GUI tests locally along with debugging activities.


docs/source/traitsui_user_manual/testing.rst
```diff
@@ -41,6 +41,7 @@ Discussions
    :maxdepth: 1
 
    testing/discussions/automated_vs_manual_test
+   testing/discussions/debugging-gui-tests-at-runtime
```
kitchoi, line 44: We can discuss whether to have the section as a standalone or merge it with one of the existing ones.

docs/source/traitsui_user_manual/testing/discussions/debugging-gui-tests-at-runtime.rst
```diff
@@ -0,0 +1,15 @@
...
+To debug an issue with a GUI test, one can use the |UITester.delay| parameter
+to slow down the test, or they may use a Python debugger to inspect internal
+states while the test is being run. In both cases, developers should
+be aware that the GUI can react to any additional events that may be caused by
+these debugging activities, and this may cause the test to behave differently
+compared to running it in normal conditions.
+
+For example, if a test relies on a GUI widget having the keyboard input focus,
+a Python debugger may interfere with the test by stealing focus from the GUI
+when a break point occurs.
```
aaronayres35, line 13: Maybe add another example specific to when the delay is used? Something like "Additionally, if a test utilizes UITester.delay, the UI under test will respond to interactions by the user, e.g. a mouse click while the test is running.

This may be overkill though so feel free to ignore!

docs/source/traitsui_user_manual/testing.rst
aaronayres35: I am a bit torn on this.  I do like having things separated out into sections with a clear purpose, but when the content on individual pages is very small and that content may fit well elsewhere it can feel a bit like overkill.  
In this case the only section that this content may fit within aside from staying in its own section is event_loop_and_exceptions as debugging methods are meddling with the event loop.  

At this point though, I think I am in favor of leaving it as is, on its own page.

aaronayres35 (approved):
LGTM!

A couple comments, but feel free to ignore

docs/source/traitsui_user_manual/testing/discussions/debugging-gui-tests-at-runtime.rst
kitchoi: I am leaning towards thinking this may be overkill at the moment. We can add more examples later though if it turns out the explanation was not clear enough (or maybe fix the explanation then).

PR #1486, opened 2021-01-19
Fix minor errors in testing docs

rahulporuri:
This PR fixes two minor errors introduced in #1482 : 

- Fix a minor coding error in a UITester example in the tutorial subsection
- Remove reference to non-existent UITester.auto_process_events attribute.

### DIFF
--- a/docs/source/traitsui_user_manual/testing/substitutions.rst
+++ b/docs/source/traitsui_user_manual/testing/substitutions.rst
@@ -20,6 +20,7 @@
 .. |TargetRegistry.register_location| replace:: :class:`~traitsui.testing.tester.target_registry.TargetRegistry.register_location`
 
 .. |UITester| replace:: :class:`~traitsui.testing.tester.ui_tester.UITester`
+.. |UITester.auto_process_events| replace:: :attr:`~traitsui.testing.tester.ui_tester.UITester.auto_process_events`
 .. |UITester.create_ui| replace:: :func:`~traitsui.testing.tester.ui_tester.UITester.create_ui`
 .. |UITester.delay| replace:: :attr:`~traitsui.testing.tester.ui_tester.UITester.delay`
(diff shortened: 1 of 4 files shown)


docs/source/traitsui_user_manual/testing/tutorials/test_with_nested_object.rst
```diff
@@ -21,7 +21,7 @@ Suppose we have a GUI application like this::
         pay_check = Instance(PayCheck)
 
         def _pay_check_default(self):
-            return PayCheck(pay=self.min_hourly_wage, n_hours=self.n_hours)
```
kitchoi, line 24: Thank you. As I was writing this, I wanted to bring in doctest, for exactly something like this, but thought the setup would be too much to review. I ran all the examples, and found this exact bug, but what I neglected to do was to copy the fix back from my local script to the documentation.

docs/source/traitsui_user_manual/testing/substitutions.rst
```diff
@@ -20,7 +20,6 @@
 .. |TargetRegistry.register_location| replace:: :class:`~traitsui.testing.tester.target_registry.TargetRegistry.register_location`
 
 .. |UITester| replace:: :class:`~traitsui.testing.tester.ui_tester.UITester`
-.. |UITester.auto_process_events| replace:: :attr:`~traitsui.testing.tester.ui_tester.UITester.auto_process_events`
```
kitchoi, line 23: Please leave this alone, we can add `auto_process_events` in the Attributes section in `UITester`. https://github.com/enthought/traitsui/blob/7a960c68d65bfd2b76a857ef6c4c0deffa256550/traitsui/testing/tester/ui_tester.py#L48-L49

That is why the link does not work. And on the other hand, this is a bit subjective, I think the bolded font for an attribute that does not have a cross reference still looks nicer than the one with double backticks.

rahulporuri: reverted

kitchoi (approved):
Thank you for the fix!

PR #1487, opened 2021-01-19
Better document auto_process_events

kitchoi:
Following up observation from refactoring the documentation: The `auto_process_events` flag is only exposed from the object `__init__` method, making it difficult to cross reference in sphinx (there exists sphinx extensions for that, but we may not want to depend on them).

The flag is an internal state of `UITester` not allowed to be mutated post instantiation. The immutability is important in my opinion because if it was allowed to mutate, one may expect this value to be observed by `UIWrapper` created prior to the mutation. Any attempt to "fix" that is going to add complexity. It is not necessary to mutate the flag, because one can simply create another `UITester` in the same test.

By exposing the attribute as a readonly property, this allows space in the API documentation to describe and expand on these details (this is documentation-driven development here).

### DIFF
--- a/traitsui/testing/tester/tests/test_ui_tester.py
+++ b/traitsui/testing/tester/tests/test_ui_tester.py
@@ -210,3 +210,15 @@ class TestUITesterFindEditor(unittest.TestCase):
         with tester.create_ui(Order(), dict(view=view)) as ui:
             tester.find_by_id(ui, "item")
             self.assertEqual(side_effect.call_count, 0)
+
+
+class TestUITesterGuiFree(unittest.TestCase):
+    """ Test GUI free interface on UITester."""
+
+    def test_auto_process_events_readonly(self):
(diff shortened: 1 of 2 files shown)


rahulporuri (approved):
LGTM.

rahulporuri:
On a tangential note, I see the following sphinx warning when I try building docs (after doing a git clean in the directory) - 

```
C:\Users\rporuri\work\github\ets\traitsui\docs\source\traitsui_dev_guide\testing.rst:127: WARNING: undefined label: advanced-testing (if the link has no caption the label must precede a section header)
```

I'm not entirely sure when this was introduced - or when this broke.

kitchoi:
> On a tangential note, I see the following sphinx warning when I try building docs (after doing a git clean in the directory) -

#1482 removed the "advanced-testing" reference. Looks like there is still one reference left behind.

PR #1489, opened 2021-01-19
Fix link label to the documentation home page

### LINKED ISSUE #1484: Buttons on User Manual have wrong labels
TraitsUI appears to have the same issue ass traits (https://github.com/enthought/traits/issues/1373) in that on the user manual the two blue buttons at the top both say user manual, even though the first links to the documentation home page, and the second to the user manual home page.

kitchoi:
Closes #1484

This is before, these buttons are shown at the very top of the page. The first button actually links to the home page. :
Screenshot 2021-01-19 at 11 05 13

This is after this PR:
Screenshot 2021-01-19 at 11 06 40

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -111,7 +111,7 @@ pygments_style = 'sphinx'
 
 # The name for this set of Sphinx documents.  If None, it defaults to
 # "<project> v<release> documentation".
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM.

Issue #1291, opened 2020-09-29
Add 'IsVisible` query class for UI Tester


aaronayres35:
In some cases, an application/ui may have certain parts of the UI change between visible and not visible depending on the state of the UI.  It is likely that this behavior is something that one would want to verify with tests.

Note that all traitsui editors have a `visible` trait, but also toolkit specific elements (e.g. QtGui.Qwidgets) have `isVisible` methods.

kitchoi:
The use case in tests might have meant to use Qt's `~isHidden` rather than `isVisible`. See the distinction [here](https://doc.qt.io/qt-5/qwidget.html#isHidden).
Wx has this `IsShownOnScreen` method, but I am not sure if it matches Qt's `~isHidden` or `IsVisible`.

kitchoi:
> The use case in tests might have meant to use Qt's ~isHidden rather than isVisible. See the distinction here.
Wx has this IsShownOnScreen method, but I am not sure if it matches Qt's ~isHidden or IsVisible.

I am taking this back :) 
From the testing perspective, one wants to inspect if a widget or a container is visible to the user, and Qt `IsVisible` is the right thing to use: A widget can be invisible because one of its container widget is invisible.

For example:
```
Group(
    Item(..., visible_when="bool_item_1"),
    Item(..., visible_when="bool_item_2"),
    visible_when="bool_group",
)
```
A query of `IsVisible` should return false if `bool_item_1` is evaluated as true but `bool_group` is evaluated as false, because that is what a test would want to check: Check what the user can actually see.

(Coming back to this issue because I am seeing more use cases for this feature. See #1490 for making an implementation will allow `IsVisible` query to be available to a more general set of GUI widget.)

PR #1490, opened 2021-01-19
Extract an interface from TargetRegistry to allow wider testing support for IsEnabled

kitchoi:
This is a draft PR to show implementations and API that would solve #1418.

Narrow goal:
- Allow alternative registries to be written in order to support testing `IsEnabled` check on any QWidget or wx.Window

Wider goal:
- Allow more flexibility in extending of the testing API

Changes:
- All the `get_*` methods on the current `TargetRegistry` are extracted to an interface, `AbstractTargetRegistry`. Any objects that implement the interface can be a registry given to `UIWrapper`.
- Some of the arguments in the `get_*` methods that used to require the type of `UIWrapper._target` is now changed to just an instance. While this adds requirements on the information required by a registry, `UIWrapper._target` is always an instance so that information is almost always available.
- A `DynamicTargetRegistry` is added. Both the existing `TargetRegistry` and the new `DynamicTargetRegistry` implement the interface.
- `UITester` now extends the given custom registries with more than one built-in registries. One of the those built-in registries is a `DynamicTargetRegistry`. On Qt, such registry supports querying for `QWidget.isEnabled()`. On wx, such registry supports `wx.Window.IsEnabled()`. With that `get_default_registry` is changed to `get_default_registries` to return a list instead of a single registry.
- The specific implementation of `IsEnabled` for `ButtonEditor` is replaced by the more general registry.

Note:
- As noted in #1418, the change in parameter types on `TargetRegistry.get_*` is strictly speaking a change to the public API. However I believe these methods are only called by `UIWrapper` and is hardly useful otherwise, so it is unlikely to break downstream code. We should probably rename the methods to mark them as internal though.
- It will be easy to add support for `IsVisible` after the changes in this PR (https://github.com/enthought/traitsui/issues/1291)
- Very early on there was an abstract base class in my draft implementation for the `testing` package... that abstract base class was killed early to defer decisions on generalizing an interface. With a concrete use case, the resulting interface here is different from that old one. I'm glad to have avoided early generalization/abstraction.

With regret, I will be closing this PR (without merging) once it is good enough to be kept for future reference. This was due to my lack of availability and I would not want to keep a PR open unless I know I can commit to follow through with it. I hope that this PR, as a record, helps solving #1418 in the future.

I will be pushing a few more changes before I close it. 
Comments are very welcome any time!

### DIFF
--- a/traitsui/testing/tester/_dynamic_target_registry.py
+++ b/traitsui/testing/tester/_dynamic_target_registry.py
@@ -39,7 +39,28 @@ class DynamicTargetRegistry(AbstractTargetRegistry):
         self.interaction_to_handler = interaction_to_handler
 
     def _get_handler(self, target, interaction):
-        """ Reimplemented AbstractTargetRegistry._get_handler """
+        """ Return a callable for handling an interaction for a given target.
+
+        This is an implementation for an abstract method.
+
+        Parameters
+        ----------
+        target : any
+            The UI target being operated on.
+        interaction : any
+            Any interaction object.
+
+        Returns
+        -------
+        handler : callable(UIWrapper, interaction) -> any
+            The function to handle the given interaction on a target.
+
+        Raises
+        ------
+        InteractionNotSupported
+            If the given target and interaction types are not supported
+            by this registry.
+        """
         if interaction.__class__ not in self._get_interactions(target):
             raise InteractionNotSupported(
                 target_class=target.__class__,
@@ -49,13 +70,46 @@ class DynamicTargetRegistry(AbstractTargetRegistry):
         return self.interaction_to_handler[interaction.__class__]
 
     def _get_interactions(self, target):
-        """ Reimplemented AbstractTargetRegistry._get_interactions """
+        """ Returns all the interactions supported for the given target.
+
+        This is an implementation for an abstract method.
+
+        Parameters
+        ----------
+        target : any
+            The UI target for which supported interactions are queried.
+
+        Returns
+        -------
+        interaction_classes : set
+            Supported interaction types for the given target type.
+        """
         if self.can_support(target):
             return set(self.interaction_to_handler)
         return set()
 
     def _get_interaction_doc(self, target, interaction_class):
-        """ Reimplemented AbstractTargetRegistry._get_interaction_doc """
+        """ Return the documentation for the given target and interaction type.
+
+        This is an implementation for an abstract method.
+
+        Parameters
+        ----------
+        target : any
+            The UI target for which the interaction will be applied.
+        interaction_class : subclass of type
+            Any class.
+
+        Returns
+        -------
+        doc : str
+
+        Raises
+        ------
+        InteractionNotSupported
+            If the given target and interaction types are not supported
+            by this registry.
+        """
         if interaction_class not in self._get_interactions(target):
             raise InteractionNotSupported(
                 target_class=target.__class__,
@@ -65,7 +119,23 @@ class DynamicTargetRegistry(AbstractTargetRegistry):
         return inspect.getdoc(interaction_class)
 
     def _get_solver(self, target, location):
-        """ Reimplemented AbstractTargetRegistry._get_solver """
+        """ Return a callable registered for resolving a location for the
+        given target and location.
+
+        This is an implementation for an abstract method.
+
+        Parameters
+        ----------
+        target : any
+            The UI target being operated on.
+        location : subclass of type
+            The location to be resolved on the target.
+
+        Raises
+        ------
+        LocationNotSupported
+            If the given locator and target types are not supported.
+        """
         raise LocationNotSupported(
             target_class=target.__class__,
             locator_class=location.__class__,
@@ -73,11 +143,43 @@ class DynamicTargetRegistry(AbstractTargetRegistry):
         )
 
     def _get_locations(self, target):
-        """ Reimplemented AbstractTargetRegistry._get_locations """
+        """ Returns all the location types supported for the given target.
+
+        This is an implementation for an abstract method.
+
+        Parameters
+        ----------
+        target : any
+            The UI target for which supported location types are queried.
+
+        Returns
+        -------
+        locators_classes : set
+            Supported locator types for the given target type.
+        """
         return set()
 
     def _get_location_doc(self, target, locator_class):
-        """ Reimplemented AbstractTargetRegistry._get_location_doc """
+        """ Return the documentation for the given target and locator type.
+
+        This is an implementation for an abstract method.
+
+        Parameters
+        ----------
+        target : any
+            The UI target being operated on.
+        locator_class : subclass of type
+            Any class.
+
+        Returns
+        -------
+        doc : str
+
+        Raises
+        ------
+        LocationNotSupported
+            If the given locator and target types are not supported.
+        """
         raise LocationNotSupported(
             target_class=target.__class__,
             locator_class=locator_class,
--- a/traitsui/testing/tester/_dynamic_target_registry.py
+++ b/traitsui/testing/tester/_dynamic_target_registry.py
@@ -12,31 +12,21 @@ from traitsui.testing.tester.exceptions import (
 )
 
 
-class BaseDynamicRegistry(AbstractTargetRegistry):
+class DynamicTargetRegistry(AbstractTargetRegistry):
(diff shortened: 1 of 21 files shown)


kitchoi:
> As noted in #1418, the change in parameter types on `TargetRegistry.get_*` is strictly speaking a change to the public API. However I believe these methods are only called by `UIWrapper` and is hardly useful otherwise, so it is unlikely to break downstream code. We should probably rename the methods to mark them as internal though.

That renaming is done now - since the argument type change is a backward incompatible change, we might as well. If the old methods are ever depended on externally (I doubt it), it is easy to put them up again as a quick fix. 

This PR is now clean to the standard as if this was aiming to be merged. Since (1) this is not a small PR, (2) I probably wouldn't have the bandwidth to follow it through, and (3) I don't want to rush it, I will close this PR later today (I will remove the branch too, but I believe they can be restored later from the PR.)

Comments are very welcome any time (before or after me closing the PR).

traitsui/testing/tester/target_registry.py
```diff
@@ -288,7 +292,7 @@ def get_location_doc(self, target_class, locator_class):
             If the given locator and target types are not supported.
         """
         self._location_registry.get_value(
-            target_class=target_class,
+            target_class=target.__class__,
             key=locator_class,
         )
         # This maybe configurable in the future via register_location
```
aaronayres35, line 298: 😄 
These changes all make a lot of sense to me.  The only methods we expect users to call are really `register_interaction` and `register_location`, for which the api has remained unchanged.  
Everything else (ie. all the `get_*` methods), I agree, are likely to not have been used externally, and really shouldn't be.

It may be slightly confusing (or surprising I guess?) that these other `TargetRegistry` methods now typically take a `target` itself rather than a `target_class` as an argument, but `register_interaction` and `register_location` still take a `target_class`, however I think that is perfectly okay.  It makes sense that registering an interaction or location is a general act, that applies to all instances of a particular class.

traitsui/testing/tester/_dynamic_target_registry.py
```diff
@@ -0,0 +1,85 @@
...
+        """ Reimplemented AbstractTargetRegistry._get_locations """
+        return set()
+
+    def _get_location_doc(self, target, locator_class):
+        """ Reimplemented AbstractTargetRegistry._get_location_doc """
+        raise LocationNotSupported(
+            target_class=target.__class__,
+            locator_class=locator_class,
+            supported=[],
+        )
```
aaronayres35, line 85: Not for this PR, but will we want to add a section to the documentation about `DynamicTargetRegsitry`? This is in a privately named module so I assume you were thinking not / we don't expect users to mess with it.

As far as users wanting to extend UITester functionality, I feel like I could imagine scenarios where it would be useful to work with an additional `DynamicTargetRegistry` rather than a `TargetRegistry` to append to the default registries already provided on a UITester.  For example right now, if this went in and say hypothetically we never put in an `IsVisible` implementation along with this.  A downstream user may want to do something like that? 
However, I can also see an argument for this only being used internally, and keeping the means of extending UITester functionality exactly as it is now.

traitsui/testing/tester/_dynamic_target_registry.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def _get_handler(self, target, interaction):
+        """ Reimplemented AbstractTargetRegistry._get_handler """
+        if interaction.__class__ not in self._get_interactions(target):
+            raise InteractionNotSupported(
+                target_class=target.__class__,
+                interaction_class=interaction.__class__,
+                supported=list(self._get_interactions(target)),
+            )
+        return self.interaction_to_handler[interaction.__class__]
```
aaronayres35, line 49: I understand the switch from `target_class` to `target`, but does `interaction_class` need to be `interaction` here (we only use `interaction.__class__`?  or is that just to be consistent?
e.g. in `_get_interaction_doc` we use the `interaction_class`

I don't think there is any problem with having it be `interaction`.  Just want to check I am not missing anything

aaronayres35 (commented):
LGTM

I have a few comments / questions.  I apologize for the extremely long delay in my review. @kitchoi if you do not currently have the bandwidth to go through with this PR, I am happy to talk with Rahul about best next steps to take for how we can get this pushed through.  I think it will provide a very nice improvement for the `UITester` 😃, thank you!

traitsui/testing/tester/_dynamic_target_registry.py
```diff
@@ -0,0 +1,85 @@
...
+from traitsui.testing.tester._abstract_target_registry import (
+    AbstractTargetRegistry,
+)
+from traitsui.testing.tester.exceptions import (
+    InteractionNotSupported,
+    LocationNotSupported,
+)
+
+
+class DynamicTargetRegistry(AbstractTargetRegistry):
```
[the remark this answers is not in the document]
kitchoi: Naming things is hard indeed. I see your point, the word "dynamic" could well be used to describe 'TargetRegistry', in a different way. I couldn't think of a better name at the moment. Suggestions welcome!

traitsui/testing/tester/_dynamic_target_registry.py
kitchoi: I think I started out keeping this internal, but I am totally open to exposing this publicly if it is useful downstream. With this made public, then we need to think harder about the name...

traitsui/testing/tester/tests/test_registry.py
```diff
@@ -52,9 +59,42 @@ def handler(wrapper, interaction):
...
+        registry.register_interaction(
+            target_class=SpecificEditor,
+            interaction_class=UserAction2,
+            handler=handler,
+        )
+
+        # then
+        self.assertEqual(
+            registry._get_interactions(SpecificEditor()),
+            set([UserAction, UserAction2])
```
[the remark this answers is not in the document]
kitchoi: Good point. `{UserAction, UserAction2}` is a perfectly good replacement here.

traitsui/testing/tester/tests/test_ui_wrapper.py
```diff
@@ -158,47 +185,47 @@ class TestUIWrapperLocationRegistry(unittest.TestCase):
...
+        wrapper = wrapper.locate("some string")
 
         self.assertEqual(wrapper._target, 2)
 
         # swap the order
         wrapper = example_ui_wrapper(
             registries=[registry1, registry2],
         )
-        wrapper = wrapper.locate(None)
+        wrapper = wrapper.locate("some other string")
```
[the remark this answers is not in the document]
kitchoi: Yes, this makes sense.

kitchoi (commented):
Thank you @aaronayres35 for the review.

> if you do not currently have the bandwidth to go through with this PR, I am happy to talk with Rahul about best next steps to take for how we can get this pushed through.

As much as I'd like to go through with this, I am wary of holding things up because of the lack of bandwidth. Though I do feel a responsibility here. Should I aim at addressing the review comments by the end of next week, and if I fail (not surprising, but I will aim not to), others can take over from then on?

aaronayres35:
> As much as I'd like to go through with this, I am wary of holding things up because of the lack of bandwidth. Though I do feel a responsibility here. Should I aim at addressing the review comments by the end of next week, and if I fail (not surprising, but I will aim not to), others can take over from then on?

That sounds perfectly reasonable to me.  Again, if you don't have the bandwidth, don't worry about it at all.  I am more than happy to take up this PR / work with Rahul to get it pushed through

kitchoi:
@aaronayres35 I hope I have addressed all the comments. I have also updated the class docstring of `DynamicTargetRegistry`, please take a look. As for the name of the registry, I couldn't think of an alternative name unfortunately. With that, I'd leave the class in an internal module so that its name and relevant documentation can be better hashed out as separate PRs before exposing the class as part of the public API.

aaronayres35:
> @aaronayres35 I hope I have addressed all the comments. I have also updated the class docstring of `DynamicTargetRegistry`, please take a look. 

Still LGTM, thank you for the detailed doc strings - they are very helpful! 

>As for the name of the registry, I couldn't think of an alternative name unfortunately. With that, I'd leave the class in an internal module so that its name and relevant documentation can be better hashed out as separate PRs before exposing the class as part of the public API.

 That makes sense, I agree

aaronayres35 (approved):

kitchoi:
Thank you.

PR #1492, opened 2021-01-20
Add GitHub Actions to test against EDM

kitchoi:
This PR contributes towards #1416

- Add GitHub Actions workflow to run tests against EDM, run it per PR.
- Change the workflow against PyPI dependencies to be run weekly 
- Skip testing the demo example that depends on pytables, on OSX. 
     - The issue that caused us to move OSX build to Appveyor (https://github.com/enthought/traitsui/pull/1160) is valid for GitHub Actions. First of all, our integration tests include a test for the demo examples (it has been useful and has caught regressions!), but one of the examples depend on pytables, whose build from EDM requires a host CPU that supports AVX 2, and the OSX host from GitHub Actions does not support that. 

Observations:
- If the workflow Ubuntu version is upgraded to 18.04, then we will need to `sudo apt-get install ` some additional libraries if we want to use the prebuilt wheel from wxPython.org, and `libsdl2-2.0-0` is either the wrong one or not enough (it works for the wheel for Ubuntu 16 though). But since TraitsUI test suite against wx fail on all platforms, and GitHub Actions does not support allow-failure, I did not investigate further.

Apart from the lack of workflow against `ets-demo`, this retains what we currently test on Travis/Appveyor.  Note that this GitHub Actions workflow excludes wx. We do test against wx on Travis, but the job is allowed to fail such that when its failure mode changes, it is often overlooked.

### DIFF
--- a/.github/workflows/test-with-edm.yml
+++ b/.github/workflows/test-with-edm.yml
@@ -32,8 +32,6 @@ jobs:
           sudo apt-get update
           sudo apt-get install libsdl2-2.0-0
         if: matrix.toolkit == 'wx'
-      - name: Install GL dependencies for Linux
-        run: sudo apt-get install libglu1-mesa-dev
       - name: Cache EDM packages
         uses: actions/cache@v2
         with:
--- a/.github/workflows/test-with-edm.yml
+++ b/.github/workflows/test-with-edm.yml
@@ -49,7 +49,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, windows-latest]
-        toolkit: ['pyqt5', 'pyside2', 'wx']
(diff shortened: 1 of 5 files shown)


kitchoi:
For reference, the move to build against EDM packages for PR builds follows an offline discussion. We don't want to exploit free resources too much, and compatibility against dependencies from EDM is important. Dependencies from PyPI are important too as it allows to stay current and test against latest packages/Python versions, so that is still kept in cron.

rahulporuri (approved):
LGTM.

With this PR, looks like we can get rid of Travis/Appveyor CI entirely. And we're also testing against EDM for all PRs but we have cron jobs for everything else.

PR #1495, opened 2021-01-21
Add instance choice classes to traitsui.api

### LINKED ISSUE #1494: InstanceChoice, InstanceChoiceItem, and InstanceFactory should be added to traitsui.api
These classes are part of the _secret_ `InstanceEditor` api, but they should not be so secret!  They are expected to need to be used along with the `InstanceEditor` for certain use cases and thus should be exposed in the api module.

aaronayres35:
closes #1494 

This PR simply adds instance choice related classes to the `traitsui.api`

### DIFF
--- a/traitsui/api.py
+++ b/traitsui/api.py
@@ -159,6 +159,14 @@ Table filter types:
 - :class:`~.RuleTableFilter`
 - :class:`~.TableFilter`
 
+Instance UI
+-----------
+
+- :class:`~.InstanceChoice`
+- :class:`~.InstanceChoiceItem`
+- :class:`~.InstanceDropChoice`
+- :class:`~.InstanceFactoryChoice`
+
 Toolkit Object
 --------------
 
@@ -378,6 +386,13 @@ from .table_filter import (
 )
 
 from .tabular_adapter import TabularAdapter
+
+from .instance_choice import (
+    InstanceChoice,
+    InstanceChoiceItem,
+    InstanceDropChoice,
+    InstanceFactoryChoice,
+)
 
 from .toolkit import toolkit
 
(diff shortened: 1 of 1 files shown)


traitsui/api.py
```diff
@@ -159,6 +159,14 @@
...
 - :class:`~.TableFilter`
 
+Instance UI
+-----------
+
+- :class:`~.InstanceChoice`
+- :class:`~.InstanceChoiceItem`
+- :class:`~.InstanceDropChoice`
+- :class:`~.InstanceFactoryChoice`
+
```
rahulporuri, line 169: Let's move the `InstanceEditor` down here, it'll be cleaner.

rahulporuri (approved):
LGTM

traitsui/api.py
aaronayres35: There are comparable sections for "Tree UI" and "Table UI", but the `TreeEditor` and `TableEditor` are still included in the full list with the other editors not in those sections.  I am still happy to make this suggested change, but wanted to point this out from a consistency stand point

rahulporuri: In that case, let's move the `TreeEditor` and `TableEditor` to the Tree UI and Table UI sections respectively - along with the change I recommended earlier.

Alternatively, if we want to list all of the available editors in one location i.e. under the Editors section (how I think things are currently), we can choose to add duplicate information in the Table UI, Tree UI and Instance UI sections.

rahulporuri (approved):
Still LGTM. Can you check if there are any new sphinx warnings generated because of these changes when the documentation is built?

aaronayres35:
> Still LGTM. Can you check if there are any new sphinx warnings generated because of these changes when the documentation is built?

Just checked and I'm not seeing any new warnings from these changes

PR #1496, opened 2021-01-21
Add GitHub Actions workflow for testing ets-demo

### LINKED ISSUE #1416: Move CI to GitHub Actions

kitchoi:
This PR adds a workflow to run the tests for ets-demo, replicating what we have on Travis.

Note:
- ~I added filters to skip builds for TraitsUI distribution if the only changed files are in `ets-demo`.~ (Edited: Did not seem to have an effect, perhaps the fact that the jobs are required for merging means they must be run?)
- Likewise, the build for testing ets-demo is only triggered if any files in `ets-demo` are changed. I made one dummy change to test this.
- The existing style check workflow already flake8 the entire source checkout, so that includes both `traitsui` and `ets-demo`.

Observation: Apparently Ubuntu 18 + EDM + PySide2-from-PyPI does not work. It is the usual complaint about not being able to load xcb plugin. Falling back to Ubuntu 16 works. This is strange because we have used Ubuntu 18 + Python (from GitHub Actions) + PySide2-from-PyPI successfully with the same linux packages.

After this PR, there are strictly two builds missing:

- Tests against wxPython. However the test suite is expected to fail. If we do add it, PR's CI status will always be red even if a successful build for the wxPython job is not required for merging. There is an open issue for GitHub Actions to support allow-failure. On the other hand, a CI job that is expected to fail often gets ignored...
- ~Cron job for ets-demo to test against ETS source: It has been there by virtue of having it mixed in with TraitsUI in the same Travis CI config. Given ets-demo is an application, I am not sure how much we want to keep that cron job.~ (Edited: The cron job is not actually testing etsdemo against ETS source)

If these two builds are not required for the migration to GitHub Actions, then this PR resolves #1416


kitchoi:
On this commit, the jobs for testing etsdemo passed: https://github.com/enthought/traitsui/pull/1496/commits/bc988686333cc16ebc85644edeacab4253c95567
On this commit, no files in `ets-demo` are changed, so the jobs for testing etsdemo are not triggered: https://github.com/enthought/traitsui/pull/1496/commits/92bf519cdfbeaebbf9ed58ef381b386a4dde3bbc

rahulporuri (approved):
LGTM. Let's open a separate issue to resurrect the wxpython job - when github actions CI adds support for expected failures. Thanks for all the awesome work kit!

kitchoi:
Thanks @rahulporuri 
I opened https://github.com/enthought/traitsui/issues/1500 to explain the current state of affair related to wxPython and CI.

PR #1498, opened 2021-01-22
Qt InstanceEditor button should never be the default

### LINKED ISSUE #1475: InstanceEditor button is default in Qt when it should not be
Given a view something like the following:
```
run_action = Action(name='Run', action='do_run')

View(
    Item('transform', editor=InstanceEditor(label='Edit...'))
    # other stuff
    buttons=[run_action],
    default=run_action,
)
```
The button in the simple `InstanceEditor` gets bound as the default (ie. has decoration, is invoked by enter).  If the item is removed, then the run button is the default as expected.

corranwebster:
Fixes #1475.

Not sure how to test this apart from a manual check.

### DIFF
--- a/traitsui/qt4/instance_editor.py
+++ b/traitsui/qt4/instance_editor.py
@@ -422,6 +422,7 @@ class SimpleEditor(CustomEditor):
         """ Creates the editor control (a button).
         """
         self._button = QtGui.QPushButton()
+        self._button.setAutoDefault(False)
         layout.addWidget(self._button)
         self._button.clicked.connect(self.edit_instance)
         # Make sure the editor is properly disposed if parent UI is closed


rahulporuri (commented):
I'm not entirely sure how I can test this PR manually. Is there an example I can run on the main branch vs this branch to compare?

corranwebster:
I don't have a complete minimal example put together, but #1475 has a sketch of the sort of thing that is needed to make the problem occur in the `View`.  I'll see if I have time to flesh that out into a stand-alone example today, which I will add to the issue.

corranwebster:
The following stripped-down example illustrates the issue:
```

from traits.api import Any, Code, HasStrictTraits, Instance, Str
from traitsui.api import (
    Action, InstanceEditor, Item, Menu, MenuBar, VGroup, View
)
from traitsui.instance_choice import InstanceChoiceItem


RunAction = Action(name="Run", action="do_run")

DEFAULT_SCRIPT = '''
def transform(data):
    return None
'''


class Transform(HasStrictTraits):

    title = Str()

    script = Code(DEFAULT_SCRIPT)

    def run(self, data):
        namespace = {}
        exec(self.script, namespace)
        transform = namespace["transform"]
        return transform(data)


class SimpleTransformView(HasStrictTraits):

    transform = Instance(Transform, ())

    data = Any

    result = Any

    def do_run(self):
        self.result = self.transform.run(self.data)

    traits_view = View(
        VGroup(
            Item(
                "transform",
                editor=InstanceEditor(
                    label="Edit...",
                    kind='modal',
                ),
            ),
            Item('data'),
            Item('result')
        ),
        menubar=MenuBar(
            Menu(RunAction, name="Run"),
        ),
        buttons=[
            RunAction,
        ],
        default_button=RunAction,
        resizable=True,
        title="Analysis",
    )

if __name__ == "__main__":
    view = SimpleTransformView()
    view.configure_traits()
```

Without the fix, this produces the following UI (note the "Edit" is blue, indicating that it is the default on OS X):
image

With the fix, we get the following UI (note the "Run" is blue, indicating that it is the default on OS X):
image

rahulporuri (approved):
LGTM. I tested the example - https://github.com/enthought/traitsui/pull/1498#issuecomment-771594780 - against PyQt4, PyQt5 and PySide2 and it works as expected (`Run` is active/enabled instead of  `Edit`).

PR #1499, opened 2021-01-22
More InstanceEditor tests

corranwebster:
This PR adds tests for the various uses of `InstanceEditor` that use a selection combobox.

PR #1503, opened 2021-01-26
Propagate UI errors to the UI's parent

### LINKED ISSUE #1501: Should InstanceEditor propagate its created ui's errors to the parent UI
Basically if you have a view for some `HasTraits` class `A` with an instance editor that displays some nested ui for an instance of object `B`, then errors originating in that nested ui are not counted towards the errors of the ui for the object `A`.  This leads to problems such as an OK button for `A` being enabled even if there is invalid data entered for some of B's traits.

Here is an example:
```
from traits.api import HasTraits, String, Instance
from traitsui.api import  InstanceEditor, Item, View, UCustom
from traitsui.testing.api import (
    IsEnabled, KeyClick, KeySequence, MouseClick, TargetByName, UITester
)


class SomeClass(HasTraits):
    some_string = String("A", maxlen=3)

    traits_view = View(Item('some_string'))


class WrapperClass(HasTraits):
    something = Instance(SomeClass, args=())

    traits_view = View(
        Item(
            'something',
            editor=InstanceEditor(),
            style='custom'
        ),
        buttons=["OK", "Cancel"],
    )


if __name__ == '__main__':
    example = WrapperClass()
    tester = UITester()
    with tester.create_ui(example) as ui:
        something_ui = tester.find_by_name(ui, "something")
        some_string_field = something_ui.locate(TargetByName('some_string'))
        some_string_field.perform(KeySequence("abcd"))
        some_string_field.perform(KeyClick("Enter"))

        ok_button = tester.find_by_id(ui, "OK")
        print("OK button is enabled? :", ok_button.inspect(IsEnabled()))
        print("InstanceEditor UI: ", something_ui._target._ui)
        print("InstanceEditor UI's parent: ", something_ui._target._ui.parent)
        print("Main UI: ", ui)
        print("InstanceEditor UI Errors: ", something_ui._target._ui.errors)
        print("Main UI Errors: ", ui.errors)
```

Running this script gives:
```
OK button is enabled? : True
InstanceEditor UI:  <traitsui.ui.UI object at 0x7f99251b3468>
InstanceEditor UI's parent:  <traitsui.ui.UI object at 0x7f9925075200>
Main UI:  <traitsui.ui.UI object at 0x7f9925075200>
InstanceEditor UI Errors:  1
Main UI Errors:  0
```

Should the nested UI errors contribute to the errors count for the main UI?  Intuitively it seems like they should.  Is this the expected behavior, or is this a bug?

aaronayres35:
fixes #1501 

This is a couple line change which simply adds a listener to a UI's errors trait.  Whenever the errors change, this change is propagated up to the UI's parent's errors trait as well.

BEWARE: although this seems like the logical thing to do it is a change in behavior and we may risk breaking code that was dependent upon the parent UI not picking up on child UI errors for whatever reason.  
The scope of this changes impact should be investigated before merging

NOTE: I merged #1499 into this PR as I found myself wanting to write basically the same thing to write new tests. If changes are made to that PR we will want to pull those in here as well

### DIFF
--- a/traitsui/ui.py
+++ b/traitsui/ui.py
@@ -867,6 +867,10 @@ class UI(HasPrivateTraits):
         if self.parent:
             self.parent.errors = self.parent.errors - old + new
 
+    def _parent_changed(self, name, old, new):
+        old.errors -= self.errors
+        new.errors += self.errors
+
     def _updated_changed(self):
         if self.rebuild is not None:
             toolkit().rebuild_ui(self)
--- a/traitsui/ui.py
+++ b/traitsui/ui.py
@@ -860,6 +860,10 @@ class UI(HasPrivateTraits):
 
     # -- Traits Event Handlers ------------------------------------------------
 
+    def _errors_changed(self, name, old, new):
+        if self.parent:
+            self.parent.errors = self.parent.errors - old + new
+
     def _updated_changed(self):
         if self.rebuild is not None:
             toolkit().rebuild_ui(self)
--- a/traitsui/ui.py
+++ b/traitsui/ui.py
@@ -868,7 +868,8 @@ class UI(HasPrivateTraits):
             self.parent.errors = self.parent.errors - old + new
 
     def _parent_changed(self, name, old, new):
-        old.errors -= self.errors
+        if old:
+            old.errors -= self.errors
         new.errors += self.errors
 
     def _updated_changed(self):
--- a/traitsui/ui.py
+++ b/traitsui/ui.py
@@ -239,6 +239,9 @@ class UI(HasPrivateTraits):
     def dispose(self, result=None, abort=False):
         """ Disposes of the contents of a user interface.
         """
+        if self.parent:
+            self.parent.errors -= self.errors
+
         if result is not None:
             self.result = result
 
--- a/traitsui/ui.py
+++ b/traitsui/ui.py
@@ -239,7 +239,7 @@ class UI(HasPrivateTraits):
     def dispose(self, result=None, abort=False):
         """ Disposes of the contents of a user interface.
         """
-        if self.parent:
+        if self.parent is not None:
             self.parent.errors -= self.errors
 
         if result is not None:
@@ -868,9 +868,10 @@ class UI(HasPrivateTraits):
             self.parent.errors = self.parent.errors - old + new
 
     def _parent_changed(self, name, old, new):
-        if old:
+        if old is not None:
             old.errors -= self.errors
-        new.errors += self.errors
+        if new is not None:
+            new.errors += self.errors
 
     def _updated_changed(self):
         if self.rebuild is not None:
--- a/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/instance_editor.py
+++ b/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/instance_editor.py
@@ -9,8 +9,13 @@
 # Thanks for using Enthought open source!
 
 from traitsui.testing.tester.command import MouseClick
+from traitsui.testing.tester.locator import Index
+from traitsui.testing.tester.query import SelectedText
+from traitsui.testing.tester._ui_tester_registry._common_ui_targets import (
+    BaseSourceWithLocation
+)
 from traitsui.testing.tester._ui_tester_registry.qt4._interaction_helpers import (  # noqa
-    mouse_click_qwidget
+    mouse_click_combobox, mouse_click_qwidget
 )
 from traitsui.testing.tester._ui_tester_registry._traitsui_ui import (
     register_traitsui_ui_solvers,
@@ -41,6 +46,38 @@ def _get_nested_ui_custom(target):
     return target._ui
 
 
+def _get_combobox(target):
+    """ Obtains a nested combobox within an Instance Editor.
+
+    Parameters
+    ----------
+    target : instance of CustomEditor
+    """
+    return target._choice
+
+
+def _click_combobox_index(wrapper, _):
+    return mouse_click_combobox(
+        combobox=_get_combobox(wrapper._target.source),
+        index=wrapper._target.location.index,
+        delay=wrapper.delay,
+    )
+
+
+def _get_combobox_text(wrapper, _):
+    return _get_combobox(wrapper._target).currentText()
+
+
+class _IndexedCustomEditor(BaseSourceWithLocation):
+    """ Wrapper class for CustomEditors with a selection.
+    """
+    source_class = CustomEditor
+    locator_class = Index
+    handlers = [
+        (MouseClick, _click_combobox_index),
+    ]
+
+
 def register(registry):
     """ Register interactions for the given registry.
 
@@ -51,6 +88,8 @@ def register(registry):
     registry : TargetRegistry
         The registry being registered to.
     """
+    _IndexedCustomEditor.register(registry)
+
     registry.register_interaction(
         target_class=SimpleEditor,
         interaction_class=MouseClick,
@@ -59,4 +98,10 @@ def register(registry):
         )
     )
     register_traitsui_ui_solvers(registry, SimpleEditor, _get_nested_ui_simple)
+
+    registry.register_interaction(
+        target_class=CustomEditor,
+        interaction_class=SelectedText,
+        handler=_get_combobox_text,
+    )
     register_traitsui_ui_solvers(registry, CustomEditor, _get_nested_ui_custom)
--- a/traitsui/testing/tester/_ui_tester_registry/wx/_traitsui/instance_editor.py
+++ b/traitsui/testing/tester/_ui_tester_registry/wx/_traitsui/instance_editor.py
@@ -9,10 +9,17 @@
 # Thanks for using Enthought open source!
 
 from traitsui.testing.tester.command import MouseClick
+from traitsui.testing.tester.locator import Index
+from traitsui.testing.tester.query import SelectedText
+from traitsui.testing.tester._ui_tester_registry._common_ui_targets import (
+    BaseSourceWithLocation
+)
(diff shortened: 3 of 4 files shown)


rahulporuri:
This is a very good instance where writing a failing test first would be useful to ensure that the fix works as expected. Can you try putting together a failing test for this behavior?

aaronayres35:
Strange, we are now seeing the seemingly unrelated error after the test was added:
```
  File "D:\a\traitsui\traitsui\.edm\envs\traitsui-test-3.6-pyqt\lib\site-packages\traitsui\testing\tests\test_gui.py", line 154, in test_qt_process_events_process_all

```

rahulporuri (approved):
The test is what I expected - and the fix seems to work as needed - but i'm not sure if this is the right fix or the best fix.

LGTM on my end but it might be good to get a second, more experienced pair of eyes.

corranwebster:
I think that this approach may fail in the situation where the child UI is regenerated in a way that the number of errors is different (eg. InstanceEditor UI is in error state, instance changes, new UI is generated which is not in error state - I think that this will not clear the old errors from the count).

It may be possible to fix by inserting code into the `dispose()` method of the UI, or in the `resynch_editor()` method of the `InstanceEditor`.

An alternative approach, which may be better (or not, I'm in two minds, but this is the way I probably would have started) is to have the `InstanceEditor` track the UI error state and set the error state of the `InstanceEditor` appropriately, and then that would automatically get picked up by the regular error mechanisms.  The downside to this is that you would need to make this change in all toolkits, and possibly in any other editor which embeds another UI (I think `TableEditor` and `TreeEditor` fall into this case at a minimum, possibly others as well).

So first approach may be better _provided_ every possible way that the error count can be accounted for is properly tracked and tested; the second approach is easier to be confident that you are covering all bases, but may be more work.

aaronayres35:
> I think that this approach may fail in the situation where the child UI is regenerated in a way that the number of errors is different (eg. InstanceEditor UI is in error state, instance changes, new UI is generated which is not in error state - I think that this will not clear the old errors from the count).
> 
> It may be possible to fix by inserting code into the `dispose()` method of the UI, or in the `resynch_editor()` method of the `InstanceEditor`.
> 
> An alternative approach, which may be better (or not, I'm in two minds, but this is the way I probably would have started) is to have the `InstanceEditor` track the UI error state and set the error state of the `InstanceEditor` appropriately, and then that would automatically get picked up by the regular error mechanisms. The downside to this is that you would need to make this change in all toolkits, and possibly in any other editor which embeds another UI (I think `TableEditor` and `TreeEditor` fall into this case at a minimum, possibly others as well).
> 
> So first approach may be better _provided_ every possible way that the error count can be changed for is properly tracked and tested; the second approach is easier to be confident that you are covering all bases, but may be more work.

I added a new failing test and a fix by adding to the ui's `dispose` method.  Note the tests are added for the instance editor but as you mentioned this should apply to more general situations such as for `TableEditor` and `TreeEditor`.  I would love to add those tests as well, but those editors are yet to be implemented for UITester, so I held off on that for now.

Also, just to clarify, currently the errors are effectively _never_ appropriately synced up between child and parent UI.  Thus, although we may not be covering all possible scenarios, any scenario we do cover should be a standalone improvement, is that correct?  In other words, at present I am unsure if there are any scenarios we are missing where the error count could be changed.  (I would imagine this currently covers most if not all cases?)  I am also unsure how we would know with certainty that every possible way the error count could be changed is handled.  Should this PR be blocked until that can be confirmed? 

I am also happy to try the other approach you mentioned of tracking the UI error state on the Editors themselves so that it gets picked up by the owning UI.  Like you mentioned that way we could be more confident we had all bases covered.

corranwebster:
Some care needs to be taken here because a partial fix can sometimes be worse than nothing, depending on the remaining failure modes.

In particular, right now because there is no tracking of the inner UI, if there are no errors then the user can _always_ hit the OK button; the problems are always of the form "possibly hitting 'OK' with an invalid input set" but the error is usually flagged visually and the user can usually fix things.  But the key is that this never stops a user from accepting valid input.

If we make a change where the error counting can be off in some situations, then we can possibly end up with a situation where the UI thinks that there is an error but there isn't, and there is no way for the user to fix things (other than perhaps cancelling and starting all over again).  From a user experience point of view, that is _much_ worse.

corranwebster:
One remaining question for this approach - is there anywhere else where we might need to change the parent's error count.  For example, is it possible for a UI object to be created already with errors?  In that case we may need to have a check somewhere in the creation process.  Or there may even be a race condition during initialization.

Also is it possible for the parent object to change during the lifetime of the UI object?  This might include during initialization if the `errors` trait happens to get set before the `parent` trait.

There's something careful thought that needs to happen.  We could guard against the parent changing by having a change handler for the `parent`, for example.

I'm also a little suspicious of the `rebuild` method.  So to be confident that we've covered everything, this needs a careful review of the life-cycle of the UI object to make sure we're not missing something.

aaronayres35:
>There's something careful thought that needs to happen. We could guard against the parent changing by having a change handler for the parent, for example.

I’ve added this in the latest commit.  I have not added an explicit test as I was unsure of a use case where this would occur. Nonetheless it makes sense as a safeguard.  An example use case and corresponding test would be very useful though. 

>For example, is it possible for a UI object to be created already with errors? In that case we may need to have a check somewhere in the creation process. Or there may even be a race condition during initialization.

 To look into this I tried to have a very simple example where I had a class with a trait `some_string = String(“abcd", maxlen=3)` (so the default value is invalid / theoretically starting with an error).  Then I wrapped that object in another class with an Instance trait / editor, and tried to see what happened.  Neither the child ui of the InstanceEditor or the full parent ui had an error.  The text box did not show up red and ok button was enabled.  Simply deleting and retyping the “d” though triggered the error (child/parent UI both have `errors == 1`, field is red, ok button disabled).  So traits defaults are allowed to be invalid correct? I do not know how the UI object could get created with errors.  If the default is invalid we don’t get an error (on the child even so no need to sync up a parent) and if an error were to be introduced programatically before the creation of the UI I believe traits would catch it? e.g I tried doing something like this: 

```
example = ObjectWithValidatedInstance(something=ValidatedEditedInstance(some_string="abcd"))
example.configure_traits()
```
and traits raised an error
```
traits.trait_errors.TraitError: The 'some_string' trait of a ValidatedEditedInstance instance must be a string <= 3 characters long, but a value of 'abcd' <class 'str'> was specified.
```
A search of the codebase for `errors` does not yield very many results.  I don't see anything suggesting a way for the UI to have errors when it is first created, but again, the code base is rather dense so I am not 100% confident I am not missing something...


Also, during my exploration into the `UI` lifecycle I ran into some confusing things.

What is the purpose of the `updated` trait (defined on both the `traitsui.ui.UI` object and the `traitsui.view.View` object).  The trait is defined and the idea behind it makes sense (“Event when the user interface has changed”, “Event when the view has been updated”), but the event never seems to get set.  I put a print statement in the `_updated_changed` method of the `UI` object and ran the full traitsUI test suite only to see it was never called (?).  This seemed very strange to me (not sure if it is old dead code or if I am missing something).   Inside `_updated_changed` is the only place I see `rebuild_ui` (e.g from `traitsui.qt4.toolkit.GUIToolkit` called).  I put a print statement in this method as well and ran the full test suite only to see it was never called.  Inside the GUIToolkit `rebuild_ui` method is the only place I am ever seeing that the `UI` objects `recycle` method is called. Again a print statement and test suite run and this method is never called either!
Looking into the history of the `traitsui/ui.py` file it seems much of this code hasn’t been touched for years so it is possible that this is all dead code, in which case I would like to see it removed.  However it is very possible I am missing something and we don’t want to risk breaking anything.  
Also, why is there an `_rebuild` callable trait?  I don’t see it used or referenced anywhere in the codebase. 

>I'm also a little suspicious of the rebuild method. So to be confident that we've covered everything, this needs a careful review of the life-cycle of the UI object to make sure we're not missing something.

The `rebuild` method gets set up as `getattr(toolkit(), "ui_" + kind)`.  The possibilities (at least on qt) are found here:
https://github.com/enthought/traitsui/blob/a4be4b539be4dfae295525f86ed18f04ad67bc71/traitsui/qt4/toolkit.py#L143-L197
These do things like
Create a `_Panel` object, call `ui.prepare_ui()` and `ui.restore_prefs()`
Or calling `BaseDialog.display_ui()`
_I_ don’t see anything that looks glaringly problematic, but I could well be missing something.  In this code path things like `ui.owner` or `ui.control` will be changed, but I don’t see anything that would trigger a need to resync the UI and parent UI errors.

traitsui/ui.py
```diff
@@ -860,6 +863,15 @@ def _get_key_bindings(self):
...
     # -- Traits Event Handlers ------------------------------------------------
 
+    def _errors_changed(self, name, old, new):
+        if self.parent:
+            self.parent.errors = self.parent.errors - old + new
+
+    def _parent_changed(self, name, old, new):
+        if old:
+            old.errors -= self.errors
+        new.errors += self.errors
```
corranwebster, line 873: I'd suggest:
```suggestion
        if old is not None:
            old.errors -= self.errors
        if new is not None:
            new.errors += self.errors
```

traitsui/ui.py
```diff
@@ -239,6 +239,9 @@ def ui(self, parent, kind):
     def dispose(self, result=None, abort=False):
         """ Disposes of the contents of a user interface.
         """
+        if self.parent:
```
corranwebster, line 242: Similarly
```suggestion
        if self.parent is not None:
```

corranwebster:
A couple of style nits, but I think we've been reasonably comprehensive in looking at where things can go wrong and dealing with them.  I'd say let's merge it and if we start seeing error handling going awry we can revert.

Or maybe we could implement a less error-prone error checking methodology!

aaronayres35:
@corranwebster good to go?

corranwebster (approved):

PR #1516, opened 2021-02-08
Expose `clearButtonEnabled` for Qt TextEditor

### LINKED ISSUE #1514: Expose `clearButtonEnabled` for Qt TextEditor
Since Qt 5.2, there is a property in `QLineEdit` that allows enabling a button for clearing the contents of the widget.

Screenshot 2021-02-08 at 11 53 07

Would it be possible to expose a flag for enabling this at the `traitsui` `TextEditor` level?
I haven't found an equivalent for the wx backend.

Here is a link to the relevant section of the Qt docs:
https://doc.qt.io/qt-5/qlineedit.html#clearButtonEnabled-prop

aaronayres35:
fixes #1514 

This PR adds a `clear_button` boolean trait on the `TextEditor` `ToolkitEditorFactory`, and uses it to call `setClearButtonEnable` if we are working with a `QLineEdit` on qt>=5.2.  It also adds a simple test to check it gets enabled as expected, and I separately modified the `TextEditor_demo.py` to confirm the changes worked as expected (changes not included in this PR).

I have only just started looking into the scope of adding something similar for wx / for a QTextEdit (used for the custom style of the `TextEditor`). For now, just adding the functionality for `QLineEdit`s on qt>=5.2 is only a couple line change.

### DIFF
--- a/traitsui/editors/text_editor.py
+++ b/traitsui/editors/text_editor.py
@@ -74,9 +74,9 @@ class ToolkitEditorFactory(EditorFactory):
     #: Is multi-line text allowed?
     multi_line = Bool(True)
 
-    #: Is editor readonly (will use custom / default editor appearance with readonly flag set to true)
-    #: in contrasrt with readonly style for item when completely another edito
-    #: is used
+    #: Is editor readonly (will use custom / default editor appearance with
+    #: readonly flag set to true) in contrasrt with readonly style for item
+    #: when completely another editor is used
     read_only = Bool(False)
 
     #: Is user input unreadable? (e.g., for a password)
@@ -98,7 +98,9 @@ class ToolkitEditorFactory(EditorFactory):
     placeholder = Str()
 
     #: Whether or not to display a clear button for the text.  This only works
-    #: in the qt>=5.2 backend for simple/text styles of the editor.
+    #: in the qt>=5.2 backend for simple/text styles of the editor.  Note this
+    #: trait is currently provisional and may be replaced in the future by a
+    #: more general feature.
     cancel_button = Bool(False)
 
     # -------------------------------------------------------------------------
--- a/traitsui/editors/text_editor.py
+++ b/traitsui/editors/text_editor.py
@@ -97,6 +97,10 @@ class ToolkitEditorFactory(EditorFactory):
     #: Grayed-out placeholder text to be displayed when the editor is empty.
     placeholder = Str()
 
+    #: Whether or not to display a clear button for the text.  This only works
+    #: in the qt>=5.2 backend for simple/text styles of the editor.
+    clear_button = Bool(False)
+
     # -------------------------------------------------------------------------
     #  Traits view definition:
     # -------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


traitsui/tests/editors/test_text_editor.py
```diff
@@ -133,6 +133,26 @@ def test_text_editor_custom_style_placeholder(self):
...
+                editor=TextEditor(clear_button=True),
+            )
+        )
+        tester = UITester()
+        with tester.create_ui(foo, dict(view=view)) as ui:
+            name_editor, = ui.get_editors("name")
+            try:
+                clear_button = name_editor.control.isClearButtonEnabled()
+            except AttributeError:
+                # placeholderText is introduced to QTextEdit since Qt 5.2
```
corranwebster, line 151:
```suggestion
                # isClearButtonEnabled is introduced to QLineEdit since Qt 5.2
```

corranwebster (approved):
Generally OK.

This looks like it is a special case of a more general feature which allows adding QAction-based buttons to line editors in Qt, so it might be better to implement the equivalent with Pyface Actions as a general feature.  So it might be a good idea to indicate that `clear_button` is "provisional".

traitsui/tests/editors/test_text_editor.py
```diff
@@ -133,6 +133,26 @@ def test_text_editor_custom_style_placeholder(self):
...
+                style="simple",
+                editor=TextEditor(clear_button=True),
+            )
+        )
+        tester = UITester()
+        with tester.create_ui(foo, dict(view=view)) as ui:
+            name_editor, = ui.get_editors("name")
+            try:
+                clear_button = name_editor.control.isClearButtonEnabled()
+            except AttributeError:
```
corranwebster, line 150: Might be better to pre-test with `hasattr` or test that `AttributeError` is because of `isClearButtonEnabled` and raise otherwise.

traitsui/editors/text_editor.py
```diff
@@ -97,6 +97,10 @@ class ToolkitEditorFactory(EditorFactory):
     #: Grayed-out placeholder text to be displayed when the editor is empty.
     placeholder = Str()
 
+    #: Whether or not to display a clear button for the text.  This only works
+    #: in the qt>=5.2 backend for simple/text styles of the editor.
+    clear_button = Bool(False)
```
corranwebster, line 102: Actually, can we call this `cancel_button` to keep it in line with the `SearchEditor` name for the same feature.

traitsui/editors/text_editor.py
```diff
@@ -74,9 +74,9 @@ class ToolkitEditorFactory(EditorFactory):
     #: Is multi-line text allowed?
     multi_line = Bool(True)
 
-    #: Is editor readonly (will use custom / default editor appearance with readonly flag set to true)
-    #: in contrasrt with readonly style for item when completely another edito
-    #: is used
+    #: Is editor readonly (will use custom / default editor appearance with
+    #: readonly flag set to true) in contrasrt with readonly style for item
```
rahulporuri, line 78:
```suggestion
    #: readonly flag set to true) in contrast with readonly style for item
```

rahulporuri (approved):
LGTM with a nitpicky suggestion

Issue #1060, opened 2020-07-23
Refactor Undo/Redo


corranwebster:
The undo/redo code is not particularly clean or well-written.  It is also not API-compatible with the apptools Undo/Redo system.

A proper re-write might mean bringing apptools Undo/Redo into Pyface (at least the core, but possibly all of it, so it can be used there) and then making the TraitsUI implement or adapt to the API for its undo/redo, so that it can integrate with application-level undo/redo.

While I suspect that this code is not heavily used externally to TraitsUI, there may need to take some care not to break the API exposed to editors and handlers without a clear migration path.

corranwebster:
As of this writing, we have moved the AppTools Undo/Redo into Pyface and made a release with this change, so it is now easily available to TraitsUI.  There is also #1510 pending, which adds testing, bugfixes, and makes the `UndoItem` API compatible with the `AbstractCommand` API.

The next step after this is probably to integrate an `UndoManager` into the `UI` class and refactor the `UndoHistory` class so that instead of a list of lists to hold the history it uses the `UndoManager`.  Almost all the logic of of the `UndoHistory` class is already implemented by the `UndoManager`, so `UndoHistory` can become a thin wrapper around this.  This change permits an application or Task's `UndoManager` to be shared with TraitsUI GUIs embedded into the view, if desired.  This is a potentially backwards-incompatible change, but only to code which is directly manipulating the history list - which really should not be happening.  This can probably be included in a minor release without breaking existing code.

One slight mismatch between the APIs is how to group multiple commands into a single command that can be performed with one undo/redo action.  The Pyface Undo/Redo handles this with the `begin_macro`/`end_macro` commands on the command stack; while `UndoHistory` provides the `extend` method and argument that adds extra commands to the current list of commands (effectively every command is a macro with at least one item) and whether or not to extend is usually controlled by the `do_undoable` code in the UI module.  At a first cut, we can support the `extend` API, but in the longer term it may make sense to instead refactor `do_undoable` and other places where `extend` is used to instead use begin/end macro.  The latter may again break third-party code which is using the `extend` method (most likely custom editor classes, so a bit more risky to change), but it may be possible to keep the `extend` API available while replacing most uses with the better API and removing the `extend` API in the next release.

[...]

PR #1517, opened 2021-02-08
Replace use of history list with CommandStack and UndoManager.

corranwebster:
This is a drop-in replacement for `UndoHistory` that is _almost_ backwards compatible:
- other than test failures due to https://github.com/enthought/pyface/issues/883 and changes needed where code was looking at `UndoHistory` internals, all previous tests pass.
- the `history` trait is no-longer used, but hasn't been removed yet (looking for feedback on this; is it better to remove now, or later)

Likelihood of 3rd party code breakage is small: only if it is dealing with internals of the Undo/Redo mechanism, and I am not aware of any such code.

This should not be merged until Pyface#883 is merged.  In fact the test runs for this PR will probably fail until Pyface#883 is merged.

This is another step towards #1060.


rahulporuri:
> This should not be merged until Pyface#883 is merged. In fact the test runs for this PR will probably fail until Pyface#883 is merged.

Not sure why CI is happy right now - but note that traitsui does not install pyface from source anymore. TraitsUI relies on the installed version of Pyface so if this PR requires pyface#883, it will need to wait for the pyface release.

corranwebster:
> Not sure why CI is happy right now - but note that traitsui does not install pyface from source anymore.

That'd be why it's happy.  I put a pyface version check to skip a couple of the tests which are expected to fail on 7.2.0 or before.

So given that, I would expect that:

- tests run against 7.2.0 will pass (with skips)
- tests of this code run against current Pyface master will fail
- tests of this code run against Pyface master after enthought/pyface#883 is merged will pass

rahulporuri (approved):
LGTM.

rahulporuri:
I've reviewed the PR - made a few recommendations - and i made those recommended changes to the PR. I think we can finally merge the PR.

PR #1518, opened 2021-02-08
[WIP] Allow TreeEditor to create new nodes for classes with required arugments

### LINKED ISSUE #1502: TreeEditor context menu action to create new nodes doesn't work with classes requiring arguments upon instantiation
Hi everyone ! I don't think we can really call this a bug, but as for now, it is not possible to use `TreeEditor`'s context menu action to create a new node, when the latter wraps a class that requires arguments upon instantiation: triggering the action doesn't let anything happen. An exception/error isn't even raised, the app doesn't crash. You can observe that using the code snippet hereafter, inspired from `traitsui.examples.demo.standard_editors.tree_editor`: you will be able to use context menu actions to create a new `Department` instance, but not a new `Employee` one.

```Python3
from traits.api import HasTraits, HasRequiredTraits, Str, Regex, List, Instance

from traitsui.api import Item, View, TreeEditor, TreeNode


class Employee(HasRequiredTraits):
    """ Defines a company employee. """

    name = Str(required=True)
    title = Str()
    phone = Regex(regex=r'\d\d\d-\d\d\d\d')

    def default_title(self):
        self.title = 'Senior Engineer'


class Department(HasTraits):
    """ Defines a department with employees. """

    name = Str('<unknown>')
    employees = List(Employee)


class Company(HasTraits):
    """ Defines a company with departments and employees. """

    name = Str('<unknown>')
    departments = List(Department)
    employees = List(Employee)


no_view = View()

tree_editor = TreeEditor(
    nodes=[
        # The first node specified is the top level one
        TreeNode(
            node_for=[Company],
            auto_open=True,
            # child nodes are
            children='',
            label='name',  # label with Company name
            view=View(['name'])
        ),
        TreeNode(
            node_for=[Company],
            auto_open=True,
            children='departments',
            label='=Departments',  # constant label
            view=no_view,
            add=[Department],
         ),
        TreeNode(
            node_for=[Company],
            auto_open=True,
            children='employees',
            label='=Employees',   # constant label
            view=no_view,
            add=[Employee]
        ),
        TreeNode(
            node_for=[Department],
            auto_open=True,
            children='employees',
            label='name',   # label with Department name
            view=View(['name']),
            add=[Employee]
        ),
        TreeNode(
            node_for=[Employee],
            auto_open=True,
            label='name',   # label with Employee name
            view=View(['name', 'title', 'phone'])
        )
    ]
)


class Partner(HasTraits):
    """ Defines a business partner."""

    name = Str('<unknown>')
    company = Instance(Company)

    traits_view = View(
        Item(name='company', editor=tree_editor, show_label=False),
        title='Company Structure',
        buttons=['OK'],
        resizable=True,
        style='custom',
        width=.3,
        height=500
    )


jason = Employee(name='Jason', title='Senior Engineer', phone='536-1057')
mike = Employee(name='Mike', title='Senior Engineer', phone='536-1057')
dave = Employee(name='Dave', title='Senior Developer', phone='536-1057')
martin = Employee(name='Martin', title='Senior Engineer', phone='536-1057')
duncan = Employee(name='Duncan', title='Consultant', phone='526-1057')

demo = Partner(
    name='Enthought, Inc.',
    company=Company(
        name='Enthought',
        employees=[dave, martin, duncan, jason, mike],
        departments=[
            Department(
                name='Business',
                employees=[jason, mike]
            ),
            Department(
                name='Scientific',
                employees=[dave, martin, duncan]
            )
        ]
    )
)


if __name__ == '__main__':
    demo.configure_traits()
```

_Tested with TraitsUI 7.1 and Traits 6.1, on Mac OSX 10.15.7_

aaronayres35:
fixes #1502 

The current solution is to first try to instantiate the class (ie. call `new_object = new_class()` like before) but to catch errors if this fails due to missing required arguments.  If we are dealing with `HasRequirdTraits` a `TraitError` will be raised and in this case, we simply look for the missing required traits and then create an instance of new_class using the defaults defined for those traits (via the `default` property of the `CTrait` object).  

I am still working on the case where we have a custom `__init__` method that is imposing the required attributes.  It is a bit more awkward, and I don't know how general we would want the solution to be.  What should be used for default values for attributes which aren't traits?  Should we try to guess, use a dumb default that might break things and ask the user to correct it somehow, or simply raise?
In general I feel like this use case is discouraged, but it could still occur. 

Also I have done manual testing by messing with the example code from the issue, but have not added a unit test for this currently.  It is a bit awkward to write a good test for currently...

### DIFF
--- a/traitsui/qt4/tree_editor.py
+++ b/traitsui/qt4/tree_editor.py
@@ -35,7 +35,7 @@ from pyface.qt import QtCore, QtGui
 from pyface.api import ImageResource
 from pyface.ui_traits import convert_image
 from pyface.timer.api import do_later
-from traits.api import Any, Event, Int, TraitError
+from traits.api import Any, Bool, Event, Int, TraitError
 from traitsui.editors.tree_editor import (
     CopyAction,
     CutAction,
@@ -44,6 +44,8 @@ from traitsui.editors.tree_editor import (
     PasteAction,
     RenameAction,
 )
+from traitsui.item import Item
+from traitsui.menu import OKCancelButtons
 from traitsui.tree_node import (
     ITreeNodeAdapterBridge,
     MultiTreeNode,
@@ -53,6 +55,7 @@ from traitsui.tree_node import (
 from traitsui.menu import Menu, Action, Separator
 from traitsui.ui_traits import SequenceTypes
 from traitsui.undo import ListUndoItem
+from traitsui.view import View
 
 from .clipboard import clipboard, PyMimeData
 from .editor import Editor
@@ -1333,7 +1336,52 @@ class SimpleEditor(Editor):
             }
             new_object = new_class(**kwargs)
 
-            prompt = True
+            added_traits_and_handlers = []
+            for trait_name, ctrait in req_traits_dict.items():
+                invalid_trait = Bool(True, visible=False)
+                new_object.add_trait(trait_name+"__invalid", invalid_trait)
+
+                def invalid_handler(event):
+                    obj = event.object
+                    setattr(obj, event.name+"__invalid", False)
+
+                new_object.observe(invalid_handler, trait_name)
+
+                added_traits_and_handlers.append(
+                    (trait_name, invalid_handler)
+                )
+
+            items = []
+            for trait_name in new_object.visible_traits():
+                if trait_name in req_traits_dict:
+                    items.append(
+                        Item(
+                            name=trait_name,
+                            invalid=trait_name+"__invalid",
+                            tooltip="This is a required trait")
+                    )
+                else:
+                    items.append(Item(trait_name))
+            modified_view = View(
+                *items,
+                buttons=OKCancelButtons
+            )
+            if new_object.edit_traits(
+                parent=self.control, kind="livemodal", view=modified_view
+            ).result:
+                for added_trait, handler in added_traits_and_handlers:
+                    new_object.observe(
+                        handler,
+                        added_trait,
+                        remove=True
+                    )
+                    new_object.remove_trait(added_trait+"__invalid")
+                self._undoable_append(node, object, new_object, False)
+
+                # Automatically select the new object if editing is being
+                # performed:
+                if self.factory.editable:
+                    self._tree.setCurrentItem(nid.child(nid.childCount() - 1))
         except TypeError:
             import inspect
             signature = inspect.signature(new_class.__init__)
@@ -1351,15 +1399,17 @@ class SimpleEditor(Editor):
                     raise 
                     #arguments[key] = None
             new_object = new_class(**arguments)
-        if (not prompt) or new_object.edit_traits(
-            parent=self.control, kind="livemodal"
-        ).result:
-            self._undoable_append(node, object, new_object, False)
+            prompt = True
+        else:
+            if (not prompt) or new_object.edit_traits(
+                parent=self.control, kind="livemodal"
+            ).result:
+                self._undoable_append(node, object, new_object, False)
 
-            # Automatically select the new object if editing is being
-            # performed:
-            if self.factory.editable:
-                self._tree.setCurrentItem(nid.child(nid.childCount() - 1))
+                # Automatically select the new object if editing is being
+                # performed:
+                if self.factory.editable:
+                    self._tree.setCurrentItem(nid.child(nid.childCount() - 1))
 
     # ----- Model event handlers: ---------------------------------------------
 
--- a/traitsui/qt4/tree_editor.py
+++ b/traitsui/qt4/tree_editor.py
@@ -1337,7 +1337,20 @@ class SimpleEditor(Editor):
         except TypeError:
             import inspect
             signature = inspect.signature(new_class.__init__)
-            print(signature)
+            arguments = {
+                param.name: param.default
+                for param in signature.parameters.values()
+                if param.name != 'self'
+            }
+            class_traits = new_class.class_traits()
+            for key in arguments:
+                if key in class_traits.keys():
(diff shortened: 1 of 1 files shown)


traitsui/qt4/tree_editor.py
```diff
@@ -1323,7 +1323,34 @@ def _menu_new_node(self, class_name, prompt=False):
...
+            }
+            class_traits = new_class.class_traits()
+            for key in arguments:
+                if key in class_traits.keys():
+                    arguments[key] = class_traits[key].default
+                if arguments[key] == inspect._empty:
+                    # we don't know what to supply as a default
+                    raise 
+                    #arguments[key] = None
+            new_object = new_class(**arguments)
```
aaronayres35, line 1353: Not sure how to best handle this scenario... @rahulporuri any thoughts on this?

traitsui/qt4/tree_editor.py
```diff
@@ -1323,7 +1323,34 @@ def _menu_new_node(self, class_name, prompt=False):
...
+            new_object = new_class()
+        except TraitError:
+            req_traits_dict = new_class.class_traits(required=True)
+            kwargs = {
+                trait_name: ctrait.default
+                for trait_name, ctrait in req_traits_dict.items()
+            }
+            new_object = new_class(**kwargs)
+
+            prompt = True
```
aaronayres35, line 1336: I set prompt to true here as a way to try to encourage the user to adjust/make any corrections that are needed since we use the defaults for required traits (may be overkill and I can undo this change)

corranwebster:
> If we are dealing with `HasRequirdTraits` a `TraitError` will be raised and in this case, we simply look for the missing required traits and then create an instance of new_class using the defaults defined for those traits (via the `default` property of the `CTrait` object).

I think some of the point of `HasRequiredTraits` is that the default values provided by `CTrait` isn't really a valid value (or doesn't result in a valid state for the instance), hence why users are required to supply them.  So this seems like a way to generate more subtle bugs, rather than failing immediately.

aaronayres35:
> > If we are dealing with `HasRequirdTraits` a `TraitError` will be raised and in this case, we simply look for the missing required traits and then create an instance of new_class using the defaults defined for those traits (via the `default` property of the `CTrait` object).
> 
> I think some of the point of `HasRequiredTraits` is that the default values provided by `CTrait` isn't really a valid value (or doesn't result in a valid state for the instance), hence why users are required to supply them. So this seems like a way to generate more subtle bugs, rather than failing immediately.

I see, that makes sense.  I tried to update the code so that if a user tries to create a new node for an object that has required traits, they are effectively forced (or strongly encouraged) to input a value for that trait.  The current implementation is really over engineered and ugly (far from being ready for thorough review / to be pushed through), but does the UX, general idea make sense?

To test it, one can run the example given in the issue description and try to use the right click context menu to create a new `Employee`.  When doing so, a ui generated from `edit_traits` being called on the object pops up with required traits fields initially set to invalid, until their value is changed, with a tooltip saying they are required.  I wasn't sure what exactly should be used to validate the users inputs for those traits, so I just tried to make sure a user put in something. 

Here is a screen recording:

https://user-images.githubusercontent.com/36972686/107439886-62fb4480-6af8-11eb-8c9d-9616de98cd5d.mov

corranwebster:
Having a look at this, I think it's adding complexity to the code but will still miss corner-cases.

The core problem we have is that the new menu item is using the class name as the hook for creating new instances, rather than somehow allowing users to supply something else to create the class instance.  The issue with `HasRequiredTraits` could be solved simply by having the developer be able to supply an arbitrary callable as a factory for new instances: this would then let them populate the required traits as needed, either by asking the user or by providing them via a closure or callable factory class.

So this requires replacing these two lines:
https://github.com/enthought/traitsui/blob/23f17d1a5b725d5388abd5cbcd98d8a614913a03/traitsui/qt4/tree_editor.py#L1325-L1326
with something else.

But notice that the `class_name` is being supplied as an argument, so we have the possibility of supplying something else here:
https://github.com/enthought/traitsui/blob/23f17d1a5b725d5388abd5cbcd98d8a614913a03/traitsui/qt4/tree_editor.py#L1320

The call to this function comes from the `Action` created here:
https://github.com/enthought/traitsui/blob/23f17d1a5b725d5388abd5cbcd98d8a614913a03/traitsui/qt4/tree_editor.py#L1070-L1081
Note that the class name has to be passed as a string, because the `action` is being given as a string which is eval-ed or exec-ed rather than a callback.  This is not very nice code.

So I would focus on a refactor of the whole of `_new_actions`:
- firstly refactor to replace the `action` string in the `Action` with an `on_perform` callable (it is expected to take the object as an argument).  Something like:
```
factory = ... # worry about how to get this later, but initially it can be the class object
def perform_add(object):
    self._menu_new_node(factory, prompt)
items.append(Action(name=name, on_perform=perform_add))
```
- if you don't like closures to capture the info you need, a subclass of `Action` which you give all the information you need as arguments would also work.
- then you refactor `_menu_new_node` so that it accepts the factory rather than the class name, and you can just do `new_object = factory()`.

Those two changes are probably enough for one PR as they just make an incremental improvement that removes a hacky eval.

You then sit down and think about how you need to change the tree node API so that it can better specify non-class factories (ie. an improvement or alternative to `get_add` and `get_name`).  You basically need some way to specify the name to display in the menu, and the factory function to call for each type of thing you want to add.  There may even be something like this already in the API.  The default implementation of this should just do what the current code does, but it should allow developers to have their own tree node classes supply alternatives.

aaronayres35:
Thank you very much @corranwebster !  Your comments have been extremely helpful/valuable for me to learn from.  I've opened #1524 and am closing this PR in favor of that one.

Issue #1052, opened 2020-07-22
Use the new traits observe instead of on_trait_change


rahulporuri:
Traits 6.1.0 introduced the new `observe` decorator and the underlying observation framework. This is meant to replace the existing `on_trait_change`-based event listeners.

Ref : https://docs.enthought.com/traits/traits_user_manual/notification.html

kitchoi:
I'd propose having CI to test against Traits 6.0 before we make this move. See https://github.com/enthought/traitsui/issues/1054

For existing functionality, we should defer the migration because otherwise we will inevitably force downstream projects to upgrade.  Use of Traits >=6.1 feature can happen for new features of TraitsUI though, because in that case projects can opt-in.

kitchoi:
Alternatively we could do something like `PrefixList` where we have a compatibility layer that does one thing if we are on Traits < 6.1 but does another for Traits >= 6.1. While it would help testing out Traits 6.1+, this could result in very messy code paths in some situations, but maybe worth considering in more trivial situations. Just a thought.

rahulporuri:
I don't think we'll be able to use a compatibility layer for `on_trait_change` vs `observe`. Can we?

kitchoi:
> I don't think we'll be able to use a compatibility layer for on_trait_change vs observe. Can we?

I am not suggesting we should do that. It is possible but likely too messy to be worth doing in most situations.

PR #1520, opened 2021-02-12
Replace use of depends_on in Traits Properties with observe

rahulporuri:
This PR replaces the use of `depends_on` metadata on Traits Properties with the use of `observe` metadata. Contributes towards #1052

### DIFF
--- a/examples/tutorials/traitsui_4.0/items.py
+++ b/examples/tutorials/traitsui_4.0/items.py
@@ -158,7 +158,7 @@ class LeagueModelView(ModelView):
     got_hit = Button('Got a Hit')
 
     # The total number of hits
-    total_hits = Property(observe='model.teams.players.hits')
+    total_hits = Property(depends_on='model.teams.players.hits')
 
     @cached_property
     def _get_total_hits(self):
--- a/examples/tutorials/traitsui_4.0/items.py
+++ b/examples/tutorials/traitsui_4.0/items.py
@@ -121,7 +121,7 @@ class Team(HasTraits):
     players = List(Player)
 
     # The number of players on the team:
-    num_players = Property(depends_on='players')
+    num_players = Property(observe='players')
 
     def _get_num_players(self):
         """ Implementation of the 'num_players' property.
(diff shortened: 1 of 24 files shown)


traitsui/ui.py
```diff
@@ -93,7 +94,12 @@ class UI(HasPrivateTraits):
     history = Any()
 
     #: The KeyBindings object (if any) for this UI:
-    key_bindings = Property(depends_on=["view._key_bindings", "context"])
+    key_bindings = Property(
+        observe=[
+            trait("view._key_bindings", optional=True),
+            "context",
+        ]
+    )
```
aaronayres35, line 102: I think this use of `trait` may be broken (but break is hidden by `optional=True`)
see https://github.com/enthought/pyface/pull/892 and https://github.com/enthought/pyface/pull/880#discussion_r573969099

examples/tutorials/traitsui_4.0/items.py
```diff
@@ -158,7 +158,7 @@ class LeagueModelView(ModelView):
     got_hit = Button('Got a Hit')
 
     # The total number of hits
-    total_hits = Property(depends_on='model.teams.players.hits')
+    total_hits = Property(observe='model.teams.players.hits')
```
aaronayres35, line 161: I believe we need to insert `.items` on this lists here, eg. `model.teams.items.players.items.hits`

aaronayres35 (commented):
Mostly LGTM, but a couple suggested changes.  Most notably in traitsui/ui.py for the use of `trait` on  the extended trait name "view._key_bindings"

I apologize for the non-unified review.  I inadvertently clicked "add single comment", and also "comment" rather than "request changes"

traitsui/ui.py
rahulporuri: I removed this to go ahead and merge this PR. We can address the few % of on trait change uses later.

examples/tutorials/traitsui_4.0/items.py
rahulporuri: same here - i undid this change.

aaronayres35 (approved):
LGTM

PR #1523, opened 2021-02-20
Dynamic otc to observe1

aaronayres35:
contributes to #1052 

This PR updates a few editors to use dynamic `observe` instead of `on_trait_change`. More similar PRs to follow

### DIFF
--- a/traitsui/editors/shell_editor.py
+++ b/traitsui/editors/shell_editor.py
@@ -56,11 +56,11 @@ class _ShellEditor(Editor):
         if locals is None:
             object = self.object
             shell.bind("self", object)
-            shell.on_trait_change(
+            shell.observe(
                 self.update_object, "command_executed", dispatch="ui"
             )
             if not isinstance(value, dict):
-                object.on_trait_change(self.update_any, dispatch="ui")
+                object.observe(self.update_any, dispatch="ui")
             else:
                 self._base_locals = locals = {}
                 for name in self._shell.interpreter().locals.keys():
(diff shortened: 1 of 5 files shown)


traitsui/editors/shell_editor.py
```diff
@@ -120,24 +120,24 @@ def update_editor(self):
...
-            self.value[name] = new
+            self.value[event.name] = event.new
 
     def dispose(self):
         """ Disposes of the contents of an editor.
         """
-        self._shell.on_trait_change(
-            self.update_object, "command_executed", remove=True
+        self._shell.observe(
+            self.update_object, "command_executed", remove=True, dispatch="ui"
```
aaronayres35, line 137: ref: https://github.com/enthought/traits/issues/1212

traitsui/qt4/button_editor.py
```diff
@@ -63,11 +63,8 @@ def init(self, parent):
             self.control = QtGui.QToolButton()
             self.control.toolButtonStyle = QtCore.Qt.ToolButtonTextOnly
             self.control.setText(self.string_value(label))
-            self.object.on_trait_change(
-                self._update_menu, self.factory.values_trait
-            )
-            self.object.on_trait_change(
-                self._update_menu, self.factory.values_trait + "_items"
+            self.object.observe(
+                self._update_menu, self.factory.values_trait + ".items"
```
rahulporuri, line 67: IIUC this works because for `container` traits, we can simply use `container.items` to listen to both changes to the `container` trait and the `items` in the container.

traitsui/qt4/list_str_editor.py
```diff
@@ -170,16 +170,16 @@ def init(self, parent):
 
         # Make sure we listen for 'items' changes as well as complete list
         # replacements:
-        self.context_object.on_trait_change(
-            self.update_editor, self.extended_name + "_items", dispatch="ui"
+        self.context_object.observe(
+            self.update_editor, self.extended_name + ":items", dispatch="ui"
```
rahulporuri, line 174: This change conflicts with the comment string - the comment says we should listen to changes in items as well as complete replacements - but it looks like we're not only listening to changes in items?

aaronayres35: Yeah I agree that is misleading.  Previously I believe the on_trait_change code was only listening to items changes so code should be consistent.  It is possible the code is supposed to listen to both as the comment says... It is hard to tell, and I'm unsure if we want to change behavior, even if the change is to the _right_ behavior.
Should I replace the `:` with a `.`?

rahulporuri: What happens if you switch `:` to `.`? Does it break the testsuite and/or examples? I'll try to git blame and see how this code evolved.

aaronayres35: No, I'm not seeing anything break

rahulporuri: I was hoping that the comment string was out of date because of some changes - but neither the comment string nor the code itself changed functionally since it was first checked in.

Let's go from `:` to `.` and see what breaks.

rahulporuri (approved):
LGTM

PR #1524, opened 2021-02-22
Refactor TreeEditor _new_actions and _menu_new_node to avoid hacky eval

aaronayres35:
Part of #1502 

This PR supersedes #1518.

This PR follows the approach outlined here: https://github.com/enthought/traitsui/pull/1518#issuecomment-776682726

### DIFF
--- a/traitsui/qt4/tree_editor.py
+++ b/traitsui/qt4/tree_editor.py
@@ -692,15 +692,6 @@ class SimpleEditor(Editor):
                 return node
         return None
 
-    def _node_for_class_name(self, class_name):
-        """ Returns the node and class associated with a specified class name.
-        """
-        for node in self.factory.nodes:
-            for klass in node.node_for:
-                if class_name == klass.__name__:
-                    return (node, klass)
-        return (None, None)
-
     def _update_icon(self, nid):
         """ Updates the icon for a specified node.
         """
--- a/traitsui/qt4/tree_editor.py
+++ b/traitsui/qt4/tree_editor.py
@@ -1052,21 +1052,25 @@ class SimpleEditor(Editor):
         object = self._data[1]
         items = []
         add = node.get_add(object)
-        if len(add) > 0:
-            for klass in add:
-                prompt = False
-                if isinstance(klass, tuple):
-                    klass, prompt = klass
-                add_node = self._node_for_class(klass)
-                if add_node is not None:
-                    class_name = klass.__name__
-                    name = add_node.get_name(object)
-                    if name == "":
-                        name = class_name
-                    factory = klass
-                    def perform_add(object):
-                        self._menu_new_node(factory, prompt)
-                    items.append(Action(name=name, on_perform=perform_add))
+        # return early if there are no items to be added in the tree
+        if len(add) == 0:
+            return items
+
+        for klass in add:
+            prompt = False
+            if isinstance(klass, tuple):
+                klass, prompt = klass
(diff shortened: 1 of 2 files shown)


traitsui/qt4/tree_editor.py
```diff
@@ -1317,13 +1314,18 @@ def _on_nid_changed(self, nid, col):
...
         node, object, nid = self._data
         self._data = None
-        new_node, new_class = self._node_for_class_name(class_name)
-        new_object = new_class()
+        try:
+            new_object = factory()
+        except:
+            from traitsui.api import raise_to_debug
+            raise_to_debug()
+            return
```
aaronayres35, line 1327: This `try: ... except: ...` was necessary since without it the new code changes would cause an exception to be raised if one tried to create a new node for an object with required traits.  Previously the hacky eval swallowed this here:
https://github.com/enthought/traitsui/blob/a4be4b539be4dfae295525f86ed18f04ad67bc71/traitsui/qt4/tree_editor.py#L1203-L1219

Once we complete the second portion of this fix, namely "You then sit down and think about how you need to change the tree node API so that it can better specify non-class factories (ie. an improvement or alternative to get_add and get_name). You basically need some way to specify the name to display in the menu, and the factory function to call for each type of thing you want to add. There may even be something like this already in the API. The default implementation of this should just do what the current code does, but it should allow developers to have their own tree node classes supply alternatives." 
we will want to remove this.  At that point a programmer will have all the tools needed to specify how to create a new node that has required arguments, and if they do so in a way that we still can not create a node we should complain and raise an exception.  However now, with the machinery not yet in place, we don't want this to fail.  This preserves the current behavior.

traitsui/qt4/tree_editor.py
```diff
@@ -1317,13 +1314,18 @@ def _on_nid_changed(self, nid, col):
             else:
                 self._set_label(nid, col)
 
-    def _menu_new_node(self, class_name, prompt=False):
+    def _menu_new_node(self, factory, prompt=False):
         """ Adds a new object to the current node.
         """
         node, object, nid = self._data
         self._data = None
-        new_node, new_class = self._node_for_class_name(class_name)
```
aaronayres35, line 1325: `_node_for_class_name` is now no longer called anywhere.  Perhaps the method can be deleted?  It could be useful in some other case so maybe we want to just leave it, but given it is currently unused, it may make more sense to just delete it and re-add it if a future use case comes about that could make use of it.

traitsui/qt4/tree_editor.py
```diff
@@ -1072,13 +1072,10 @@ def _new_actions(self, node, object):
                     name = add_node.get_name(object)
```
rahulporuri, line 1072: This is a nit-picky comment but as long as we are refactoring code in this PR, how about refactoring the `_new_actions` method to return early if `len(add) == 0`.

```
# return early if there are no items to be added in the tree
if len(add) == 0:
    return items

for klass in add:
    ....
```

Refactoring the code like above makes the code slightly more readable by removing one level of indentation.

Similarly, we can change `if add_node is not None` to `if add_node is None` to remove one more level of indentation in the code.

traitsui/qt4/tree_editor.py
rahulporuri: `_node_for_class_name` looks straightforward so if you don't find any users of it, please go ahead and remove it.

rahulporuri (approved):
LGTM

PR #1525, opened 2021-02-23
Dynamic otc to observe 2

aaronayres35:
contributes to #1052

This PR updates a few files to use dynamic `observe` instead of `on_trait_change`. More similar PRs to follow

### DIFF
--- a/traitsui/qt4/ui_base.py
+++ b/traitsui/qt4/ui_base.py
@@ -402,7 +402,6 @@ class BaseDialog(BasePanel):
 
         def set_status_text(event):
             text = event.new
-            if control:
-                control.setText(text)
+            control.setText(text)
 
         return set_status_text
--- a/traitsui/qt4/ui_base.py
+++ b/traitsui/qt4/ui_base.py
@@ -105,19 +105,22 @@ class BasePanel(_BasePanel):
 
         return button
 
-    def _on_undoable(self, state):
+    def _on_undoable(self, event):
         """Handles a change to the "undoable" state of the undo history
         """
+        state = event.new
         self.undo.setEnabled(state)
 
-    def _on_redoable(self, state):
+    def _on_redoable(self, event):
         """Handles a change to the "redoable" state of the undo history.
         """
+        state = event.new
         self.redo.setEnabled(state)
 
-    def _on_revertable(self, state):
+    def _on_revertable(self, event):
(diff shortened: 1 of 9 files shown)


traitsui/qt4/ui_base.py
```diff
@@ -387,7 +390,7 @@ def _add_statusbar(self):
                     name = name[col + 1 :]
                 obj = self.ui.context[obj]
                 set_text = self._set_status_text(item_control)
-                obj.on_trait_change(set_text, name, dispatch="ui")
+                obj.observe(set_text, name, dispatch="ui")
```
aaronayres35, line 393: These need to be disconnected at some point (I think the lack of being disconnected is causing the latest CI failure), although I'm not sure where/when that should happen.  I suppose in the dispose method with a `if self.ui.view.statusbar is not None` check?

For now I tried adding a dumb fix to just check if the control exists in the handler before trying to call `.set_text()` on it

aaronayres35:
Ah I realize my hacky fix was dumb...

traitsui/qt4/ui_live.py
```diff
@@ -201,7 +201,7 @@ def init(self, ui, parent, style):
                         self.control.accept,
                         default=default,
                     )
-                    ui.on_trait_change(self._on_error, "errors", dispatch="ui")
+                    ui.observe(self._on_error, "errors", dispatch="ui")
```
rahulporuri, line 204: We don't seem to be unhooking this listener.

traitsui/qt4/ui_modal.py
```diff
@@ -155,7 +155,7 @@ def init(self, ui, parent, style):
                         self.control.accept,
                         default=default,
                     )
-                    ui.on_trait_change(self._on_error, "errors", dispatch="ui")
+                    ui.observe(self._on_error, "errors", dispatch="ui")
```
rahulporuri, line 158: Looks like we have two listeners that don't get unhooked in this module. Instead of going over your PRs again and creating an issue pointing out unhooked listeners like we did in Pyface, let's create an issue in traitsui now - which we can continually update?

rahulporuri (approved):
LGTM with a couple of comments.

traitsui/qt4/ui_modal.py
aaronayres35: I've opened #1526

PR #1537, opened 2021-03-12
Document enabled_when / visible_when better

### LINKED ISSUE #833: Document that enabled_when / visible_when don't update for nested traits
If the string provided to `enabled_when` or `visible_when` on an `Item` refers to a nested trait, then the enabled-ness or visibility don't update when the nested trait value changes (see example below). This feels inconsistent with the behavior of `on_trait_change` or `depends_on`.

A typical workaround is to add a shadow trait to the parent class (for example in the below, adding a trait `a_is_magic_number = Property(depends_on='a:x')`. This workaround feels like redundant work, though.

#### Example

```python
from traits.api import *
from traitsui.api import *


class A(HasTraits):
    number = Int()
    is_magic_number = Property(depends_on='number')

    def _get_is_magic_number(self):
        return self.number == 1729

    traits_view = View(Item('number'))


class B(HasTraits):
    a = Instance(A, ())
    done = Button()

    traits_view = View(
        UItem('a', style='custom'),
        UItem('done', enabled_when='a.is_magic_number'),
    )


B().configure_traits()
```

aaronayres35:
closes #833 

This PR adds to the trait definition comments for `enabled_when` / `visible_when` to comment on some common points of confusion.  It also mentions in the view context section of the docs the HasTraits `trait_context` method to override the default context.

### DIFF
--- a/docs/source/traitsui_user_manual/view.rst
+++ b/docs/source/traitsui_user_manual/view.rst
@@ -244,9 +244,17 @@ Visibility and status
    :term:`Handler` (see :ref:`controlling-the-interface-the-handler`).
 
    enabled_when: str
-      Expression that determines whether the group can be edited.
+      Python expression that determines whether the group can be edited.
+      The expression will be evaluated any time a trait on an object in the
+      UI's context is changed. As a result, changes to nested traits that don't
+      also change a trait on some object in the context may not trigger the
+      expression to be evaluated.
    visible_when: str
-      Expression that determines visibility of group.
+      Python expression that determines visibility of group. The expression
+      will be evaluated any time a trait on an object in the UI's context is
+      changed. As a result, changes to nested traits that don't also change a
+      trait on some object in the context may not trigger the expression to be
+      evaluated.
    defined_when: str
       Expression that determines inclusion of group in parent.
    has_focus: bool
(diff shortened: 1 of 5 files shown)


traitsui/group.py
```diff
@@ -159,16 +159,27 @@ class Group(ViewSubElement):
...
     #: become visible. All **visible_when** conditions are checked each time
-    #: that any trait value is edited in the display. Therefore, you can use
-    #: **visible_when** conditions to hide or show groups in response to user
-    #: input.
+    #: that any trait value on an object in the UI's context is changed.
+    #: Therefore, you can use **visible_when** conditions to hide or show
+    #: groups in response to user input. Be aware that this only applies to
+    #: traits in the UI's context. As a result, changes to nested traits may
+    #: not trigger the expression to be checked they don't also change a trait
+    #: on some object in the context. Additionally, the expression needs to be
```
rahulporuri, line 167: I reordered the sentence a little and fixed what I think is a typo - `they` to `that` in `... they dont also change a trait ...`

If you are happy with this rewording, these changes need to be propogated to the other docstrings.

```suggestion
    #: traits in the UI's context. As a result, changes to nested traits that
    #: don't also change a trait on some object in the context may not
    #: trigger the expression to be checked. Additionally, the expression
    #: needs to be
```

docs/source/traitsui_user_manual/view.rst
```diff
@@ -244,7 +244,7 @@ Visibility and status
    :term:`Handler` (see :ref:`controlling-the-interface-the-handler`).
 
    enabled_when: str
-      Expression that determines whether of group can be edited.
+      Expression that determines whether the group can be edited.
```
rahulporuri, line 247: It'll be very useful to add a note somewhere here - similar to the node that we added in code. It's always better to mention this in user facing docs instead of in code.

rahulporuri: and repeat the same comment a second time later in this file where we document `enabled_when`, `visible_when` again.

rahulporuri (commented):
Almost LGTM with a few comments/requests for changes/additions.

docs/source/traitsui_user_manual/view.rst
aaronayres35: I've added some comments here.  One thing I am noticing is that `defined_when` seems to be handled differently internally... I was going to make a general comment about all 3 at the bottom, but looking at `traitsui.ui.py` it appears `defined_when` is doing something slightly different?  I need to look at the code more closely.  For now I just added slightly redundant comments to both the enabled_when and visible_when description

rahulporuri (approved):
LGTM

PR #1540, opened 2021-03-15
Don't use html for a variable name when using html module from standard library

### LINKED ISSUE #1538: Help Button issue
When a HelpButton is pushed it does not appear and I get the follow Traceback:

```
  File "/Users/meglinsky/opt/anaconda3/lib/python3.8/site-packages/traitsui/base_panel.py", line 91, in _on_help
    self.ui.handler.show_help(self.ui.info)
  File "/Users/meglinsky/opt/anaconda3/lib/python3.8/site-packages/traitsui/handler.py", line 186, in show_help
    on_help_call()(info, control)
  File "/Users/meglinsky/opt/anaconda3/lib/python3.8/site-packages/traitsui/help.py", line 30, in default_show_help
    toolkit().show_help(info.ui, control)
  File "/Users/meglinsky/opt/anaconda3/lib/python3.8/site-packages/traitsui/qt4/toolkit.py", line 321, in show_help
    ui_panel.show_help(ui, control)
  File "/Users/meglinsky/opt/anaconda3/lib/python3.8/site-packages/traitsui/qt4/ui_panel.py", line 363, in show_help
    html.escape(item.get_label(ui)),
UnboundLocalError: local variable 'html' referenced before assignment
```

aaronayres35:
fixes #1538 

Previously we were using the html module from the python standard library in a function, as well as defining a local variable `html` that held on to the html content to be used to create an `HTMLHelpWindow` instance.  This PR adds a very simple test to showcase the bug, and fixes it by giving the `html` variable a new name `html_content`.   Additionally, I changed the import statement to only import what we need from `html`.  Namely, `from html import escape`.  Either of these fixes would work alone (it may be overkill to do them both), but I figured it couldn't hurt.

### DIFF
--- a/traitsui/qt4/tests/test_ui_panel.py
+++ b/traitsui/qt4/tests/test_ui_panel.py
@@ -13,6 +13,7 @@ import unittest
 from traits.api import HasTraits, Int, Str, Instance
 from traitsui.api import View, Item, Group
 from traitsui.menu import ToolBar, Action
+
 from traitsui.tests._tools import (
     create_ui,
     requires_toolkit,
--- a/traitsui/qt4/tests/test_ui_panel.py
+++ b/traitsui/qt4/tests/test_ui_panel.py
@@ -11,9 +11,10 @@
 import unittest
 
 from traits.api import HasTraits, Int, Str, Instance
-from traitsui.api import View, Item, Group
+from traitsui.api import HelpButton, View, Item, Group
 from traitsui.menu import ToolBar, Action
 
+from traitsui.testing.api import MouseClick, UITester
 from traitsui.tests._tools import (
     create_ui,
     requires_toolkit,
@@ -40,6 +41,18 @@ class FooPanel(HasTraits):
 
     def _toolbar_default(self):
         return ToolBar(Action(name="Open file"))
+
+
+class HelpPanel(HasTraits):
+    my_int = Int(2, help='this is the help for my int')
+
+    def default_traits_view(self):
+        view = View(
+            Item(name="my_int"),
+            title="HelpPanel",
+            buttons=[HelpButton],
+        )
+        return view
 
 
 class FooDialog(HasTraits):
@@ -172,6 +185,14 @@ class TestUIPanel(unittest.TestCase):
             # No button
             self.assertIsNone(ui.control.findChild(QtGui.QPushButton))
 
+    def test_show_help(self):
+        panel = HelpPanel()
+
+        tester = UITester()
+
+        with tester.create_ui(panel) as ui:
+            help_button = tester.find_by_id(ui, 'Help')
+            help_button.perform(MouseClick())
 
 @requires_toolkit([ToolkitName.qt])
 class TestPanelLayout(unittest.TestCase):
--- a/traitsui/qt4/tests/test_ui_panel.py
+++ b/traitsui/qt4/tests/test_ui_panel.py
@@ -11,10 +11,8 @@
 import unittest
 
 from traits.api import HasTraits, Int, Str, Instance
-from traitsui.api import HelpButton, View, Item, Group
+from traitsui.api import View, Item, Group
 from traitsui.menu import ToolBar, Action
-
-from traitsui.testing.api import MouseClick, UITester
 from traitsui.tests._tools import (
     create_ui,
     requires_toolkit,
@@ -41,18 +39,6 @@ class FooPanel(HasTraits):
 
     def _toolbar_default(self):
         return ToolBar(Action(name="Open file"))
-
-
-class HelpPanel(HasTraits):
-    my_int = Int(2, help='this is the help for my int')
-
-    def default_traits_view(self):
-        view = View(
-            Item(name="my_int"),
-            title="HelpPanel",
-            buttons=[HelpButton],
-        )
-        return view
 
 
 class FooDialog(HasTraits):
@@ -184,14 +170,6 @@ class TestUIPanel(unittest.TestCase):
 
             # No button
             self.assertIsNone(ui.control.findChild(QtGui.QPushButton))
-
-    # Regression test for enthought/traitsui#1538
-    def test_show_help(self):
-        panel = HelpPanel()
-        tester = UITester()
-        with tester.create_ui(panel) as ui:
-            help_button = tester.find_by_id(ui, 'Help')
-            help_button.perform(MouseClick())
 
 
 @requires_toolkit([ToolkitName.qt])
--- a/traitsui/qt4/tests/test_ui_panel.py
+++ b/traitsui/qt4/tests/test_ui_panel.py
@@ -185,14 +185,14 @@ class TestUIPanel(unittest.TestCase):
             # No button
             self.assertIsNone(ui.control.findChild(QtGui.QPushButton))
 
+    # Regression test for enthought/traitsui#1538
     def test_show_help(self):
         panel = HelpPanel()
-
         tester = UITester()
-
         with tester.create_ui(panel) as ui:
             help_button = tester.find_by_id(ui, 'Help')
             help_button.perform(MouseClick())
+
 
 @requires_toolkit([ToolkitName.qt])
 class TestPanelLayout(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


aaronayres35:
CI is stalling on the new tests but I am unable to reproduce locally.  I thought it may have been an issue from the Modal Dialog, but I am not sure.  It also seems to be OS/toolkit specific? In some cases the test runs fine on CI. 

I will investigate more tomorrow

aaronayres35:
I was able to reproduce the test suite hanging on an Ubuntu machine. Investigating more now

aaronayres35:
I got this working locally on Ubuntu, but I had to force the help dialog to be modal... :/ which we very well may want to avoid.  The help window that pops up is blocking the test, however it is not modal so trying to us modal dialog tester did not work.  However if I call `exec_` instead of `show` the dialog is modal and the test works exactly as expected.  See:
https://srinikom.github.io/pyside-docs/PySide/QtGui/QDialog.html#PySide.QtGui.PySide.QtGui.QDialog.exec_

This is a change in behavior and so should probably be avoided.

I am unsure how to resolve this test hanging without this...
I tried to simply close the original window by clicking ok while the help window was up, but that proved unsuccessful.
I will keep looking for an alternative

traitsui/qt4/ui_panel.py
```diff
@@ -1283,7 +1283,7 @@ class HTMLHelpWindow(QtGui.QDialog):
     """ Window for displaying Traits-based help text with HTML formatting.
     """
 
-    def __init__(self, parent, html, scale_dx, scale_dy):
+    def __init__(self, parent, html_content, scale_dx, scale_dy):
```
rahulporuri, line 1286: Let's undo this change - it is unnecessary and it breaks backwards compatibility in a public object.

IIUC, for the scope of this method, `html` variable name will be bound to the argument passed in.

traitsui/qt4/ui_panel.py
```diff
@@ -1310,7 +1310,7 @@ def __init__(self, parent, html, scale_dx, scale_dy):
 
         # Position and show the dialog
         position_window(self, parent=parent)
-        self.show()
+        self.exec_()
```
rahulporuri, line 1313: I'd like this PR merged by the end of this week so if we haven't figured out how to add a regression test by then, let's just merge this PR and open an issue to write a regression test.

traitsui/qt4/ui_panel.py
rahulporuri: note that the variables created in `show_help` and `show_help_popup` can still stay as `html_content` - that definitely makes the code more readable.

traitsui/wx/ui_panel.py
```diff
@@ -1152,7 +1152,7 @@ class HTMLHelpWindow(wx.Frame):
     """ Window for displaying Traits-based help text with HTML formatting.
     """
 
-    def __init__(self, parent, html, scale_dx, scale_dy):
+    def __init__(self, parent, html_content, scale_dx, scale_dy):
```
rahulporuri, line 1155: Same as with the other module, let's not modify the public API of this object.

rahulporuri (commented):
LGTM with a few comments/caveat. I tested this with the code from https://github.com/enthought/traitsui/issues/1538 - and it works as expected.

traitsui/tests/test_ui_panel.py
```diff
@@ -55,3 +75,25 @@ def test_grouped_layout_with_springy(self):
...
+
+    # Regression test for enthought/traitsui#1538
+    @unittest.skipIf(no_modal_dialog_tester, "ModalDialogTester unavailable")
+    def test_show_help(self):
+
+        # This help window is not actually modal, when opened it will be the
+        # active window not active modal widget
+        class MyTester(ModalDialogTester):
+            def get_dialog_widget(self):
+                return self._qt_app.activeWindow()
```
aaronayres35, line 87: FWIW the test passed for me on mac and ubuntu locally without this _but_ without this the call of `x.click_button(OK)` effectively ends up doing nothing as we are trying to call a method on None.

This way we are actually closing the window we open which feels like the right thing to do.

The real cause of the test hang I believe was not having `auto_process_events=False`.  I should have realized this much faster than I did.  Previously I had only remembered this being necessary when using `UITester` _inside_ the use of the ModalDialogTester callback.  
For example:
https://github.com/enthought/traitsui/blob/62033f5e1319d349bb3e980d9083f2c2a24a431a/traitsui/tests/editors/test_instance_editor.py#L211-L243

In cases where we just use UI Tester to trigger the dialog being opened, it has not been needed (until now at least AFAICR) 
e.g.
https://github.com/enthought/traitsui/blob/62033f5e1319d349bb3e980d9083f2c2a24a431a/traitsui/examples/demo/Standard_Editors/tests/test_ButtonEditor_demo.py#L42-L69

Nonetheless, this test does what we want it to now and test suite seems to not be hanging

traitsui/qt4/ui_panel.py
```diff
@@ -393,8 +393,8 @@ def show_help_popup(event):
     # of the object with the 'help' trait):
     help = getattr(control, "help", None)
     if help is not None:
-        html = template.item_html % (control.GetLabel(), help)
-        HTMLHelpWindow(control, html, 0.25, 0.13)
+        html_content = template.item_html % (control.GetLabel(), help)
+        HTMLHelpWindow(control, html_content, 0.25, 0.13)
```
jwiggins, line 397: Not here, but in the `HTMLHelpWindow.__init__` method, there is still a parameter named `html`

traitsui/qt4/tests/test_ui_panel.py
```diff
@@ -13,7 +13,6 @@
 from traits.api import HasTraits, Int, Str, Instance
 from traitsui.api import View, Item, Group
 from traitsui.menu import ToolBar, Action
-
```
jwiggins, line 16: Nit: watch your diffs

traitsui/tests/test_ui_panel.py
jwiggins: But now this test is Qt-specific, without being marked as such

traitsui/wx/ui_panel.py
```diff
@@ -414,8 +414,8 @@ def show_help_popup(event):
     # of the object with the 'help' trait):
     help = getattr(control, "help", None)
     if help is not None:
-        html = template.item_html % (control.GetLabel(), help)
-        HTMLHelpWindow(control, html, 0.25, 0.13)
+        html_content = template.item_html % (control.GetLabel(), help)
+        HTMLHelpWindow(control, html_content, 0.25, 0.13)
```
jwiggins, line 418: Same story as Qt, `HTMLHelpWindow.__init__` has a parameter named `html`

jwiggins (commented):
Mostly looks OK, but the test is Qt-only

traitsui/qt4/ui_panel.py
aaronayres35: I had made this change previously but Rahul suggested we avoid changing the public api to be safe
https://github.com/enthought/traitsui/pull/1540#discussion_r599310531

traitsui/tests/test_ui_panel.py
aaronayres35: I think it implicitly will be skipped if there is no qt as modal dialog tester only exists for qt currently, but you are right we should definitely be explicit here in saying this is a very qt specific test.
Will update this now

rahulporuri (approved):
LGTM

traitsui/qt4/ui_panel.py
rahulporuri: I'm okay with making this change now. I searched through github and there seem to be no direct users of the traitsui class `HTMLHelpWindow` - https://github.com/search?p=1&q=HTMLHelpWindow+extension%3Apy&type=Code

jwiggins: Ah right. Keyword args.

aaronayres35:
CI is failing with:
```
  File "D:\a\traitsui\traitsui\.edm\envs\traitsui-test-3.6-pyqt5\lib\site-packages\traitsui\testing\tests\test_gui.py", line 154, in test_qt_process_events_process_all

```

I believe we have/had an issue open related to this intermittent failure.  Looking for issue number now

Issue #298, opened 2016-08-15
Label ignores visible_when


jonathanrocher:
In the following snippet, I would expect the label not to show, but it shows:

```
from traits.api import HasTraits, Int, Str
from traitsui.api import View, Item, Label, HGroup

class A(HasTraits):
    x = Int

    view = View(
        Label("This is a message in a group that shouldn't show.",
               visible_when="x != 0"),
        Item('x'),
        resizable=True
    )

A(x=0).configure_traits()
```

As a workaround, embedding the Label in a group makes it work. (And it also looks nicer because there is some padding around the message.)

corranwebster:
Not a known bug, and _should_ in principle work (`Label` is just a subclass of `Item`), so the fact that it doesn't is a bug.  Happens in both Wx and Qt toolkits.  Don't have time to look into it more right now.

Thanks for the report.

itziakos:
`Label` also ignores `enabled_when` for similar reasons. 

*Note: I do not thing that fixing this is easy*

tmreay:
Ran in to this issue in the TEL project. A fix would be nice, but looking how the `Label.__init__` works, I agree that it's not going to be easy. We will wrap in a `Group` as a work around.

rahulporuri:
another example to reproduce this error 

```
from traits.api import *
from traitsui.api import *

class Test(HasTraits):

    show = Bool(True)

    button = Button("Hide")

    traits_view = View(
        Label("I refuse to hide!", visible_when="show"),
        UItem("button"),
    )

    def _button_fired(self):
        self.show = False


if __name__ == "__main__":
    Test().configure_traits()
```

even if we are unable to fix this issue immediately, we should create a commonly known issues page in the documentation and mention this there.

rahulporuri:
A workaround that doesn't need the use of a `Group` wrapping the `Label` is. Thanks @pedrorivotti 

```
from traits.api import *
from traitsui.api import *

class Test(HasTraits):
    show = Bool(True)
    button = Button("Hide")
    _dummy_text = Str("I refuse to hide!")

    traits_view = View(
        UItem("_dummy_text", style="readonly", visible_when="show"),
        UItem("button"),
    )

    def _button_fired(self):
        self.show = False

if __name__ == "__main__":
    Test().configure_traits()
```

aaronayres35:
closing as this was fixed in #1544

PR #1544, opened 2021-03-19
[WIP] Label {visible/enabled}_when

aaronayres35:
addresses #298 

This PR is miles away from being ready for a review.  At this point it is basically chicken scratch to try and get something to work.  I would like to have a discussion about possible approaches for how to best solve the issue.  The current start of an implementation here I don't image will be how we wish to proceed long term.

Nonetheless, the solution here for qt only seems to work as expected, at least for the simple examples I have tried, eg. the ones described on the issue.  I effectively tried to recreate the approach used to handle {visible/enabled}_when in `traitsui.ui.py` only to match the slightly different scenario.

### DIFF
--- a/traitsui/qt4/ui_panel.py
+++ b/traitsui/qt4/ui_panel.py
@@ -29,6 +29,7 @@ import re
 from pyface.qt import QtCore, QtGui
 
 from traits.api import Any, Instance, Undefined
+from traits.observation.api import match
 
 from traitsui.api import Group
 
@@ -720,42 +721,25 @@ class _GroupPanel(object):
 
         return outer
 
-
-    def helper(self, item, label):
-        print('HEYYYYYYYYY')
-        print(item)
-        print(label)
-        print(item.visible_when)
+    def _label_when(self):
+        self._evaluate_label_condition(self._label_enabled_whens, "enabled")
+        self._evaluate_label_condition(self._label_visible_whens, "visible")
+    
+    def _evaluate_label_condition(self, conditions, kind):
         context = self.ui._get_context(self.ui.context)
 
-        try:
-            cond_value = eval(item.visible_when, globals(), context)
-            print(cond_value)
+        method_dict = {"visible": "setVisible", "enabled": "setEnabled"}
 
-            # add to update lists only if at_init is True (called on
-            # initialization), or if the editor state has to change
+        for when, label in conditions:
+            method_to_call = getattr(label, method_dict[kind])
+            try:
+                cond_value = eval(when, globals(), context)
+                method_to_call(cond_value)
+            except Exception:
+                # catch errors in the validate_when expression
+                from traitsui.api import raise_to_debug
 
-            label.setVisible(cond_value)
-
-        except Exception:
-            # catch errors in the validate_when expression
-            from traitsui.api import raise_to_debug
-
-            raise_to_debug()
-        
-        try:
-            cond_value = eval(item.enaled_when, globals(), context)
-
-            # add to update lists only if at_init is True (called on
-            # initialization), or if the editor state has to change
-
-            label.setEnabled(cond_value)
-
-        except Exception:
-            # catch errors in the validate_when expression
-            from traitsui.api import raise_to_debug
-
-            raise_to_debug()
+                raise_to_debug()
 
     def _add_items(self, content, outer=None):
         """Adds a list of Item objects, creating a layout if needed.  Return
@@ -769,6 +753,9 @@ class _GroupPanel(object):
         group = self.group
         show_left = group.show_left
         columns = group.columns
+
+        self._label_enabled_whens = []
+        self._label_visible_whens = []
 
         # See if a label is needed.
         show_labels = False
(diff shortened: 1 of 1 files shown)


rahulporuri:
orthogonal to this change, i'm really interested in refactoring the `_add_items` method to make it a lot easier to read and understand, potentially by creating other private methods which get used by `_add_items` e.g. `_add_labels` which handles the case where an item is a `Label`.

this might make for an interesting pair programming session.

corranwebster:
The layout generation code is a continuing source of problems - it is opaque, custom for each backend, and not dynamic.  The right way to fix this is to add a cross-toolkit abstraction of layout, tabs, splitters, etc. and have them controlling the layout dynamically and accessible from the main UI object.

In the meantime, do the best you can with what we have :/

aaronayres35:
I am finding it quite difficult to write a regression test for this...
Labels are not Editors (hence the solution implemented in this PR!), but that means we cant access them easily via UITester in testing.  Even further, we can manually access the `Item` objects for the labels, but those don't actually carry any reference to or info about the underlying widget.  The widget get created inside `traitsui.qt4.ui_panel._GroupPanel._add_label_item` and is held in the temporary `label` variable there.  This then gets added to the `_GroupPanel` layout by the call `self._add_widget(inner, label, row, col, show_labels)`.  
So, to access this widget in the test, we would first need to access the specific `_GoupPanel` object, and then iterate through the widgets in its layout to find the one we want to test.

I am unable to even get my hands on the correct `_GoupPanel` object.  I can get a `traitsui.qt4.ui_live._LiveWindow`, `traitsui.qt4.ui_base._StickyDialog`, `PyQt5.QtWidgets.QMainWindow`, etc. It is extremely convoluted trying to follow the logic of what all is happening in traitsUI/Qt / to find the object I am looking for in such nested class structures.

I am hopeful I'm missing something trivial and this can be done in a much simpler way. For right now I think I am going to stop looking into this as I feel like I am making no progress on something which feels like it should be so simple.
If I test manually the solution very clearly works (I can see the label enable/disable and/or toggle its visibility)

traitsui/wx/ui_panel.py
```diff
@@ -1033,6 +1069,10 @@ def add_items(self, content, panel, sizer):
             # Save the reference to the label control (if any) in the editor:
             editor.label_control = label
 
+        if (len(self._label_enabled_whens) + len(self._label_visible_whens)) > 0:
+            for object in self.ui.context.values():
+                object.on_trait_change(lambda: self._label_when(), dispatch="ui")
```
aaronayres35, line 1074: ideally we would call `self._label_when()` after this for loop, just like we do in the qt version.  The problem is, if we do this, if you run the following example:
```
from traits.api import HasTraits, Int, Str
from traitsui.api import View, Item, Label, HGroup

class A(HasTraits):
    x = Int(1)

    view = View(
        Label("Visible", visible_when="x != 0"),
        Label("Enabled", enabled_when="x != 0"),
        Item('x'),
        resizable=True
    )

if __name__ == '__main__':
    A().configure_traits()
```

you will _never_ be able to see a "Visible" label, even if you change `x` to not be 0.  (enabled works perfectly fine).  I think the reason is that, if we set the label to invisible right off the bat, it doesn't get included in the layout or something like this. Then everytime in the future, showing/hiding it does nothing. 

If we don't have `self._label_when()` (as is currently the case), then by default both the "Visible" and "Enabled" labels are shown at the start, (even though we shouldn't see "Visible" and "Enabled" should be greyed out).  From then on whenever anything is changed on the UI they will get synced up with their `enabled_when`/`visible_when` expressions and work as desired.

This is not ideal, but better than just not working at all.  I could not find a way to have the visible label be invisible to start, but then show up later if the `visible_when` condition ever got switched to True.

traitsui/qt4/ui_panel.py
```diff
@@ -733,6 +761,9 @@ def _add_items(self, content, outer=None):
         show_left = group.show_left
         columns = group.columns
 
+        self._label_enabled_whens = []
+        self._label_visible_whens = []
```
rahulporuri, line 765: let's just make these private `List` traits on the class instead?

traitsui/wx/ui_panel.py
```diff
@@ -742,6 +769,10 @@ def add_items(self, content, panel, sizer):
         col = -1
         col_incr = 1
         self.label_flags = 0
+
+        self._label_enabled_whens = []
```
rahulporuri, line 773: looks like we should make the changes made in https://github.com/enthought/traitsui/pull/1549 to qt to wx as well.

rahulporuri (approved):
LGTM. I tested this on qt with the MRE in the issue and this solution works as expected. 

Tangentially, it looks like we can continue refactoring the toolkit-specific `ui_panel` classes.

traitsui/wx/ui_panel.py
rahulporuri: let's convert this into an issue

traitsui/wx/ui_panel.py
aaronayres35: I've opened #1592

traitsui/qt4/ui_panel.py
aaronayres35: It looks like `_GroupPanel` has `object` as a base class / is not a `HasTraits` class

traitsui/wx/ui_panel.py
aaronayres35: I've opened #1593

PR #1546, opened 2021-03-19
manually set the text format to plain text in error

### LINKED ISSUE #1543: Hidden error message parts in Qt
The `error` implementation of qt4 Editor class passes the raw exception string to the QMessageBox:

https://github.com/enthought/traitsui/blob/7a6d46ad10ea6c6b5a8dcc3b4939a9670c2a52ad/traitsui/qt4/editor.py#L72-L83

I think that it interprets the message string as rich text by default, rather than plain text. The result is that if the error message includes HTML-like syntax, then things can go missing. This can happen with the `RangeEditor` as in this small example:

```python

class Foo(HasTraits):
    x = Range(low=0.0, high=1.0, value=0.5, exclude_low=True)

Foo().configure_traits()
```

If you move the value all the way to 0.0, then the true exception string is:

> The 'x' trait of a Foo instance must be 0.0 < a floating point number <= 1.0, but a value of 0.0 <class 'float'> was specified.

But the following dialog is shown:

image

You can see that the portions of text enclosed in `<>` go poof.

aaronayres35:
fixes #1543

Note I am observing a segfault when I set the slider to 0 in the example given in the issue and then close the information dialog (this occurs both with and without the changes of this PR)
Im not sure why we weren't previously using `information` from the `pyface.api` here (see: https://github.com/enthought/pyface/blob/65813858730a61f25618e98669bcc5347d44935a/pyface/message_dialog.py#L15-L43)

In any case, now in order to call `setTextFormat` we would not be ably to use pyface directly anyway / we can no longer call the static function `QtGui.QMessageBox.information`.

This is what I currently see if I set the slider to 0.0:
Screen Shot 2021-03-19 at 3 19 51 PM

As opposed to this from before:

Screen Shot 2021-03-19 at 3 21 01 PM

I am unsure why 2 dialogs show up (this is probably a separate bug?)

A test for this would likely require UITester + ModalDialogTester to look at the displayed text in the created modal dialog. I will investigate this now

### DIFF
--- a/traitsui/tests/test_editor.py
+++ b/traitsui/tests/test_editor.py
@@ -937,7 +937,10 @@ class TestEditor(BaseTestMixin, GuiTestAssistant, unittest.TestCase):
     # regression test for enthought/traitsui#1543
     @requires_toolkit([ToolkitName.qt])
     @unittest.skipIf(no_modal_dialog_tester, "ModalDialogTester unavailable")
-    @unittest.skipIf(is_mac_os, "There is a separate issue on OSX")
+    @unittest.skipIf(
+        is_mac_os,
+        "There is a separate issue on OSX. See enthought/traitsui#1550"
+    )
     def test_editor_error_msg(self):
         from pyface.qt import QtGui
 
--- a/traitsui/tests/test_editor.py
+++ b/traitsui/tests/test_editor.py
@@ -921,6 +921,7 @@ class TestEditor(BaseTestMixin, GuiTestAssistant, unittest.TestCase):
     def test_editor_error_msg(self):
         from pyface.toolkit import toolkit_object
         from pyface.constant import OK
+        from pyface.qt import QtGui
         from traits.api import HasTraits, Range
 
         from traitsui.testing.api import (
@@ -949,7 +950,25 @@ class TestEditor(BaseTestMixin, GuiTestAssistant, unittest.TestCase):
             def trigger_error():
                 x_range_textbox.perform(KeyClick('Enter'))
 
+            def check_and_close(mdtester):
+                try:
+                    with mdtester.capture_error():
+                        print(mdtester.get_dialog_widget())
+                        self.assertTrue(
+                            mdtester.has_widget(
+                                text="The 'x' trait of a Foo instance must be "
+                                     "0.0 < a floating point number <= 1.0, "
+                                     "but a value of 0.0 <class 'float'> was "
+                                     "specified.",
+                                type_=QtGui.QMessageBox,
+                            )
+                        )
+
+                finally:
+                    mdtester.close(accept=True)
+                    self.assertTrue(mdtester.dialog_was_opened)
+
 
             mdtester = ModalDialogTester(trigger_error)
-            mdtester.open_and_run(lambda x: x.click_button(OK))
+            mdtester.open_and_run(check_and_close)
             self.assertTrue(mdtester.dialog_was_opened)
--- a/traitsui/tests/test_editor.py
+++ b/traitsui/tests/test_editor.py
@@ -10,7 +10,10 @@
 
 import unittest
 
-from traits.api import Any, Bool, Event, Float, HasTraits, Int, List, Undefined
+from pyface.toolkit import toolkit_object
+from traits.api import (
+    Any, Bool, Event, Float, HasTraits, Int, List, Range, Undefined
+)
 from traits.trait_base import xgetattr
 
 from traitsui.context_value import ContextValue, CVFloat, CVInt
@@ -18,9 +21,22 @@ from traitsui.editor import Editor
 from traitsui.editor_factory import EditorFactory
 from traitsui.handler import default_handler
 from traitsui.ui import UI
-from traitsui.tests._tools import (
-    BaseTestMixin, GuiTestAssistant, no_gui_test_assistant
+from traitsui.testing.api import (
+    KeyClick, KeySequence, Textbox, UITester
 )
+from traitsui.tests._tools import (
+    BaseTestMixin,
+    GuiTestAssistant,
+    is_mac_os,
+    no_gui_test_assistant,
+    requires_toolkit,
+    ToolkitName,
+)
+
+ModalDialogTester = toolkit_object(
+    "util.modal_dialog_tester:ModalDialogTester"
+)
+no_modal_dialog_tester = ModalDialogTester.__name__ == "Unimplemented"
 
 
 class FakeControl(HasTraits):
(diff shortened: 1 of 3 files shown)


aaronayres35:
I have the bones set up for a test, but the test hits the segfault I describe above.  This will likely need some other fix

rahulporuri:
> Note I am observing a segfault when I set the slider to 0 in the example given in the issue and then close the information dialog (this occurs both with and without the changes of this PR)

I'm not seeing this on windows with pyqt5. What toolkit are you seeing that behavior with? And if the segfault is persistent, there is something horrible going on so that needs a separate issue.

rahulporuri:
>  Im not sure why we weren't previously using information from the pyface.api here (see: https://github.com/enthought/pyface/blob/65813858730a61f25618e98669bcc5347d44935a/pyface/message_dialog.py#L15-L43)

I honestly dont know why I bothered but I dug into the code to realize that `pyface.api.information` isnt used here simply because it was added in 2008 (https://github.com/enthought/pyface/commit/7aa4fb96539fb6b9d79cefd5231fec06a7d898b8) whereas this module was added in 2007 (https://github.com/enthought/traitsui/commit/266a874d02eb7c64967c746d548eaa61a83557a8#diff-e282d9e0c90155c775f3886e3bd7b8d7d4f9f2a9d1b2ca7f64547d7c8a898646).

It looks like we might actually be able to replace the `QMessageBox` with a call to `information`.

rahulporuri:
> I am unsure why 2 dialogs show up (this is probably a separate bug?)

I'm not seeing this either. Your fix works for me on windows with python 3.6 and pyqt5 and I just see the one dialog with the correct/expected text.

aaronayres35:
> > Im not sure why we weren't previously using information from the pyface.api here (see: https://github.com/enthought/pyface/blob/65813858730a61f25618e98669bcc5347d44935a/pyface/message_dialog.py#L15-L43)
> 
> I honestly dont know why I bothered but I dug into the code to realize that `pyface.api.information` isnt used here simply because it was added in 2008 ([enthought/pyface@7aa4fb9](https://github.com/enthought/pyface/commit/7aa4fb96539fb6b9d79cefd5231fec06a7d898b8)) whereas this module was added in 2007 ([266a874#diff-e282d9e0c90155c775f3886e3bd7b8d7d4f9f2a9d1b2ca7f64547d7c8a898646](https://github.com/enthought/traitsui/commit/266a874d02eb7c64967c746d548eaa61a83557a8#diff-e282d9e0c90155c775f3886e3bd7b8d7d4f9f2a9d1b2ca7f64547d7c8a898646)).
> 
> It looks like we might actually be able to replace the `QMessageBox` with a call to `information`.

We cant easily use `information` or `MessageDialog` form pyface as we need direct access to the control that gets created in `_create_control`, which is not called until the dialog is open.  But we want to set the text format before the dialog gets opened.  This should be an option in pyface and I will open a separate PR there to allow specifying the text formatting option in `information` / `MessageDialog`

aaronayres35:
> > I am unsure why 2 dialogs show up (this is probably a separate bug?)
> 
> I'm not seeing this either. Your fix works for me on windows with python 3.6 and pyqt5 and I just see the one dialog with the correct/expected text.

I tried locally with Ubuntu and pyqt5 and everything is exactly as I would expect.  It seems to be a macos specific thing

On macOS, I git cleaned and create a fresh edm env and ran an example on master.  I still see 2 dialogs and the segfault when trying to close one of them.  I believe there is a bug here (potentially at the qt level) with macOS that is unrelated to the changes made in this PR.  I will open an issue, but I am unsure what to do with this PR for the test

aaronayres35:
I pushed hoping to see if the macOS CI would have the same issue I was seeing.  Unfortunately, the appveyor build has no logs (although it did fail).

Seemingly orthogonally, the Ubuntu pyside2 GH action is failing on a completely different test (????):
```
est_base_url_changed (traitsui.tests.editors.test_html_editor.TestHTMLEditor) ... Command '['edm', 'run', '-e', 'traitsui-test-3.6-pyside2', '--', 'python', '-W', 'default', '-m', 'coverage', 'run', '-p', '-m', 'unittest', 'discover', '-v', 'traitsui']' returned non-zero exit status 252.
No xvfb processes to kill
```

traitsui/tests/test_editor.py
```diff
@@ -917,3 +917,54 @@ def test_sync_value_both(self):
...
+                                type_=QtGui.QMessageBox,
+                            )
+                        )
+                finally:
+                    mdtester.close(accept=True)
+                    self.assertTrue(mdtester.dialog_was_opened)
+
+            mdtester = ModalDialogTester(trigger_error)
+            mdtester.open_and_run(check_and_close)
+            self.assertTrue(mdtester.dialog_was_opened)
```
aaronayres35, line 970: The level of complexity for this test now seems to far exceed what it is trying to test :/
Perhaps the test could be trivia and just reach into Qt and make sure the `textFormat` property gets set? Like I do in https://github.com/enthought/pyface/pull/907

The macOS weirdness I am seeing feels like an orthogonal problem, and I am still very unclear on what the cause is.

jwiggins: > The macOS weirdness I am seeing feels like an orthogonal problem, and I am still very unclear on what the cause is.

Yes. Definitely something strange going on here.

jwiggins:
Checking the text of the dialog in the test seems OK to me (or is that not working consistently?).

I think the bigger issue is the Qt-specific nature of the test code. I know we're not running Wx CI jobs, but it still feels a bit wrong to have a bunch of local imports and Qt specific code in a single test. Do you plan to remove the local imports before merging this?

aaronayres35:
> Checking the text of the dialog in the test seems OK to me (or is that not working consistently?).
> 
> I think the bigger issue is the Qt-specific nature of the test code. I know we're not running Wx CI jobs, but it still feels a bit wrong to have a bunch of local imports and Qt specific code in a single test. Do you plan to remove the local imports before merging this?

Yeah I will clean up the imports and specify the test requires Qt specifically before merging.
But currently there is something bad happening on OSX.  If you run the example are you seeing the double dialogs / possible segfault when you try to close them locally?

jwiggins:
Yes, the example from the issue causes a segfault when I run it on macOS. I believe that's a separate issue.

aaronayres35:
> Yes, the example from the issue causes a segfault when I run it on macOS. I believe that's a separate issue.

Thank you very much for confirming that!  I think I will just skip this test on OSX on CI then and open a separate issue

aaronayres35:
CI is still failing on Ubuntu/Pyside2 in a seemingly unrelated test...
I am unable to reproduce the test failure locally (on both OSX and Ubuntu)

When I run the relevant test module thought I do see the following:
"Release of profile requested but WebEnginePage still not deleted. Expect troubles !"

traitsui/tests/test_editor.py
```diff
@@ -917,3 +933,48 @@ def test_sync_value_both(self):
 
         with self.assertTraitDoesNotChange(user_object, "user_auxiliary"):
             editor.auxiliary_value = 14
+
+    # regression test for enthought/traitsui#1543
+    @requires_toolkit([ToolkitName.qt])
+    @unittest.skipIf(no_modal_dialog_tester, "ModalDialogTester unavailable")
+    @unittest.skipIf(is_mac_os, "There is a separate issue on OSX")
```
jwiggins, line 940: Do we have an issue # to reference here?

jwiggins:
Given all the crazy failures you're seeing, now might be a good time to add `PYTHONFAULTHANDLER=1` to the test environment.

jwiggins (commented):
Changes look OK. Anything left here?

aaronayres35:
> Changes look OK. Anything left here?

I think we are good to go?  I am confused why the `test_base_url_changed (traitsui.tests.editors.test_html_editor.TestHTMLEditor)` test failure occurred and also confused why it suddenly went away... 🤔 
It had occurred in multiple previous CI runs and somehow has disappeared now without any attempt at a fix

jwiggins:
Oooh boy. Intermittent failures are the best, huh?

We should probably just leave an issue and move on.

aaronayres35:
> Oooh boy. Intermittent failures are the best, huh?
> 
> We should probably just leave an issue and move on.

I've opened #1551

jwiggins (approved):

PR #1549, opened 2021-03-23
Refactor _add_items method of _GroupPanel object

aaronayres35:
ref: https://github.com/enthought/traitsui/pull/1544#issuecomment-804545226


rahulporuri (approved):
Ignoring the WIP in the PR title, these changes LGTM.

The only comment I wish I had made when we were originally working on this PR is - it would have been awesome if we had separated each refactoring step into a single commit. That would have made it abundantly clear to a reviewer that we're just moving code around.

PR #1553, opened 2021-03-24
Update new test_editor_error_msg

### LINKED ISSUE #1551: Intermittent failure on test_base_url_changed in traitsui.tests.editors.test_html_editor.py
Observed in #1546 

From the CI logs:
```
test_base_url_changed (traitsui.tests.editors.test_html_editor.TestHTMLEditor) ... Command '['edm', 'run', '-e', 'traitsui-test-3.6-pyside2', '--', 'python', '-W', 'default', '-m', 'coverage', 'run', '-p', '-m', 'unittest', 'discover', '-v', 'traitsui']' returned non-zero exit status 252.
Killing the following xvfb processes: 5529
```

after enabling faulthandler in tests this failure was no longer observed.  I was unable to reproduce failure locally on MacOS or Ubuntu (failure was occurring in Ubuntu/Pyside2 CI run).

I apologize for the vague issue, I do not currently have a good understanding of what occurred or how to easily reproduce it, but didn't want it to go untracked entirely.

aaronayres35:
_Hopefully_ fixes #1551 

The intermittent test failure first appeared in pr #1546, and after my realization on #1540 I am suspect of `auto_process_events` on UITester being relevant here as well (although this is still more of a hunch than a well understood solution).  This PR sets auto_process_events=False (the test still runs as expected with ModalDialogTester taking the full event processing responsibility, so even if this was not the cause of 1551 this change is still viable).

In addition, looking back at the test from 1546 and looking more closely at the code in ModalDialogTester, I realize that the `text` arg in `has_widget` corresponds to the text the widget holds which is the same value regardless of the format (it was not corresponding to the text actually displayed by the widget) so the previous test wasn't really hitting the core issue.  I added an assert statement explicitly checking the `textFormat` matches what it was set to.

### DIFF
--- a/traitsui/tests/test_editor.py
+++ b/traitsui/tests/test_editor.py
@@ -942,13 +942,13 @@ class TestEditor(BaseTestMixin, GuiTestAssistant, unittest.TestCase):
         "There is a separate issue on OSX. See enthought/traitsui#1550"
     )
     def test_editor_error_msg(self):
-        from pyface.qt import QtGui
+        from pyface.qt import QtCore, QtGui
 
         class Foo(HasTraits):
             x = Range(low=0.0, high=1.0, value=0.5, exclude_low=True)
 
         foo = Foo()
-        tester = UITester()
+        tester = UITester(auto_process_events=False)
         with tester.create_ui(foo) as ui:
 
             x_range = tester.find_by_name(ui, "x")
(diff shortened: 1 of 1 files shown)


aaronayres35:
#1551 is not occurring which is a good sign! Doesn't guarantee it is resolved though as it did not occur on every run in 1546

rahulporuri (approved):
LGTM

PR #1554, opened 2021-03-25
Fix RGBColor hex int to tuple color conversion

### LINKED ISSUE #1531: RGBColor does not correctly convert hexadecimal integer to a tuple of floats
When I input an integer in 0xRRGGBB format to the RGBColor trait, I get the following error:

`... , line 59, in convert_to_color
    ((num / 0x100) & 0xFF) / 255.0,
TypeError: unsupported operand type(s) for &: 'float' and 'int'`

I'm still using TraitsUI 7.0.1, but I think the issue is still relevant. I think this is due to Python converting the result of the first division to a float, and this might be solved by replacing the normal division by an integer division (`//`):

https://github.com/enthought/traitsui/blob/bffbb3b518c57ac1a4178b60b942f0f6a9831cd2/traitsui/null/rgb_color_trait.py#L52

aaronayres35:
fixes #1531

This PR simply replaces `/` with `//` to avoid trying to `&` a float with an int.  Note the wx version already had this.  It also adds a very simple regression test

Note that as mentioned https://github.com/enthought/traitsui/issues/1531#issuecomment-790573071 this function is effectively copied 3 separate times and we will ultimately want to unify them in one place.  However, the wx version has an additional check for `wx.Colour` inputs, so they are not exactly the same.  This very simple PR fixes the bug, the architecture / code organization issue can be addressed separately.

### DIFF
--- a/traitsui/null/rgb_color_trait.py
+++ b/traitsui/null/rgb_color_trait.py
@@ -49,7 +49,7 @@ def convert_to_color(object, name, value):
         num = int(value)
         return (
             (num / 0x10000) / 255.0,
-            ((num / 0x100) & 0xFF) / 255.0,
+            ((num // 0x100) & 0xFF) / 255.0,
             (num & 0xFF) / 255.0,
         )
     raise TraitError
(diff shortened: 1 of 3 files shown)


rahulporuri (approved):
LGTM

PR #1552, opened 2021-03-24
Add 'IsVisible` query class for UI Tester

aaronayres35:
closes #1291 

Now that #1490 has been merged this process was quite simple :)

This PR makes it so that now any UIWrapper object wrapping a target with a control that is either a QtWidget or wx.Window can now support an `IsVisible` query. A comparable thing was done for `IsEnabled` in 1490.

This PR also updates the relevant tests for `_control_widget_registry` and adds a test for the `visible_when.py` demo.

PR #1555, opened 2021-03-25
Make Scrollable group respect visible_when

### LINKED ISSUE #1512: Scrollable Group cannot become non-visible
When a `Group` is `scrollable=True` and its `visible_when` depends on a dynamic property, the scrollability prevents the group from becoming non-visible.

An example code:
```python
from traits.api import *
from traitsui.api import *
​
class Foo(HasTraits):
    bar = Str("bar!")
​
    baz = Str("Baz?")
​
    enabled = Enum("Yes", "No")
​
        view = View(
            Item("enabled"),
            HGroup(
                VGroup(
                    Item("bar"),
                    scrollable=True,
                    visible_when="enabled=='Yes'",
                ),
                VGroup(
                    Item("baz"),
                    scrollable=True,
                    visible_when="enabled=='No'",
                ),
            ),
        )
        return view
​
if __name__=="__main__":
    f = Foo()
    f.configure_traits()
```

I would expect the only _one_ group is visible at a time (because the `visible_when` conditions are mutually exclusive). However, the UI always allocated some space for both groups:
Screenshot 2021-02-03 at 17 15 23

Screenshot 2021-02-03 at 17 15 45

Two workarounds:
- Make the groups `scrollable=False` (not always an option)
- Wrap each group into yet one more layer, and move the visibility to the outer one:
```python
                VGroup(
                    VGroup(
                        Item("bar"),
                        scrollable=True,
                    ),
                    visible_when="enabled=='Yes'",
                ),
```

aaronayres35:
Note: this PR sits on top of #1552 as it uses `IsVisible` to write a regression test.  1552 should be merged first, and any changes from PR review suggestions should be merged in here.

Currently the diff on this PR is ugly as a result.  The only relevant files being changed here are `traitsui/qt4/ui_panel.py` and `traitsui/qt4/tests/test_ui_panel.py`.

fixes #1512 

The problem code was setting up an editor to hook up the `enabled_when` and `visible_when` checks for the group using the control for the contained widget.  In reality, if we had a scrollable group, we want these to get hooked up for the QScrollArea containing that widget not just the widget itself.  Simply moving the conditional for handling scrollable groups before handling the group's enabled/visible when checks solves this problem.

### DIFF
--- a/.github/workflows/test-etsdemo.yml
+++ b/.github/workflows/test-etsdemo.yml
@@ -8,7 +8,7 @@ on:
     - 'ets-demo/**'
 
 env:
-  INSTALL_EDM_VERSION: 3.2.1
+  INSTALL_EDM_VERSION: 3.2.3
 
 jobs:
 
--- a/.github/workflows/test-with-edm.yml
+++ b/.github/workflows/test-with-edm.yml
@@ -11,7 +11,7 @@ name: Test with EDM
 on: pull_request
 
 env:
-  INSTALL_EDM_VERSION: 3.2.1
+  INSTALL_EDM_VERSION: 3.2.3
 
 jobs:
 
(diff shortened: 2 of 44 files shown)


rahulporuri (approved):
LGTM. That was quick :D. I tested this fix with the code in #1512 and it works as expected now.

PR #1557, opened 2021-03-25
Document `entries` parameter not supported qt for FileEditor/DirectoryEditor

### LINKED ISSUE #1530: File / FileEditor no history (`entries` option not working)
```python
class Test(HasTraits):
    p = File(entries=10, dialog_style='open')
```
```python
class Test(HasTraits):
    p = File(editor=FileEditor(entries=10, dialog_style='open'))
```

Neither shows the history combo box with `entries=10` provided.
Then found out the source code of `traitsui.editors.file_editor.FileEditor` says:
```python
    #: FIXME: add support
    entries = Int(10)
```
So does this mean `entries` option is actually not supported yet? Then why does the official document shows examples of `FileEditor` with history combo box?

aaronayres35:
closes #1530 
Although we may want to have another issue open for adding qt support

This PR simply aims to better document the fact that the `entries` parameter for keeping a history of past entries on the FileEditor (and subsequently the DirectoryEditor) is only supported on the wx backend, but not on qt.  

In addition, this PR makes a fix (similar to the one made in #1067) to fix the `History_demo.py` on wx.  Now if one runs the demo in an environment with wx they can see what the history functionality is actually capable of.

### DIFF
--- a/docs/source/traitsui_user_manual/factories_basic.rst
+++ b/docs/source/traitsui_user_manual/factories_basic.rst
@@ -591,10 +591,11 @@ the local system hierarchy. The styles of this editor are shown in Figure 31.
 
 The default version of the simply style displays a text box and a :guilabel:`Browse`
 button. Clicking :guilabel:`Browse` opens a platform-specific file selection dialog box.
-If you specify the *entries* keyword parameter with an integer value to the
+On the wx backend, if you specify the *entries* keyword parameter with an integer value to the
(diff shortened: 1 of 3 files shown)


rahulporuri (approved):
LGTM but we definitely need a separate issue regarding implementing this in qt.

aaronayres35:
> LGTM but we definitely need a separate issue regarding implementing this in qt.

I've opened #1558

PR #1559, opened 2021-03-26
Dont crash when failing to add item to list

### LINKED ISSUE #1154: List(Directory(exists=True)) causes Aborted (core dumped)
I notice this using traits v6.1.1 and traitsui v7.0.1 on Ubuntu. I have not checked other platforms.

Here is code to reproduce:
```python
from traits.api import Directory, HasTraits, List

class Foo(HasTraits):
    dirs = List(Directory(exists=True))

Foo().configure_traits()
```
Triggering the 'Add' action from the list editor hamburger menu causes an abort.

aaronayres35:
closes #1154 

This PR currently catches the TraitError to prevent the app from crashing.  If you run the short example given in the issue now and add a new item, you see an error dialog 
Screen Shot 2021-03-26 at 1 57 17 PM
like before, but clicking ok just returns you to thee original UI, without crashing.  It also adds a regression test.

There may be other changes we want to make down the road, but this is a good first step which resolves the major problem of the issue.

### DIFF
--- a/traitsui/qt4/list_editor.py
+++ b/traitsui/qt4/list_editor.py
@@ -28,7 +28,7 @@ from pyface.qt import QtCore, QtGui
 
 from pyface.api import ImageResource
 
-from traits.api import Str, Any, Bool, Dict, Instance, List
+from traits.api import Str, Any, Bool, Dict, Instance, List, TraitError
 from traits.trait_base import user_name_for, xgetattr
 
 # FIXME: ToolkitEditorFactory is a proxy class defined here just for backward
@@ -311,7 +311,12 @@ class SimpleEditor(Editor):
         index += offset
         item_trait = self._trait_handler.item_trait
         value = item_trait.default_value_for(self.object, self.name)
-        self.value = list[:index] + [value] + list[index:]
+        try:
+            self.value = list[:index] + [value] + list[index:]
+        # if the default new item is invalid, we just don't add it to the list.
+        # traits will still give an error message, but we don't want to crash
+        except TraitError:
+            pass
         self.update_editor()
 
     def add_before(self):
--- a/traitsui/qt4/list_editor.py
+++ b/traitsui/qt4/list_editor.py
@@ -316,7 +316,8 @@ class SimpleEditor(Editor):
         # if the default new item is invalid, we just don't add it to the list.
         # traits will still give an error message, but we don't want to crash
         except TraitError:
-            pass
+            from traitsui.api import raise_to_debug
+            raise_to_debug()
         self.update_editor()
 
     def add_before(self):
--- a/traitsui/tests/editors/test_list_editor.py
+++ b/traitsui/tests/editors/test_list_editor.py
@@ -11,7 +11,9 @@
 import unittest
 
 from pyface.toolkit import toolkit_object
-from traits.api import Directory, HasStrictTraits, Instance, Int, List, Str
+from traits.api import (
+    Directory, HasStrictTraits, Instance, Int, List, Str, TraitError
+)
 
 from traitsui.api import Item, ListEditor, View
 from traitsui.testing.api import (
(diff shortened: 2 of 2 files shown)


traitsui/tests/editors/test_list_editor.py
```diff
@@ -170,6 +177,36 @@ def test_index_out_of_range(self):
...
+
+            mdtester = ModalDialogTester(trigger_error)
+            mdtester.open_and_run(lambda x: x.close())
+            # we want an error dialog to open, but don't want to raise a
+            # TraitError and crash the full application
+            self.assertTrue(mdtester.dialog_was_opened)
+            self.assertTrue(mdtester.result)
+
+            self.assertEqual(len(obj.dirs), 0)
+
```
aaronayres35, line 209: This test is a little ugly but it ultimately does what I wanted it to.

The behavior this PR capitalizes on is that if a trait is set to an invalid value, https://github.com/enthought/traitsui/blob/1c880889f19abbda50deddd497f7a24721b3a13a/traitsui/qt4/editor.py#L72-L90 will be called.  This is good as it gives an error dialog to the user saying why they can't do what they just did.  However, we don't really want this to crash the application if it happens when trying to add a new item to the list, so we swallow the exception there, and the list will just be left unchanged to how it was before.

traitsui/qt4/list_editor.py
```diff
@@ -311,7 +311,12 @@ def add_item(self, offset):
         index += offset
         item_trait = self._trait_handler.item_trait
         value = item_trait.default_value_for(self.object, self.name)
-        self.value = list[:index] + [value] + list[index:]
+        try:
+            self.value = list[:index] + [value] + list[index:]
+        # if the default new item is invalid, we just don't add it to the list.
+        # traits will still give an error message, but we don't want to crash
+        except TraitError:
+            pass
```
aaronayres35, line 319: It feels not ideal to just swallow the exception and leaving the list as it was before, but that feels better than crashing entirely

corranwebster: Yes, use `raise_to_debug`.  We should probably using that more often.

traitsui/qt4/list_editor.py
```diff
@@ -311,7 +311,12 @@ def add_item(self, offset):
         index += offset
         item_trait = self._trait_handler.item_trait
         value = item_trait.default_value_for(self.object, self.name)
```
corranwebster, line 313: This feels like the real problem.  We are relying on the trait to give us the default value, which is great when it works, but it is possible for the default value to _not_ validate for various reasons.  Having this happen is effectively a programmer error outside our control from our point of view - the trait doesn't have a valid default!  So crashing is OK, but it can't be a hard crash.

This PR fixes the hard crash, which is good.

But there is a secondary fix, which is that the buggy code may be outside of the control of the person using the editor, so we need _another_ way of specifying the default value that doesn't rely on the trait.  It feels like the list editor should grow a new "default_factory" `Callable` trait that, if not `None`, is called instead of this.

traitsui/tests/editors/test_list_editor.py
```diff
@@ -170,6 +179,35 @@ def test_index_out_of_range(self):
             with self.assertRaises(IndexError):
                 people_list.locate(Index(10))
 
+    # regression test for enthought/traitsui#1154
+    @requires_toolkit([ToolkitName.qt])
+    @unittest.skipIf(no_modal_dialog_tester, "ModalDialogTester unavailable")
```
rahulporuri, line 184: I'm not entirely sure why the modal dialog tester would be unavailable with the qt toolkit.

aaronayres35: AFAIK It wouldn't be, I can remove this

rahulporuri (approved):
LGTM

traitsui/qt4/list_editor.py
rahulporuri: looks like we need to open a separate issue for ^

aaronayres35: I'v opened #1627

PR #1571, opened 2021-04-06
Support testing with simple FileEditor (without dialog)

kitchoi:
This PR adds testing support for the simple FileEditor via its text widget.

- `DisplayedText`, `KeyClick`, `KeySequence` can now be used for testing with a simple FileEditor
- A test for the FileEditor demo example is added, and an entry is added to the documentation.

This is motivated by an observation of a similar extension being made in a downstream project. One will extend using the `_file_name` attribute on the editor, acknowledging that such extension relies on implementation details of the editor. If the testing support is implemented in traitsui, then the support can be maintained along with those implementation details.

In the context of testing an application, it is often sufficient to set a file path via the text widget instead of via the file modal dialog.

This goes some way towards solving #1339. It also extends TraitsUI's own test coverage for the simple FileEditor.

Observations:
- The smoke tests for wx FileEditor used to fail, but not any more since #889 was solved.
- The `entries` trait on the FileEditor factory is only supported by wx. If set on wx, the simple FileEditor will have a combo box instead of a text widget. The simple FileEditor on wx is effectively two editors in one. That branch is not tested here.
- On the demo, if one sets an existing file path for the simple FileEditor, the custom FileEditor will normalize the file path.  So if one sets a relative path to the simple FileEditor, they will get an absolute path back unless the custom FileEditor is removed. That's because in the demo, multiple editors are tied to the same trait.
- The test for the FileEditor demo fails on wx due to #752. The presence of the custom FileEditor caused it. If the custom FileEditor was removed, the test will pass for wx.

### DIFF
--- a/traitsui/examples/demo/Standard_Editors/tests/test_FileEditor_demo.py
+++ b/traitsui/examples/demo/Standard_Editors/tests/test_FileEditor_demo.py
@@ -1,0 +1,61 @@
+# (C) Copyright 2004-2021 Enthought, Inc., Austin, TX
+# All rights reserved.
+#
+# This software is provided without warranty under the terms of the BSD
+# license included in LICENSE.txt and may be redistributed only under
+# the conditions described in the aforementioned license. The license
+# is also available online at http://www.enthought.com/licenses/BSD.txt
+#
+# Thanks for using Enthought open source!
+
+"""
+This example demonstrates how to test interacting with a textbox created
+using FileEditor.
+
+The GUI being tested is written in the demo under the same name (minus the
+preceding 'test') in the outer directory.
+"""
+
+import os
+import runpy
+import unittest
+
+from traitsui.testing.api import (
+    DisplayedText, KeyClick, KeySequence,  UITester
+)
+
+#: Filename of the demo script
+FILENAME = "FileEditor_demo.py"
+
+#: Path of the demo script
+DEMO_PATH = os.path.join(os.path.dirname(__file__), "..", FILENAME)
+
+
+class TestFileEditorDemo(unittest.TestCase):
+
+    def test_run_demo(self):
+        demo = runpy.run_path(DEMO_PATH)["demo"]
+
+        tester = UITester()
+        with tester.create_ui(demo) as ui:
+            # simple FileEditor
+            simple_file_field = tester.find_by_id(ui, "simple_file")
+
+            # Modify the value on the GUI
+            # The path does not exist, whic is allowed by the trait.
+            # The custom FileEditor will ignore nonexisting file path.
+            simple_file_field.perform(KeySequence("some_path"))
+            simple_file_field.perform(KeyClick("Enter"))
+            self.assertEqual(demo.file_name, "some_path")
+
+            # Modify the value on the object
+            demo.file_name = FILENAME
+            self.assertEqual(
+                simple_file_field.inspect(DisplayedText()), FILENAME
+            )
+
+
+# Run the test(s)
+unittest.TextTestRunner().run(
+    unittest.TestLoader().loadTestsFromTestCase(TestFileEditorDemo)
+)
--- a/traitsui/examples/demo/Standard_Editors/tests/test_FileEditor_demo.py
+++ b/traitsui/examples/demo/Standard_Editors/tests/test_FileEditor_demo.py
@@ -54,6 +54,10 @@ class TestFileEditorDemo(unittest.TestCase):
                 simple_file_field.inspect(DisplayedText()), FILENAME
             )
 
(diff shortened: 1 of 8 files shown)


kitchoi:
Note: One can apply more or less the same change for the simple DirectoryEditor

traitsui/examples/demo/Standard_Editors/tests/test_FileEditor_demo.py
```diff
@@ -0,0 +1,61 @@
...
+    def test_run_demo(self):
+        demo = runpy.run_path(DEMO_PATH)["demo"]
+
+        tester = UITester()
+        with tester.create_ui(demo) as ui:
+            # simple FileEditor
+            simple_file_field = tester.find_by_id(ui, "simple_file")
+
+            # Modify the value on the GUI
+            # The path does not exist, whic is allowed by the trait.
```
aaronayres35, line 45: nit: "which"

traitsui/examples/demo/Standard_Editors/tests/test_FileEditor_demo.py
```diff
@@ -0,0 +1,61 @@
...
+#: Filename of the demo script
+FILENAME = "FileEditor_demo.py"
+
+#: Path of the demo script
+DEMO_PATH = os.path.join(os.path.dirname(__file__), "..", FILENAME)
+
+
+class TestFileEditorDemo(unittest.TestCase):
+
+    def test_run_demo(self):
```
aaronayres35, line 36: You say in the PR description: "The test for the FileEditor demo fails on wx due to #752".  Should this have an `@requires_toolkit(['ToolkitName.qt'])` in that case?
/ a comment regarding the failure being linked to having a custom editor included in the demo?

traitsui/tests/editors/test_file_editor.py
```diff
@@ -27,33 +26,95 @@ class FileModel(HasTraits):
...
+class TestSimpleFileEditor(BaseTestMixin, unittest.TestCase):
+    """ Test FileEditor (simple style). """
+
+    def setUp(self):
+        BaseTestMixin.setUp(self)
+
+    def tearDown(self):
+        BaseTestMixin.tearDown(self)
+
+    def test_simple_editor_set_text_to_nonexisting_path(self):
```
aaronayres35, line 42: From the Travis CI logs it looks like this test and `test_simple_editor_auto_set_text` are failing on wx
<details>

```
FAIL: test_simple_editor_auto_set_text (traitsui.tests.editors.test_file_editor.TestSimpleFileEditor)
  File "/home/travis/.edm/envs/traitsui-test-3.6-wx/lib/python3.6/site-packages/traitsui/tests/editors/test_file_editor.py", line 78, in test_simple_editor_auto_set_text
    self.assertEqual(obj.filepath, "some_file.txt")
AssertionError: 'ome_file.txts' != 'some_file.txt'
- ome_file.txts
?             -
+ some_file.txt
? +
FAIL: test_simple_editor_set_text_to_nonexisting_path (traitsui.tests.editors.test_file_editor.TestSimpleFileEditor)
  File "/home/travis/.edm/envs/traitsui-test-3.6-wx/lib/python3.6/site-packages/traitsui/tests/editors/test_file_editor.py", line 54, in test_simple_editor_set_text_to_nonexisting_path
    self.assertEqual(obj.filepath, "some_file.txt")
AssertionError: 'ome_file.txt' != 'some_file.txt'
- ome_file.txt
+ some_file.txt
? +
```

</details>
perhaps we skip them on wx and leave a comment?  They are allowed to fail, so CI is still green

aaronayres35 (approved):
LGTM

Just a couple comments about skipping tests on wx

traitsui/examples/demo/Standard_Editors/tests/test_FileEditor_demo.py
kitchoi: Good point. Yes I should add the decorator.

kitchoi: > Yes I should add the decorator.

On second thought, this is a demo test so I'd want to avoid importing anything from `traitsui.tests._tools`.
Will go with a comment instead :/

traitsui/tests/editors/test_file_editor.py
kitchoi: This is peculiar... they did pass on my host machine (OSX). Unfortunately I couldn't afford to dive deeper at the moment (sorry). I will add a comment and skip them here.

aaronayres35: No worries!

Issue #1556, opened 2021-03-25
Images used in user manual The Predefined Trait Editor Factories page should be updated


aaronayres35:
The images show UI's that look pretty outdated.

It would be nice to have the images be regenerated using pyqt5

rahulporuri:
looks like you mean the images in this page - https://docs.enthought.com/traitsui/traitsui_user_manual/factories_basic.html

corranwebster:
If you do this, write a script to automate it.  Many of the images come from example/demo scripts.

PR #1574, opened 2021-04-08
Add script to regenerate screenshots of editors for documentation

aaronayres35:
Part of #1556 

This PR adds a script `regenerate_example_screenshots.py` which looks through all the demo examples for files ending with `Editor_demo.py`, launches the demo and takes a screenshot of the UI.  The screenshot is saved to a .png file of the same name as the demo.

Note, some editors in the docs do not have examples currently, and for others, more than one screenshot is needed.  As a result this PR is simply the first step of a larger process towards automating this whole thing.  Separate follow up issues/PRs will need to be opened until we are at a place where we have fully automated the regeneration of all the relevant screenshots for the docs.  

Much of the code for the script was directly pulled from https://github.com/enthought/traitsui/blob/master/integrationtests/test_all_examples.py

~I am also unsure where the script should sit.  I will move it as needed (it shouldn't stay where it is)~ Edit I moved it into a new `scripts` folder sitting inside the user manual docs.


aaronayres35:
Ah I had forgot about this.  I will close this PR now, see comment https://github.com/enthought/pyface/pull/765#issuecomment-717406171

aaronayres35:
I've opened #1579

docs/source/traitsui_user_manual/scripts/regenerate_example_screenshots.py
```diff
@@ -0,0 +1,151 @@
...
+    )
+
+
+def get_python_files(directory):
+    """ Report Python files to be tested or to be skipped.
+
+    Returns
+    -------
+    accepted_files : list of str
+        Python file paths to be tested.
```
rahulporuri, line 41: > Python file paths to be tested.

needs to be updated

docs/source/traitsui_user_manual/scripts/regenerate_example_screenshots.py
```diff
@@ -0,0 +1,151 @@
...
+            accepted_files.append(path)
+
+    # Avoid arbitrary ordering from the OS
+    return sorted(accepted_files)
+
+
+def get_editor_example_filees(python_files):
+    Editor_demo_files = []
+    for filename in python_files:
+        if filename.endswith('Editor_demo.py'):
```
rahulporuri, line 58: did you take a look at the examples to ensure that this convention is followed by all of the relevant examples files?

docs/source/traitsui_user_manual/scripts/regenerate_example_screenshots.py
```diff
@@ -0,0 +1,151 @@
...
+    with UITester().create_ui(instance, ui_kwargs) as ui:
+        ui_pixmap = ui.control.grab()
+        ui_pixmap.save(
+            os.path.join(
+                "docs",
+                "source",
+                "traitsui_user_manual",
+                "images",
+                screenshot_name
+            )
```
rahulporuri, line 97: looks like this is the reason why this script needs to be run from the root of the directory.

like we discussed over chat, we can simply document the fact that this script needs to be run from the root.

rahulporuri (approved):
LGTM with a couple of comments. I guess you were planning on deleting the old image files in a separate PR?

docs/source/traitsui_user_manual/scripts/regenerate_example_screenshots.py
aaronayres35: This is true for all the examples in `examples/demo/StandardEditors`.  There are a couple others which endup having this name eg `examples/demo/Advanced/Dynamic_EnumEditor_demo.py` which may not yet be currently included in the docs page.  Others which do _not_ have this name may potentially be relevant in the docs, but are not currently included.  Most of those that do not have this naming convention are showcasing more advanced/specific features and aren't what we would want to screenshot to show in the docs most likely.

This does tie into the need for special handling certain files though, and again this can be done in follow up PR(s) where this script progressively gets more an more complex or other scripts are added

aaronayres35:
> I guess you were planning on deleting the old image files in a separate PR?

I'm happy to do that here actually.  It is probably easier that way to make sure everything is kept track of / nothing gets forgotten

docs/source/traitsui_user_manual/factories_basic.rst
```diff
@@ -272,18 +270,18 @@ display a Color trait or edit it by selecting a color from the palette available
 in the underlying GUI toolkit. The four styles of color editor are shown in
 Figure 26.
 
-.. figure:: images/color_editors.jpg
```
rahulporuri, line 275: looks like you forgot to remove this file

Issue #1462, opened 2021-01-05
Make Qt WebEngine/WebKit an optional dependency for traitsui.api


kitchoi:
This issue is for making `HTMLEditor` an optional import of `traitsui.api` such that this module can still be used as the convenient entry point when (1) HTMLEditor is not needed (2) Qt is required but Qt WebEngine/WebKit are not available.

Motivation:
- Qt WebKit has been deprecated in favour of WebEngine. Building (and importing) Qt WebEngine is a nontrivial business. HTMLEditor is not as commonly required a feature.

Context:
- `traitsui.api` is provided for downstream projects to import various features of TraitsUI. This issue is only about the API module and keeping it convenient. Whether importing `traitsui.editors.html_editor` should fail when the web dependencies are missing is an orthogonal matter.

Related:
- ArrayEditor requires NumPy and is also an optional feature of `traitsui.api`. Currently a warning is emitted but it does seem noisy (see https://github.com/enthought/traitsui/issues/731)
- PythonEditor and PythonShell are optional features of `pyface.api` as well, see https://github.com/enthought/pyface/pull/796 (and its referenced issues)

kitchoi:
Note that there exists `traitsui.editors.api` which contains a subset of `traitsui.api`. This issue probably applies to `traitsui.editors.api` as well.
That said, offline discussioin suggests that `traitsui.api` is recommended over `traitsui.editors.api`.

PR #1575, opened 2021-04-08
Get pyside2 from edm rather than pip

aaronayres35:
Following the lead of chaco, we get pyside2 from edm rather than pip by explicitly adding the enthought/lgpl repo to do the install

### DIFF
--- a/etstool.py
+++ b/etstool.py
@@ -242,7 +242,7 @@ def install(runtime, toolkit, environment, editable, source):
     # pip install pyqt5 and pyside2, because we don't have them in EDM yet
     if toolkit == 'pyside2':
         commands.append(
-            "edm run -e {environment} -- pip install pyside2"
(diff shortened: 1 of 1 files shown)


rahulporuri:
sigh. that's the reason why we can't use pyside2 from edm. i think we already have an issue regarding the use of pyside2 from edm. if there is, we can just close this PR for now.

aaronayres35:
Currently I see #1462 which covers he source of the issue (although maybe we want a more specific issue about using pyside2 that links to that issue, but that may be unnecessary.  In any case I am closing this now

aaronayres35:
Meant to comment this here earlier, accidentally commented on another PR

Ah I had forgot about the `QtWebKit` problem.  I will close this PR now, see comment https://github.com/enthought/pyface/pull/765#issuecomment-717406171

PR #1581, opened 2021-04-20
Replace "Either" trait with "Union"

rahulporuri:
This PR replaces the `Either` trait with the `Union` trait.

> A new Union trait type has been added. This is intended as a simpler replacement for the existing Either trait type, which will eventually be deprecated.

See https://github.com/enthought/traits/releases/tag/6.1.0


rahulporuri:
I think this might be an intermittent failure. Restarting CI job now.

```
FAIL: test_modal_view_application_cancel (traitsui.tests.test_view_application.TestViewApplication)
  File "/Users/runner/work/traitsui/traitsui/.edm/envs/traitsui-test-3.6-pyside2/lib/python3.6/site-packages/traitsui/tests/test_view_application.py", line 180, in test_modal_view_application_cancel
    self.assertFalse(self.event_loop_timeout)
AssertionError: True is not false
```

traitsui/ui_editors/data_frame_editor.py
```diff
@@ -306,13 +306,13 @@ class DataFrameEditor(BasicEditorFactory):
     show_titles = Bool(True)
 
     #: Optional list of either column ID or pairs of (column title, column ID).
-    columns = List(Either(Str, Tuple(Str, Str)))
+    columns = List(Union(Str, Tuple(Str, Str)))
```
aaronayres35, line 309: I believe these would have previously defaulted to `None`, but now default to `""`.

traitsui/context_value.py
```diff
@@ -111,7 +111,7 @@ def CVType(type, **metadata):
         type or an instance of the ContextValue class.
     """
     metadata.setdefault("sync_value", "to")
-    return Either(type, InstanceOfContextValue, **metadata)
+    return Union(type, InstanceOfContextValue, **metadata)
```
aaronayres35, line 114: Either would have defaulted to None here, but None may not have been one of the allowed values.  From the docs:
"If a default value is not defined, Union uses the default value from the first trait in its definition, whereas Either uses None.

For example:

`Either(Int(), Float())`

has a default value of None. However None is not one of the allowed values. If the trait is later set to None from a non-None value, a validation error will occur.

If the trait definition is changed to:

`Union(Int(), Float())`

Then the default value will be 0, which is the default value of the first trait.

To keep None as the default, use None as the first item:

`Union(None, Int(), Float())`

With this, None also becomes one of the allowed values."

It seems like there isn't a way to _exactly_ preserve this behavior?  We may not want /need to preserve the behavior?

aaronayres35 (approved):
LGTM, but I am still a little unclear on what thee recommended migration pattern is for cases where `Either` was defaulting to `None`, but `None` was not an allowed value...

traitsui/context_value.py
corranwebster: Actually, in this specific case (since my head is in this example), I suspect that the `Union` behaviour is preferable.

But more generally, yes, we need to be careful of changes in behaviour like this.

traitsui/ui_editors/data_frame_editor.py
rahulporuri: in practice, the default behavior looks the same to me

```
>>> from traits.api import *
>>> class A(HasTraits):
...     cols = List(Either(Str, Tuple(Str, Str)))
...     cols2 = List(Union(Str, Tuple(Str, Str)))
...
>>> a = A()
>>> a.cols
[]
>>> a.cols2
[]
```

rahulporuri:
@aaronayres35 i think both of your concerns have been addressed. Am i okay to merge?

traitsui/ui_editors/data_frame_editor.py
aaronayres35: Ah I see, seems the `List` removes any discrepancies.  Thanks!

aaronayres35:
> @aaronayres35 i think both of your concerns have been addressed. Am i okay to merge?

Looks like CI is failing with `ValueError: Union default value should be set via 'default_value', not 'default'.`
Aside from that, yep my concerns have been addressed - LGTM

rahulporuri:
> Looks like CI is failing with ValueError: Union default value should be set via 'default_value', not 'default'.

Sorry. I fixed that right as you were commenting. It's because of new changes that got merged in from the master branch.

PR #1584, opened 2021-04-22
Remove unused images in docs

rahulporuri:
This PR removes a few unused images from the docs. One of the files removed is a "Thumbs.db", which i suspect was mistakenly checked into the git repository. The other files are related to the now removed themed ui functionality in traitsui.


aaronayres35 (approved):
LGTM

PR #1587, opened 2021-04-23
Update "super" usage in "traitsui.qt4"

rahulporuri:
This PR updates `super` usage in the `traitsui.qt4` submodule. The updates were done using automated regex-based search and replace. After the updates were done, we checked each of the changes to ensure that we were calling super with the right arguments before.

Ref https://github.com/enthought/traits/pull/1280


aaronayres35 (approved):
LGTM

PR #1588, opened 2021-04-23
Update "super" usage in "traitsui.wx"

rahulporuri:
This PR updates the `super` usage in classes in `traitsui.wx` submodule. The update was done using automated regex-based search and replace. After the changes were made, each changes was individually checked to ensure that we were calling super with the right arguments before.

Ref https://github.com/enthought/traits/pull/1280. Also see related #1587


aaronayres35 (approved):
LGTM

PR #1599, opened 2021-04-26
Use "PrefixList" from traits >= 6.1

### LINKED ISSUE #883: Traits 6.1.0 deprecation warning
For an example of deprecation warning, see https://travis-ci.org/github/ericpre/hyperspy/jobs/696413265

```
  /home/travis/virtualenv/python3.8.0/lib/python3.8/site-packages/traitsui/view.py:111: DeprecationWarning: 'TraitPrefixList' trait handler has been deprecated. Use PrefixList instead.
    TraitPrefixList(
```
and 
```
  /home/travis/virtualenv/python3.8.0/lib/python3.8/site-packages/traits/traits.py:330: DeprecationWarning: 'TraitMap' trait handler has been deprecated. Use Map instead.
    other.append(TraitMap(map))
```

### LINKED ISSUE #1296: Remove compatible layers for supporting Traits 6.0
As Traits 6.1 is still relatively recent, current TraitsUI master is still tested against Traits 6.0 (#1054) to minimize potential version compatibility issues downstream. In order to achieve this, trivial compatibility layers have been put up, for example this:
https://github.com/enthought/traitsui/blob/be3d68ad61ac80dea943c2c496f55b6f91dd75d6/traitsui/helper.py#L193-L200

But eventually TraitsUI would want to fix issues that require Traits 6.1+ and those compatibility layers would be obsolete. This issue can be used to track code that can be removed when Traits 6.0 support is finally dropped, i.e. this issue should be closed when those compatibility layers are removed.

Related: #1052

rahulporuri:
This PR relies on `traits.api` directly for `PrefixList` instead of relying on a compatibility layer to support traits 6.0. We bump the traits dependency to `>=6.1`. Note that we would have relied on `>=6.1` anyway because of the on triat change -> observe changes.

Note that the compat layer was introduced in https://github.com/enthought/traitsui/pull/1053/ and fixes #1296

Note that the original issue was planned for traitsui 7.1.0. Ref https://github.com/enthought/traitsui/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22Release+7.1%22

### DIFF
--- a/traitsui/__init__.py
+++ b/traitsui/__init__.py
@@ -14,7 +14,7 @@ try:
 except ImportError:
     __version__ = "not-built"
 
-__requires__ = ["traits>=6.0.0", "pyface>=7.1.0"]
+__requires__ = ["traits>=6.1.0", "pyface>=7.1.0"]
 __extras_require__ = {
     "wx": ["wxpython>=4", "numpy"],
     "pyqt": ["pyqt>=4.10", "pygments"],
--- a/traitsui/helper.py
+++ b/traitsui/helper.py
@@ -177,17 +177,3 @@ def compute_column_widths(available_space, requested, min_widths, user_widths):
 # -------------------------------------------------------------------------
 
 SequenceTypes = (tuple, list)
-
-
-# -------------------------------------------------------------------------
-#  Wrapper for TraitPrefixList deprecation:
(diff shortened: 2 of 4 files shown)


aaronayres35 (approved):
LGTM

Issue #1473, opened 2021-01-07
Remove backwards compatibility toolkit factory imports


rahulporuri:
The commit https://github.com/enthought/traitsui/commit/b492c838bc8ef57ed3bd833622014fbfcc625e7c introduced a number of imports across the codebase for backwards compatibility, first available as part of the 4.0.0 release of TraitsUI (at the very least). We should now be able to remove all of the backwards compat imports from the codebase.

PR #1600, opened 2021-04-26
Remove Editor and EditorFactory factory methods on Toolkit objects

### LINKED ISSUE #1597: Remove Editor and EditorFactory factory methods from Qt GUIToolkit
https://github.com/enthought/traitsui/blob/18e35a3525bc0a0c9482c23abfbae5d4626d1064/traitsui/qt4/toolkit.py#L455-L463

https://github.com/enthought/traitsui/blob/18e35a3525bc0a0c9482c23abfbae5d4626d1064/traitsui/qt4/toolkit.py#L465-L688

The "Editor Factory factory methods" on the `GUIToolkit` object seem to be a relic of earlier versions of TraitsUI. These methods aren't used in the package and we should be able to remove them without any issues.

Update : Looking at old versions of TraitsUI, these methods were used in traitsui editors e.g. asking for the `ButtonEditor` editor would call `GUIToolkit().button_editor`.

rahulporuri:
This PR removes the Editor and EditorFactory factory methods on the base Toolkit object - and the toolkit-specific Toolkit objects for the various supported toolkits i.e. null, qt and wx.

These methods were earlier used by the user facing editors e.g. BooleanEditor would call toolkit().boolean_editor. This is no longer necessary as the user now requests an EditorFactory which can directly access the BooleanEditor toolkit-specific implementation.

fixes #1597 

Note that this PR makes it easy to solve #1473 , which will be handled in a separate PR.


traitsui/qt4/toolkit.py
```diff
@@ -451,238 +451,3 @@ def font_trait(self, *args, **traits):
         from . import font_trait as ft
```
aaronayres35, line 451: I imagine all these factory methods for trait types can go as well?  I'm guessing that will be a follow up PR?

aaronayres35 (approved):
LGTM

traitsui/qt4/toolkit.py
aaronayres35: ah maybe not https://github.com/enthought/traitsui/blob/master/traitsui/toolkit_traits.py

rahulporuri: exactly. the toolkit traits are used and exposed to end users. I'll evaluate the other methods on the toolkit object later.

PR #1609, opened 2021-04-26
Expose notify interval in VideoEditor

mdsmarte:
Follow up to PR #745 that adds the ability to control the [notify interval](https://doc.qt.io/qt-5/qmediaobject.html#notifyInterval-prop) of the media player.

When running the `VideoEditor` demo, you should see position updates occur approx every 0.5s now (previously 1s).

### DIFF
--- a/traitsui/qt4/video_editor.py
+++ b/traitsui/qt4/video_editor.py
@@ -181,6 +181,10 @@ class VideoEditor(Editor):
     #: Synchronized to the trait named by factory.image_func.
     image_func = Callable()
 
+    #: The change in position required for an update to be emitted.
+    #: Synchronized to the trait named by factory.notify_interval.
+    notify_interval = Float(1.0)
+
     def update_to_regular(self):
         if self.surface is not None:
             self.surface.frameAvailable.disconnect(self.control.setImage)
@@ -232,6 +236,7 @@ class VideoEditor(Editor):
         self._update_muted()
         self._update_volume()
         self._update_playback_rate()
+        self._update_notify_interval()
 
         self._connect_signals()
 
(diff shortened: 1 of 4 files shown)


traitsui/qt4/video_editor.py
```diff
@@ -314,6 +323,9 @@ def _media_status_changed_emitted(self, error):
     def _buffer_status_changed_emitted(self, error):
         self.buffer = self.media_player.bufferStatus()
 
+    def _notify_interval_changed_emitted(self, interval):
+        self.notify_interval = interval / 1000.0
```
jwiggins, line 327: Does this need to be in a `with self.updating_value():` block like the `position` trait?

jwiggins (approved):
Basically LGTM but with one question.

traitsui/qt4/video_editor.py
mdsmarte: Possibly.... I couldn't understand why this was done for `_position_changed_emitted` but not the other `*_emitted` methods though. Is there a reason?

jwiggins: That was the fix for choppy playback on Linux and Windows. See https://github.com/enthought/traitsui/pull/745#discussion_r607888802 for more discussion

Basically, we want to call `self.media_player.setNotifyInterval` if the `notify_interval` trait changed for any other reason **except** `_notify_interval_changed_emitted` being called.

Unless it doesn't matter.

traitsui/qt4/video_editor.py
```diff
@@ -181,6 +181,10 @@ class VideoEditor(Editor):
     #: Synchronized to the trait named by factory.image_func.
     image_func = Callable()
 
+    #: The change in position required for an update to be emitted.
+    #: Synchronized to the trait named by factory.notify_interval.
```
corranwebster, line 185: Should document the time units.

If the units aren't seconds, then it's probably worth having this value in seconds and then converting to Qt's preferred units (I _think_ you're doing that).

jwiggins: This is documented in the factory

traitsui/qt4/video_editor.py
mdsmarte: Ah that makes sense. It doesn't seem to matter on Mac/Windows. (Added an `observe` for `notify_interval` and Qt never emits a change signal, when both streaming and watching a local video.)

Issue #1586, opened 2021-04-23
Can we update "*Editor = ToolkitEditorFactory" in "traitsui.editors"?


rahulporuri:
https://github.com/enthought/traitsui/blob/18e35a3525bc0a0c9482c23abfbae5d4626d1064/traitsui/editors/boolean_editor.py#L83

Code like what you see above can be found across multiple editor factories available in `triatsui.editors.api`.

@rkern added this comment in an internal discussion

> The original implementations by Dave Morrill, the original author of Traits and TraitsUI, tend to be consistent (ArrayEditor = ToolkitEditorFactory). Those of us who had to implement new EditorFactories later (CSVListEditor is a relatively late addition) tend to ignore that convention as it is, indeed, weird and seemingly pointless.

AFAIK, we have no need at the moment to name the classes `ToolkitEditorFactory` so can we update the class names to directly use the relevant name e.g. `BooleanEditor` in the above example.

corranwebster:
Removing the `ToolkitEditorFactory` names entirely may break pickles.  I would suggest a two-phase approach:

- rename the classes to what we want (eg. `class BooleanEditor`) but also keep the `ToolkitEditorFactory` name around (eg. do a `ToolkitEditorFactory = BooleanEditor`)
- at some later time remove the `ToolkitEditorFactory` name.

PR #1610, opened 2021-04-26
Rename editor factory classes

rahulporuri:
This PR swaps the editor names with the new alias `ToolkitEditorFactory` e.g. `ArrayEditor` is the class name now and `ToolkitEditorFactory` is now the alias. Note that this PR doesn't address all of the uses of `ToolkitEditorFactory` as the class name. There are all of the trivial changes and i'll open a separate PR to address the non-trivial changes.

addresses a part of #1586

### DIFF
--- a/traitsui/editors/boolean_editor.py
+++ b/traitsui/editors/boolean_editor.py
@@ -18,7 +18,7 @@ from traits.api import Dict, Str, Any
 
 from ..view import View
 
-from .text_editor import ToolkitEditorFactory as EditorFactory
+from .text_editor import TextEditor as EditorFactory
 
 # -------------------------------------------------------------------------
 #  Trait definitions:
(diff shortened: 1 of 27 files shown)


traitsui/editors/boolean_editor.py
```diff
@@ -18,7 +18,7 @@
 
 from ..view import View
 
-from .text_editor import ToolkitEditorFactory as EditorFactory
+from .text_editor import TextEditor as EditorFactory
```
aaronayres35, line 21: why import this as `EditorFactory`?

aaronayres35 (approved):
LGTM

traitsui/editors/boolean_editor.py
rahulporuri: no good reason, really. cleaning these up as well

rahulporuri: cleaned up.

PR #1611, opened 2021-04-26
Update changelog

aaronayres35:
This PR starts the process of updating the changelog in advance of the upcoming 7.2.0 release

It also pulls in the news fragment system previously introduced in apptools (see https://github.com/enthought/apptools/pull/181) to make the process of updating the changelog more efficient in future releases.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -3,9 +3,11 @@ Traits UI Changelog
 
 Release 7.2.0
 -------------
-TraitsUI 7.1.0 is a minor release which 
-
-This release is compatible with Traits 6.1+.
+TraitsUI 7.2.0 is a minor release which includes numerous bug fixes,
+documentation imporvements, code maintenance changes, and enhancements.
+Of particular note is the start of the migration process from using
+``on_trait_change`` to using the new Traits ``observe`` framework. As a result,
+TraitsUI now requires Traits >= 6.1.
 
 Highlights of this release
 ~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


rahulporuri:
@aaronayres35 looks like the only thing remaining on this PR are the detailed notes and highlights?

CHANGES.txt
```diff
@@ -1,6 +1,147 @@
...
+~~~~~~~~~~~~~~~~~~
+
+* This release of TraitsUI now depends on Traits 6.1+. Also, deprecated code /
+  modules have been removed. Namely, the :mod:`traitsui.image <traitsui.image>`
+  module which was moved to ``pyface.image``, ``editors_gen`` modules, Editor
+  and EditorFactory factory methods on Toolkit objects, and more.  For a
+  complete list, see PRs in the "Removals" section below.  These were all
+  generally unused / deprecated for sometime. Also, importing directly from
+  ``traitsui.editors`` has been deprecated. Please update imports to import
+  directly from :mod:`traitsui.api` or :mod:`traitsui.editors.api`.
```
aaronayres35, line 29: Which should we be recommending here? I listed both currently, but I imagine we just want people to import from `traitsui.api`?  
This has been discussed before

rahulporuri: >  This has been discussed before

I dont remember that discussion - or the conclusion of that discussion.
I'm fine with listing both. I completely forgot about the fact that the editor factories are also importable via `traitsui.api`. I think we can leave it up to the users to choose. I don't think we'll remove `traitsui.editors.api`.

rahulporuri (approved):
LGTM

PR #1614, opened 2021-04-27
Remove code handling old/outdated versions of wx

rahulporuri:
At the moment, we only support the latest versions of WxPython (4.x) so this PR removes the code which was explicitly handling these old/outdated versions.


aaronayres35 (approved):
LGTM

PR #1621, opened 2021-04-27
Don't require synced traits with VideoEditor

jwiggins:
@corranwebster This is an implementation of what I believe you described as the intended use of `Editor` synced traits, applied to `VideoEditor`.

PR #1628, opened 2021-04-28
Change video editor to use ContextValue infrastructure

corranwebster:
This is an alternative approach for #1621.

### DIFF
--- a/traitsui/editors/video_editor.py
+++ b/traitsui/editors/video_editor.py
@@ -80,7 +80,8 @@ class VideoEditor(BasicEditorFactory):
 
     #: A ContextValue reference to a trait that will synchronize to the
     #: media status of the player.  The value is a string.
-    video_error = Instance(ContextValue, args=(), allow_none=False, sync_value='to')
+    video_error = Instance(ContextValue, args=(), allow_none=False,
+                           sync_value='to')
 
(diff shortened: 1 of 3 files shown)


traitsui/editors/video_editor.py
```diff
@@ -39,50 +40,59 @@ class VideoEditor(BasicEditorFactory):
...
+    #: The initial playback position of the video player or a ContextValue
+    #: reference to a trait that will synchronize to the playback position.
+    #: The value is a floating point time in seconds.
+    position = CVType(Float, sync_value='both')
+
+    #: A ContextValue reference to a trait that will synchronize to the
+    #: duration of the video.  The value is a floating point time in
+    #: seconds.
+    duration = Instance(ContextValue, args=(), allow_none=False,
+                        sync_value='to')
```
jwiggins, line 69: `args` can't be empty, otherwise you get an exception when nothing is passed to the editor factory.

corranwebster: Arg.  Should probably be `('',)` there and elsewhere.  Or maybe want to fix the definition of `ContextValue`.

corranwebster:
Closing as #1621 has everything that this does now.

PR #1632, opened 2021-04-29
use UI Tester for all tests in test_html_editor.py

aaronayres35:
#1551 

Has been coming up more frequently it seems lately.  This PR will not change that, but seeing the failure made me notice we were not using UI Tester on the failing test, and instead were using the old `create_ui` function which now just goes through UITester itself anyway.  This PR updates two tests to use `UITester.create_ui` instead


rahulporuri (approved):
LGTM

PR #1634, opened 2021-04-29
[WIP] Add default factory callable

### LINKED ISSUE #1627: Allow specifying callable for generating new item in List
See comment:
https://github.com/enthought/traitsui/pull/1559#discussion_r621020135 and https://github.com/enthought/traitsui/issues/1154#issuecomment-827456146 

The original issue was #1154 which was _parttially_ resolved by #1559.  That PR prevented a hard crash, but we still need to provide a way for the problem to be avoided altogether.  Namely, we need to provide a means for sspecifying how to create a new item to add to the list.

aaronayres35:
follow up PR to #1559 
closes #1627

This PR adds a `default_factory` `Callable` trait on the `ListEditor` factory class.  Then it uses this callable to create a new item in the `add_item` methods of the toolkit specific editors if it was specified.  Otherwise, it defaults back to the old behavior of using the default value for the trait type in the list.

### DIFF
--- a/traitsui/editors/list_editor.py
+++ b/traitsui/editors/list_editor.py
@@ -99,6 +99,10 @@ class ToolkitEditorFactory(EditorFactory):
     #: Show a right-click context menu for the notebook tabs?  (Qt only)
     show_notebook_menu = Bool(False)
 
+    #: Factory callable that will be called to create the new element to add to
+    #: this list.  If None, the default value for the trait of interest is used.
+    default_factory = Callable()
+
     # -- Notebook Specific Traits ---------------------------------------------
 
     #: Are notebook items deletable?
--- a/traitsui/wx/list_editor.py
+++ b/traitsui/wx/list_editor.py
@@ -16,7 +16,8 @@ import wx
 
 import wx.lib.scrolledpanel as wxsp
 
-from traits.api import Str, Any, Instance, Property, Bool, cached_property
+from traits.api import (
+    Any, Bool, cached_property, Instance, Property, Str, TraitError
 from traits.trait_base import user_name_for, xgetattr
 
 from traitsui.ui_traits import Image, convert_bitmap
@@ -343,8 +344,17 @@ class SimpleEditor(Editor):
         list, index = self.get_info()
         index += offset
         item_trait = self._trait_handler.item_trait
-        value = item_trait.default_value_for(self.object, self.name)
-        self.value = list[:index] + [value] + list[index:]
+        if self.factory.default_factory:
+            value = self.factory.default_factory()
+        else:
+            value = item_trait.default_value_for(self.object, self.name)
+        try:
+            self.value = list[:index] + [value] + list[index:]
+        # if the default new item is invalid, we just don't add it to the list.
+        # traits will still give an error message, but we don't want to crash
+        except TraitError:
+            from traitsui.api import raise_to_debug
+            raise_to_debug()
         wx.CallAfter(self.update_editor)
 
     def add_before(self):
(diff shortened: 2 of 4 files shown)


corranwebster (approved):
LGTM

corranwebster:
Actually, although this is good, might want to pass additional arguments to the default factory in case what you create depends on the context.

traitsui/editors/list_editor.py
```diff
@@ -84,6 +86,18 @@ class ListEditor(EditorFactory):
     #: Show a right-click context menu for the notebook tabs?  (Qt only)
     show_notebook_menu = Bool(False)
 
+    #: Factory callable that will be called to create the new element to add to
+    #: this list. If None, the default value for the trait of interest is used.
+    default_factory = Callable()
```
rahulporuri, line 91: i'd call this an `item_factory` because that's what it is - a factory function to create items - in the context of the list trait. `list_item_factory` felt unnecessarily verbose.

traitsui/editors/list_editor.py
```diff
@@ -84,6 +86,18 @@ class ListEditor(EditorFactory):
...
+
+    #: List of positional arguments to be passed to the default factory
+    #: callable when creating new elements
+    default_args = List()
+
+    #: Dictionary of keyword arguments to be passed to the default factory
+    #: callable when creating new elements
+    default_kwargs = Dict()
```
rahulporuri, line 99: i'd rename `default_args` and `default_kwargs` to `item_factory_args` and `item_factory_kwargs` to reflect where they get used.

I think this will make the calling code readable - 

```
value = self.factory.item_factory(*self.factory.item_factory_args, **self.factory.item_factory_kwargs)
```

traitsui/editors/list_editor.py
```diff
@@ -84,6 +86,18 @@ class ListEditor(EditorFactory):
     #: Show a right-click context menu for the notebook tabs?  (Qt only)
     show_notebook_menu = Bool(False)
 
+    #: Factory callable that will be called to create the new element to add to
+    #: this list. If None, the default value for the trait of interest is used.
+    default_factory = Callable()
+
+    #: List of positional arguments to be passed to the default factory
+    #: callable when creating new elements
+    default_args = List()
```
rahulporuri, line 95: nit- the args to a callable need to be a `Tuple`, not a `List`

traitsui/editors/list_editor.py
```diff
@@ -84,6 +86,18 @@ class ListEditor(EditorFactory):
     #: Show a right-click context menu for the notebook tabs?  (Qt only)
     show_notebook_menu = Bool(False)
 
+    #: Factory callable that will be called to create the new element to add to
+    #: this list. If None, the default value for the trait of interest is used.
```
rahulporuri, line 90:
```suggestion
    #: Factory that will be called to create and add a new element to
    #: this list. If None, the default value for the trait of interest is used.
```

traitsui/wx/list_editor.py
```diff
@@ -343,8 +345,19 @@ def add_item(self, offset):
...
+            )
+        else:
+            value = item_trait.default_value_for(self.object, self.name)
+        try:
+            self.value = list[:index] + [value] + list[index:]
+        # if the default new item is invalid, we just don't add it to the list.
+        # traits will still give an error message, but we don't want to crash
+        except TraitError:
+            from traitsui.api import raise_to_debug
+            raise_to_debug()
```
rahulporuri, line 360: question : looks like we missed this in the earlier PR

rahulporuri (approved):
Mostly LGTM with comments regarding the new trait names

traitsui/wx/list_editor.py
aaronayres35: yeah wx got neglected before 😞

PR #1637, opened 2021-05-02
Remove CI test against Traits 6.0

kitchoi:
Since the release of Pyface 7.3 in late February, the CI job supposed to be run against Traits 6.0 was actually run against Traits 6.2 (Traits gets upgraded upon installing Pyface).
#1599 also explicitly drops support for Traits 6.0

This PR removes the CI workflow that is no longer relevant.


rahulporuri (approved):
LGTM. Thanks @kitchoi !

PR #1639, opened 2021-05-03
skip video editor test without numpy / move numpy import to when it is needed / list numpy as test dependency

### LINKED ISSUE #1636: Test test_video_editor_basics should be skipped when NumPy is absent
NumPy is an optional dependency for Qt backend, tests requiring NumPy are skipped if NumPy is absent, e.g. Tests for ArrayEditor and DataFrameEditor are skipped if NumPy is absent.

However upon running the following:
```
$ pip install -e .[pyside2,test]
$ python -m unittest discover -t . -v traitsui
```
One gets the following test error:
```
ERROR: test_video_editor_basics (traitsui.tests.editors.test_video_editor.TestVideoEditor)
...
    return original_import(name, *args, **kwargs)
ModuleNotFoundError: No module named 'numpy'
```

aaronayres35:
closes #1636

This PR does a number of things.  It moves the `import numpy as np` statement from the top of the module to inside the `__init__` method so it is only imported when needed.  It skips the video editor tests when numpy is not installed.  (both of these follow exactly was it done for the `DataFrameEditor`.  

Lastly, I added `numpy` as a test dependency in `extras_require`.  My understanding was that if I do `pip install -e .[test]` I should have all the deps needed to run _every_ test in the test suite

### DIFF
--- a/traitsui/__init__.py
+++ b/traitsui/__init__.py
@@ -43,6 +43,7 @@ __extras_require__ = {
     "test": [
         # Dependencies for running test suites.
         "packaging",
+        "numpy"
     ],
 }
 
--- a/traitsui/qt4/video_editor.py
+++ b/traitsui/qt4/video_editor.py
@@ -9,9 +9,6 @@
 # Thanks for using Enthought open source!
 
 """Traits UI 'display only' video editor."""
-
-import numpy as np
-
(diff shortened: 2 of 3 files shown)


rahulporuri (approved):
This works fine.

PR #1652, opened 2021-05-07
explicitly install swig 3.0.12 for cron job

aaronayres35:
This PR is in response to the failing cron job https://github.com/enthought/traitsui/runs/2514003659?check_suite_focus=true 

I believe failure was a result of https://github.com/enthought/enable/issues/360

This PR explicitly installs swig 3.0.12 before trying to run tests

**Checklist**

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -32,6 +32,7 @@ jobs:
         # Dependencies include enable, which has known issue with swig 4.0
         # seeenthought/enable#360
         run: sudo apt-get install swig=3.0.12
+        if: startsWith(matrix.os, 'ubuntu')
       - name: Install Python packages and dependencies
         run: |
           python -m pip install --upgrade pip setuptools wheel
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -4,9 +4,9 @@
 
 name: Integration tests
 
-on:
-  schedule:
-    - cron:  '0 0 * * 4'
+on: pull_request
+  #schedule:
+  #  - cron:  '0 0 * * 4'
 
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -4,9 +4,9 @@
 
 name: Integration tests
 
-on:
-  schedule:
-    - cron:  '0 0 * * 4'
+on: pull_request
+  #schedule:
+  #  - cron:  '0 0 * * 4'
```
aaronayres35, line 9: Will revert this before merging.  Needed to verify the changes made

aaronayres35:
```
Run sudo apt-get install swig
  sudo apt-get install swig
  shell: /usr/bin/bash -e {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.6.13/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.6.13/x64/lib
Reading package lists...
Building dependency tree...
Reading state information...
swig is already the newest version (4.0.1-5build1).
0 upgraded, 0 newly installed, 0 to remove and 25 not upgraded.
```
Still using swig 4, i still need to look into this more

.github/workflows/integration-tests.yml
```diff
@@ -4,20 +4,22 @@
...
 
 jobs:
   test:
     strategy:
       matrix:
-        os: [ubuntu-latest]
+        # Note that Ubuntu 20 cannot be used because its SWIG version is 4.0
+        # (enthought/enable#360)
```
rahulporuri, line 16: i'd be surprised if we can't explicitly install swig 3 on ubuntu 20

rahulporuri (approved):
Mostly LGTM with one question about swig version vs ubuntu version

.github/workflows/integration-tests.yml
aaronayres35: I've opened #1660

PR #1666, opened 2021-05-11
Explicitly require traits 6.2 and pyface 7.2

aaronayres35:
This was accidentally missed before the release.  With recent changes `TraitsUI` now rquires newer `Traits` and `Pyface`.

These are a result of using `Property(observe=...)` and `from pyface.undo.api import ...`
**Checklist**
~- [ ] Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~

### DIFF
--- a/traitsui/__init__.py
+++ b/traitsui/__init__.py
@@ -14,7 +14,7 @@ try:
 except ImportError:
     __version__ = "not-built"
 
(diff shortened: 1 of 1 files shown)


rahulporuri (approved):
LGTM

PR #1668, opened 2021-05-12
require pyface 7.3

aaronayres35:
With pyface 7.2 there was a consistent test failure.  It appears we actually require pyface 7.3

**Checklist**
~- [ ] Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~


rahulporuri (approved):
LGTM

Issue #1673, opened 2021-05-18
GroupEditor does not call super()


corranwebster:
The helper class `GroupEditor` does not call `super()` in its `__init__()` and so Traits will not properly initialize:
https://github.com/enthought/traitsui/blob/7681a2b8b851394411213e6f1eeda18b1a02062d/traitsui/qt4/ui_panel.py#L1255-L1262
(similarly for wx).

This could be part of the problem for issues involving `enabled_when`, etc not applying to labels or other components (eg. #758) or updating properly (or not, but it definitely and anti-pattern for Traits to do this).

aaronayres35:
This is a bit awkward because `super` would end up calling:
https://github.com/enthought/traitsui/blob/0c30b77bc5cf6ba8aa4512f4e32834da269c83c4/traitsui/editor.py#L523-L538

In this case, users of `GroupEditor` and its subclasses do not currently specify a `parent`.  Further, looking at the body of the above method, the `GroupEditor` does not have a `name` / there is no specific trait the editor is editing, which will lead to a break in this call as `self.name` will be `<undefined>`.  Perhaps we could directly call `__init__` on the base class higher up the chain (in this case `HasPrivateTraits`).  
ie. at the start of the `GroupEditor.__init__` method call `HasPrivateTraits.__init__(self, **traits)` or `super(Editor, self).__init__(**traits)` but I feel like that smells a bit.
Also, as a side note, I am unsure why parent is a required positional arg in the above code as it is unused in the body of the method?

corranwebster:
Right - I think that what is needed is a superclass of `Editor` which both `GroupEditor` and `Editor` inherit from.  I've been looking into it and I'm not quite sure which bits of `Editor` that `GroupEditor` is using, but I think it is at least the enable/disable and visibility stuff, but also possibly the control creation/deletion code.  It doesn't need the code for synchronization of traits with the UI context.

As a more immediate work-around, it might make sense to have `GroupEditor.__init__()` do a direct `HasTraits.__init__()` call, which will at least make sure that the Traits machinery is all properly hooked up.

PR #1674, opened 2021-05-19
Call HasPrivateTraits.__init__() in GroupEditor.__init__

aaronayres35:
Part of #1673 

This PR takes the simpler quick fix approach as mentioned in this comment: https://github.com/enthought/traitsui/issues/1673#issuecomment-843282178

It is certainly worth the investigation into the alternate approach described in that comment.  This was just a very simple fix that could be done in the short term

**Checklist**

### DIFF
--- a/traitsui/qt4/ui_panel.py
+++ b/traitsui/qt4/ui_panel.py
@@ -28,7 +28,7 @@ import re
 
 from pyface.qt import QtCore, QtGui
 
-from traits.api import Any, Instance, Undefined, HasPrivateTraits
+from traits.api import Any, Instance, HasPrivateTraits, Undefined
 from traits.observation.api import match
 
 from traitsui.api import Group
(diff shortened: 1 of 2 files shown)


traitsui/qt4/ui_panel.py
```diff
@@ -1259,6 +1259,7 @@ class GroupEditor(Editor):
     def __init__(self, **traits):
         """ Initialise the object.
         """
+        HasPrivateTraits.__init__(self, **traits)
```
rahulporuri, line 1262: For future reference, it'd be useful if you could add a comment here regd why `HasPrivateTraits.__init__` is being called instead of `Editor.__init__`. Linking to the issue is also a good idea.

rahulporuri (approved):
LGTM

traitsui/qt4/ui_panel.py
rahulporuri: @aaronayres35 can you address this comment? Once you do, we can merge this PR

aaronayres35: Ah, yep sorry about the delay, this PR was lost in the shuffle.  I've added a comment now.

rahulporuri:
still LGTM

PR #1684, opened 2021-06-11
Add RangeEditor support for format_func

### LINKED ISSUE #1650: RangeEditor does not support format_func
All editors have two traits to support converting values into strings for display:

- `format`, which is a `%`-style format string
- `format_func`, which is an arbitrary callable

However the `RangeEditor` completely ignores the existence of `format_func` and only ever uses `format`.  This is a problem if you wanted to display values as times or something similar (the particular use-case here is that we want to display frame numbers as `{minute}:{second}.{fractions of a second}` but we can't do that from an integer value with just `%` strings).

An alternative would be to provide a "slider-only" range editor which allows end-users to supply their own text field with whatever formatters they want.

aaronayres35:
closes #1650 

This PR adds support for `format_func` for the `SimpleSliderEditor` `RangeEditor`.  It does so by calling the `string_value` method defined on the factory, rather than just `self.format % _____`.  See method definition here:
https://github.com/enthought/traitsui/blob/e2521b4cf3ca882591c68ff57e795867964725ec/traitsui/editor_factory.py#L190-L208

I am just realizing while opening this that although the base `EditorFactory` class defines a `format_str` trait, the `RangeEditor` defines its own `format` trait to hold the format string instead.  I suspect this will cause problems, hence the WIP for now.  I would like to rename `format` to `format_str` but that is part of the public api... 
Perhaps I will need to redefine the `string_value` method on `RangeEditor` and have it be exactly the same as the base class just replacing `format_str` with `format`... 🤔 

EDIT: I went with the approach mentioned of "redefine the `string_value` method on `RangeEditor` ...".  It smells a bit, but seems necessary (?) I am not sure why RangeEditor chose to use its own trait there previously rather than using `format_str`. 
**Checklist**

### DIFF
--- a/traitsui/editors/range_editor.py
+++ b/traitsui/editors/range_editor.py
@@ -69,6 +69,7 @@ class RangeEditor(EditorFactory):
     format_str = Str("%s")
 
     #: Deprecated: Please use ``format_str`` instead.
+    #: See enthought/traitsui#1704
     #: Formatting string used to format value and labels.
     format = Property(Str, observe='format_str')
 
--- a/traitsui/editors/range_editor.py
+++ b/traitsui/editors/range_editor.py
@@ -10,6 +10,7 @@
 
 """ Defines the range editor factory for all traits user interface toolkits.
 """
+import warnings
 
 from types import CodeType
 
@@ -64,8 +65,11 @@ class RangeEditor(EditorFactory):
     #: The name of an [object.]trait that defines the high value for the range
     high_name = Str()
 
-    #: Formatting string used to format value and labels
-    format = Str("%s")
+    # set format_str default
+    format_str = Str("%s")
+
+    #: Deprecated: Formatting string used to format value and labels
+    format = Property(Str, observe='format_str')
 
     #: Is the range for floating pointer numbers (vs. integers)?
     is_float = Bool(Undefined)
@@ -194,6 +198,23 @@ class RangeEditor(EditorFactory):
     # -------------------------------------------------------------------------
     #  Property getters.
     # -------------------------------------------------------------------------
+
+    def _get_format(self):
+        warnings.warn(
+            "Use of format trait is deprecated. Use format_str instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self.format_str
+
+    def _set_format(self, format_string):
+        warnings.warn(
+            "Use of format trait is deprecated. Use format_str instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        self.format_str = format_string
+
     def _get_simple_editor_class(self):
         """ Returns the editor class to use for a simple style.
 
@@ -283,30 +304,6 @@ class RangeEditor(EditorFactory):
             ui
         )
         return super().custom_editor(ui, object, name, description, parent)
-
-    def string_value(self, value, format_func=None):
-        """ Returns the text representation of a specified object trait value.
-
-        If the **format_func** attribute is set on the editor factory, then
-        this method calls that function to do the formatting.  If the
-        **format** attribute is set on the editor factory, then this
-        method uses that string for formatting. If neither attribute is
-        set, then this method just calls the appropriate text type to format.
-
-        This is slightly modified for the base EditorFactory inplementation to
-        use this class' ``format`` trait, as opposed to the ``format_str``
-        trait defined on the base class.
-        """
-        if self.format_func is not None:
-            return self.format_func(value)
-
-        if self.format != "":
-            return self.format % value
-
-        if format_func is not None:
-            return format_func(value)
-
-        return str(value)
 
 
 # This alias is deprecated and will be removed in TraitsUI 8.
--- a/traitsui/editors/range_editor.py
+++ b/traitsui/editors/range_editor.py
@@ -301,7 +301,7 @@ class RangeEditor(EditorFactory):
             return self.format_func(value)
 
         if self.format != "":
-            return self.format_str % value
+            return self.format % value
 
         if format_func is not None:
             return format_func(value)
--- a/traitsui/editors/range_editor.py
+++ b/traitsui/editors/range_editor.py
@@ -284,6 +284,30 @@ class RangeEditor(EditorFactory):
         )
         return super().custom_editor(ui, object, name, description, parent)
 
+    def string_value(self, value, format_func=None):
+        """ Returns the text representation of a specified object trait value.
+
+        If the **format_func** attribute is set on the editor factory, then
+        this method calls that function to do the formatting.  If the
+        **format** attribute is set on the editor factory, then this
+        method uses that string for formatting. If neither attribute is
+        set, then this method just calls the appropriate text type to format.
+
+        This is slightly modified for the base EditorFactory inplementation to
+        use this class' ``format`` trait, as opposed to the ``format_str``
+        trait defined on the base class.
+        """
+        if self.format_func is not None:
+            return self.format_func(value)
+
+        if self.format != "":
+            return self.format_str % value
+
+        if format_func is not None:
+            return format_func(value)
+
+        return str(value)
+
 
 # This alias is deprecated and will be removed in TraitsUI 8.
 ToolkitEditorFactory = RangeEditor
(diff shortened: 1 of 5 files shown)


rahulporuri (commented):
This approach works but I would like to see the alternative approach before choosing one over the other.

IMO, in the long term, it would be better if we deprecated and remove `RangeEditor.format` and just use `EditorFactory.format_str`. The right way to do that at the moment would be to make `RangeEditor.format` a property that sets `RangeEditor.format_str` while raising a deprecation warning.

traitsui/editors/range_editor.py
```diff
@@ -284,6 +284,30 @@ def custom_editor(self, ui, object, name, description, parent):
...
         return super().custom_editor(ui, object, name, description, parent)
 
+    def string_value(self, value, format_func=None):
+        """ Returns the text representation of a specified object trait value.
+
+        If the **format_func** attribute is set on the editor factory, then
+        this method calls that function to do the formatting.  If the
+        **format** attribute is set on the editor factory, then this
+        method uses that string for formatting. If neither attribute is
+        set, then this method just calls the appropriate text type to format.
```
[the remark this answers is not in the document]
aaronayres35: I agree, this is quite strange and unexpected behavior...
I will open an issue pointing to the origin of this method (i.e. `EditorFactory.string_value`)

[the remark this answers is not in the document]
aaronayres35: I've opened #1695

traitsui/qt4/range_editor.py
```diff
@@ -80,7 +80,7 @@ class SimpleSliderEditor(BaseRangeEditor):
     #: High value for the slider range
     high = Any()
 
-    #: Formatting string used to format value and labels
+    #: Deprecated: This trait is no longer used.
```
aaronayres35, line 83: I was thinking this should become a property that mirrors the editors `factory_str` but I don't think that is needed / allowed? (ie an instance of an editor can't set its factory's trait). We ultimately had previously just set `self.format = factory.format` in the `init` method anyways.

I am worried there may be a change in behavior if a user was to try to set `format` on an instance of a `RangeEditor`.  e.g `my_editor = RangeEditor()`, `my_editor.format = "something"`.   Previously, because we had  been using `self.format` throughout instead of `string_value`, so this would actually be used.  Now, AFAICT, the best we can do is pass it as a kwarg to `string_value` if a user tries to set it on an instance, but that will no longer take priority if the factory has its trait set...

I don't think we want to change any behavior here (even if the current behavior is strange).  I am unsure how to best move forward.

@rahulporuri any thoughts on this?  I hope my explanation made some sense, if not we can discuss tomorrow at standup

aaronayres35: So initially I expected this Test to pass on master, but fail on this branch:
```
def test_format_behavior_change(self):
        # format trait has been deprecated in favor of format_str. However,
        # behavior should be unchanged.
        model = RangeModel()
        with self.assertWarns(DeprecationWarning):
            my_editor = RangeEditor(format="%s ...")
            my_editor.format = "%s"
            view = View(
                Item(
                    "float_value",
                    editor=my_editor
                )
            )
        tester = UITester()
        with tester.create_ui(model, dict(view=view)) as ui:
            float_value_field = tester.find_by_name(ui, "float_value")
            float_value_text = float_value_field.locate(Textbox())
            self.assertEqual(
                float_value_text.inspect(DisplayedText()), "0.1"
            )
 ```
 
 but it passes on both.  I then realized `my_editor` is always just an instance of the `traitsui.editors.RangeEditor` `EditorFactory`.  The actual `Editor` instances are only created when e.g. the `_get_simple_editor_class` method is called.  If I access the `Editor` directly, e.g. doing something like this:

 ```
def test_format_behavior_change(self):
        # format trait has been deprecated in favor of format_str. However,
        # behavior should be unchanged.
        model = RangeModel()
        view = View(
            Item(
                "float_value",
                editor=RangeEditor(format="%s ...")
            )
        )
        tester = UITester()
        with tester.create_ui(model, dict(view=view)) as ui:
            float_value_field = tester.find_by_name(ui, "float_value")
            float_value_field._target.format = "%s"
            float_value_text = float_value_field.locate(Textbox())
            self.assertEqual(
                float_value_text.inspect(DisplayedText()), "0.1"
            )
```
I expected it to pass on master and fail on this branch.  However, this fails on both... 
So, AFAICT, behavior is not changed by this PR.

traitsui/wx/range_editor.py
```diff
@@ -74,7 +74,7 @@ class SimpleSliderEditor(BaseRangeEditor):
     #: High value for the slider range
     high = Any()
 
-    #: Formatting string used to format value and labels
+    #: Deprecated: This trait is no longer used.
```
aaronayres35, line 77: Rather than simply commenting on the deprecation, I would like to make this a property which raises a deprecation warning. 
Ideally, getting the property would return the `self.factory.format_str` trait, and like wise setting it would set that trait.  However it feels incorrect for an editor instance to be trying to change a trait on its factory...

Also as I think more about this PR, I am worried about Hyrum's law here.  Even if the behavior change seems "more" correct, it is _possible_ that users are dependent on the current behavior.

traitsui/qt4/range_editor.py
aaronayres35: EDIT: see the recently added test

traitsui/editors/range_editor.py
```diff
@@ -64,8 +65,11 @@ class RangeEditor(EditorFactory):
     #: The name of an [object.]trait that defines the high value for the range
     high_name = Str()
 
-    #: Formatting string used to format value and labels
-    format = Str("%s")
+    # set format_str default
```
rahulporuri, line 68:
```suggestion
    #: Formatting string used to format value and labels
```

traitsui/editors/range_editor.py
```diff
@@ -64,8 +65,11 @@ class RangeEditor(EditorFactory):
...
+    format_str = Str("%s")
+
+    #: Deprecated: Formatting string used to format value and labels
```
rahulporuri, line 71:
```suggestion
    #: Deprecated: Please use ``format_str`` instead.
    #: Formatting string used to format value and labels.
```

traitsui/qt4/range_editor.py
rahulporuri: Can you search for users of the `format` trait on this editor? I'm not even sure how one would go about specifically searching for such uses. Whether or not users exist would help us decide on whether or not we need to convert `format` into a property as well - and make getting/setting `format` raise warnings as well.

rahulporuri (approved):
LGTM with a couple of comments.

Can you update the PR description as well? I think it is outdated as we modified the approach after the first review.

traitsui/qt4/range_editor.py
aaronayres35: I was unable to find any users trying to set format on the Editor instance directly.
I did see users setting it on the Factory object, ie `RangeEditor(format= ...)`, but that is covered by the use of property/deprecation in `traitsui/editors/range_editor.py`.

TBH I am not certain what the pattern would be to set it on the Editor instance.  I guess accessing it through `UIInfo`?

aaronayres35: From traitsUI docs: "Note that not every trait on every editor can react to changes: some values are only used at editor creation time; however all editors support dynamically changing the enabled, visible and invalid traits. This feature can sometimes allow developers to avoid having to create a custom Handler subclass."

aaronayres35:
I've updated the news fragment, although it is still a "bugfix".  It really sort of belongs in "deprecation" as well... Should I include it in both (ie add another news fragment with the same message just tagged as deprecation perhaps?

aaronayres35:
CI is failing on ubuntu-latest with pyside2:
```
Command '['edm', 'run', '-e', 'traitsui-test-3.6-pyside2', '--', 'python', '-W', 'default', '-m', 'unittest', 'discover', '-v', '/home/runner/work/traitsui/traitsui/integrationtests']' returned non-zero exit status 252.
No xvfb processes to kill
```
This is also occurring on #1705

rahulporuri:
still LGTM

rahulporuri (approved):
still LGTM

rahulporuri:
@aaronayres35 can you update the PR description because it feels outdated

PR #1686, opened 2021-06-14
Dont test against PyQt4 on CI

aaronayres35:
part of https://github.com/enthought/ets/issues/57

This PR removes testing with pyqt4 from CI

**Checklist**
- [x] Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)

PR #1688, opened 2021-06-17
Remove support for PyQt in etstool helpers

rahulporuri:
This PR removes support for the `PyQt` toolkit in the `etstool.py` helper modules - because we dont test against `PyQt` on CI any longer (ref #1686 ). Note that the `README` has also been updated to not mention `PyQt`.

**Checklist**
- [ ] ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~ This change in itself isn't newsworthy.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -75,9 +75,8 @@ If you want to run traitsui, you must also install:
 
 You will also need one of the following backends:
 
-- PyQt
 - wxPython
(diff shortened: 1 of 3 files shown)


aaronayres35 (approved):
LGTM

PR #1691, opened 2021-06-18
Move "array_editor" to "StandardEditors" contributed examples

### LINKED ISSUE #1646: Move "array_editor" to "StandardEditors" contributed examples
The PR https://github.com/enthought/traitsui/pull/1640 cleans up the `"array_editor"` documentation example. It makes sense to contribute this to the etsdemo application as there is no example of how an array editor can be used at the moment. If we choose to contribute the `"array_editor"` example, we will also want to generate a new screenshot and update the documentation with the same.

aaronayres35:
closes #1646 

This PR takes the example `array_eitor.py` which used to sit in the documentation and moves it into the `traitsui/examples/demo/Standard_Editors` module.  It also generates a new screenshot for the example and then uses these in the documentation.  It removes the old screenshot and example file.

**Checklist**


rahulporuri (commented):
It looks like there are a couple of bugs here.

- the separators between the different sections of the UI seem off. They look off in the image and they look off locally when i run the example.
- the integer sections of the editor used to show `0` earlier but now they show `0.0`.

aaronayres35:
Once #1702 is merged I will regenerate the screenshots and this PR should be good to go

rahulporuri:
@aaronayres35 i'm still not conviced about #1702 but I don't want this PR stuck in limbo.

Can you do some git magic locally to generate a screenshot without the separator issue - and push that screenshot on this branch? That way, we can push this PR ahead without being blocked on #1702 .

aaronayres35:
> @aaronayres35 i'm still not conviced about #1702 but I don't want this PR stuck in limbo.
> 
> Can you do some git magic locally to generate a screenshot without the separator issue - and push that screenshot on this branch? That way, we can push this PR ahead without being blocked on #1702 .

That makes sense to me.  TBH I am a little unsure on 1702 still myself.  I don't really understand how previously things all looked correct for me locally, and now there was an off by one error (I think change occurred after updating macOS, but that doesn't mean change is correct as it might break code for older macOS?)
I've regenerated the correct looking screenshot on this PR and added a news fragment.  This look okay @rahulporuri

rahulporuri (approved):
LGTM 🎉 Thanks for the changes!

PR #1694, opened 2021-06-21
Fix ArrayEditor data type check

aaronayres35:
This PR was motivated by this comment:
https://github.com/enthought/traitsui/pull/1691#pullrequestreview-688037697
specifically the second bullet.

Note that adding `print(object.dtype.type)` while running `ArrayEditor_demo.py` previously was giving `<class 'numpy.int64'>`, whereas `kind` gives `"i"` as we expect.

**Checklist**


rahulporuri (approved):
LGTM

PR #1698, opened 2021-06-22
Update note in docs as we now have pyqt5 screenshots

aaronayres35:
This PR simply updates a note in the documentation as we now have screenshots using pyqt5.  Some still have the old wx screenshots though which will progressively be updated over time. 

**Checklist**
~- [ ] Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~ Too minor to warrant a news fragment


rahulporuri (approved):
LGTM

Issue #1312, opened 2020-10-06
Horizontal ruled separator does not appear under Groups


chigozienri:
`Item('_')` does not produce a visible separator when immediately underneath a `Group`
To replicate:
```
from traits.api import *
from traitsui.api import *

class SeparatorTest(HasTraits):
    view = View(
        Label('Above'),
        Item('_'),
        Label('Below'),
        Item('_'),
        Group(Label('Caverns')),
        Item('_'),
        Group(Label('Dungeons')),
        Item('_'),
        Group(Label("Extra dungeons")),
        Item('_'),
        Label('Foundations')
    )

if __name__ == "__main__":
    separator = SeparatorTest()
    separator.configure_traits()
```
<img width="206" alt="separator" src="https://user-images.githubusercontent.com/14337872/95202840-c63b3a00-07d9-11eb-8abc-f5cb3bbba1fe.png">

rahulporuri:
@nicolasap-dm points to these two locations as the source of errors - https://github.com/enthought/traitsui/blob/94da89c3199d033f231990dd3c783253a6260dd6/traitsui/qt4/ui_panel.py#L631-L635 https://github.com/enthought/traitsui/blob/ca847133947e5ad41f6bb55942c5c9ed64d80bbe/traitsui/item.py#L91-L93

nicolasap-dm:
(I haven't followed the whole logic there, so they may be completely misguided. On the surface, however, the snippets _look_ relevant.)

aaronayres35:
Looking into this a bit it seems as though the separator is there, it just gets covered by a subsequent group(?).
The last separator which is followed by a Label is shown even though it comes after a group. Also if you remove one of the separators between groups you can see that the 2 groups are shown closer together. i.e. the `Item("-")` is there separating the groups, but the corresponding sunken line is not visible

aaronayres35:
Note, as a workaround I believe moving the separator inside the group "fixes" this:
ie replacing
```
Group(Label('Caverns')),
Item('_'),
```
with 
```
Group(
            Label('Caverns'),
            Item('_'),
        ),
```
yields
<img width="207" alt="Screen Shot 2021-06-23 at 4 10 13 PM" src="https://user-images.githubusercontent.com/36972686/123168506-b5048d00-d43d-11eb-8d70-382355d4eeed.png">

aaronayres35:
I believe the error here is really whenever a separator is the only Item in a Group (even an implicit group).
[...]

PR #1702, opened 2021-06-24
Fix separator alignment

aaronayres35:
This issue was observed when working on #1691, and again while working on addressing #1312.

Previously I remember having run the example code in #1312 without this issue... I recently updated to macOS Big Sur, but I am unsure if that is related.  I do not know what changed to cause this, especially since it appears to be an off by one error (which somehow wasn't a problem before ?).  
In any case, running the example code given in #1312 without this fix I see:
pre_fix
and with this change:
post_fix

**Checklist**

### DIFF
--- a/traitsui/qt4/ui_panel.py
+++ b/traitsui/qt4/ui_panel.py
@@ -1010,13 +1010,10 @@ class _GroupPanel(object):
 
         # See if the layout is a grid.
         if row >= 0:
-            # Move to the start of the next row if necessary.
+            # Move to the start of the row if necessary.
             if col > 0:
                 col = 0
-                row += 1
 
-            # Skip the row we are about to do.
-            row += 1
 
             # Allow for the columns.
             if show_labels:


aaronayres35:
Looking closer at the screenshots it appears as though the separator below the "Below" label is unaffected (i.e. it was already in the right place before)
I suspect because it was already the last row in the Group

aaronayres35:
Also, with the changes in this PR I see this for the ArrayEditor example:
Screen Shot 2021-06-24 at 8 42 14 AM

aaronayres35:
@mdickinson rahul mentioned you may still be on an older version of macOS.   If that's the case and you get a chance could you run the example code:
```
from traits.api import *
from traitsui.api import *

class SeparatorTest(HasTraits):
    view = View(
        Label('Above'),
        Item('_'),
        Label('Below'),
        Item('_'),
        Group(Label('Caverns')),
        Item('_'),
        Group(Label('Dungeons')),
        Item('_'),
        Group(Label("Extra dungeons")),
        Item('_'),
        Label('Foundations')
    )

if __name__ == "__main__":
    separator = SeparatorTest()
    separator.configure_traits()
```

On both master / this branch to see what looks correct.

mdickinson:
Here's what I see with PySide2 / Python 3.6 / EDM env on macOS 10.15.7.

First on master:

Screenshot 2021-07-23 at 09 24 44

Then on this branch:

Screenshot 2021-07-23 at 09 25 16

mdickinson:
Sorry, that second image is wrong (I forgot that etstool doesn't make an editable install by default). Here's the correct image
Screenshot 2021-07-23 at 09 28 27

In short: everything looks good on macOS 10.15.

aaronayres35:
Thank you @mdickinson!

rahulporuri (approved):
I ran the script that generates screenshots locally on windows and everything looks good on this branch. Sorry for the glacial pace of PR review

PR #1705, opened 2021-06-28
[WIP] add support for {enabled/visible}_when for Tabbed

### LINKED ISSUE #758: visible_when=False in Vgroup Tabbed  stiil visible
Hi there,

I'm facing the following issue.

If i set 2 traits items and set one of them to be not visible, it is stil visible.

I join a CME to be more explicit : Tab #1 should be not visible.

What am I doing wrong?

Thanks in advance.

Regards

PS : Debian 10 x86_64 / Python 3.7.3, with Traits libs from git source (Traits 6.1.0 dev1414 & TraitsUI 6.2.0 dev123)

---------------------------------------------------

#! /usr/bin/env python3

from traits.api import Int, Float, HasTraits
from traitsui.api import (Group, HGroup, Item, Tabbed, VGroup, View)

class Foo(HasTraits):

    a = Int
    b = Float

    view = View(Tabbed(VGroup(HGroup(Item('a')),
                              label='Tab #1',
                              visible_when='False'),
                       VGroup(HGroup(Item('b')),
                              label='Tab #2',
                              visible_when='True')))

if (__name__ == '__main__'):
    foo = Foo()
    foo.configure_traits()

aaronayres35:
This will close #758

Note this is a new feature (so clearly a change in behavior).
For example, running the example given in the issue replacing `visible_when` with `enabled_when`, on master you see:
enabled_master
whereas on this branch you see:
enabled_branch

**Checklist**

### DIFF
--- a/traitsui/qt4/ui_panel.py
+++ b/traitsui/qt4/ui_panel.py
@@ -28,7 +28,7 @@ import re
 
 from pyface.qt import QtCore, QtGui
 
-from traits.api import Any, HasPrivateTraits, Instance, Undefined
+from traits.api import Any, HasPrivateTraits, Instance, Undefined, List, Property
 from traits.observation.api import match
 
 from traitsui.api import Group
@@ -293,7 +293,7 @@ def panel(ui):
     return panel
 
 
-def _fill_panel(panel, content, ui, item_handler=None):
+def _fill_panel(panel, content, ui, item_handler=None, _visible_when_groups=None, _enabled_when_groups=None):
     """Fill a page based container panel with content.
     """
     active = 0
@@ -306,7 +306,6 @@ def _fill_panel(panel, content, ui, item_handler=None):
         if isinstance(item, Group):
             if item.selected:
                 active = index
-
             gp = _GroupPanel(item, ui, suppress_label=True)
             page = gp.control
             sub_page = gp.sub_control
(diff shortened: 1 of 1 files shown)


traitsui/qt4/ui_panel.py
```diff
@@ -1300,6 +1324,95 @@ class TabbedFoldGroupEditor(GroupEditor):
...
+                    # check that the tab for this widget is not already showing
+                    if self.container.indexOf(widget) == -1:
+                        method_to_call(idx, widget, label)
+                else:
+                    method_to_call_name = "remove" + tab_or_item
+                    method_to_call = getattr(
+                        self.container, method_to_call_name
+                    )
+                    # check that the tab for this widget is already showing
+                    if self.container.indexOf(widget) != -1:
```
aaronayres35, line 1408: theoretically, if the widget is visible, `idx == self.container.indexOf(widget)`.  However, perhaps the tab could change its index during the lifetime of the tab widget, in which case we should call `method_to_call(self.container.indexOf(widget))` instead?

corranwebster: I think that this comment raises a significant issue that needs to be considered: the index of the widget in the `Tabbed` group doesn't necessarily match the index of the Qt tab, because visibility is controlled by adding/removing the widget from the container.

I think things will go wrong in situations like the following:
- have tabs T0, T1, T2, T3, all initially visible
- T1 is hidden, so we get T0, T2, T3, but now the indices of T2 and T3 at the Qt level are 1 and 2, but at the TraitsUI level are 2 and 3.
- If we hide T2 in this state, the current code will actually hide T3 (and hiding T3 will result in an error?)
- After hiding T2 we have T0 and T2 at the Qt level.  If we un-hide T2, then it will try to insert it at Qt index 2 with the current code, and so we would get no change as it is already showing

I think there are situations where you can get things showing up in pretty much arbitrary order (eg. I think hide T2, hide T1, show T2 ends up with T0, T3, T2 as the layout).

I could be missing something about indices are being computed, but it looks like you can't assume that TraitsUI and Qt-level indices match, and computing the right Qt level index for re-insertion needs some thought (I think we need to work out how many TraitsUI tabs before the tab of interest are visible).

corranwebster (commented):
I think that the logic around the correspondence between Qt and TraitsUI indices can go wrong in complex scenarios of showing/hiding tabs

Issue #1533, opened 2021-03-10
Skip Wx test failures / dont allow CI to fail


aaronayres35:
Currently, the test suite is allowed to fail on wx, but this adds the risk of us introducing regressions without realizing it (merging green PRs that actually introduce new wx failures).  

We want to halt this now, skipping any currently failing tests on wx (they can be fixed over time), but requiring the remaining test suite to pass.  This way we will catch new failures before the causing code gets merged.

See: https://github.com/enthought/pyface/pull/900#pullrequestreview-608287222

PR #1707, opened 2021-06-29
UITester support for TableEditor

aaronayres35:
After looking though these changes again, I felt like this branch (same as #1706) really was a good starting point.  I went ahead and rebased on top of master, and will clean up this PR as needed with the intention that this ultimately will be merged.

I'm struggling to remember why I hadn't opened a PR for this previously.  I suspect it was that the wx implementation is incomplete.  Qt I _believe_ is fully functional (?)
wx has allowed failures on CI, and many TableEditor tests already fail. I remember this making adding functionality difficult because even if I thought i had wx support working tests may have still failed for other reasons...

EDIT: I am going to remove any wx changes in this PR and defer adding wx support for TableEditor to a later PR
**Checklist**


traitsui/testing/tester/_ui_tester_registry/qt4/_interaction_helpers.py
```diff
@@ -205,6 +205,127 @@ def mouse_click_item_view(model, view, index, delay):
...
+    """
+    check_q_model_index_valid(index)
+    widget = view.indexWidget(index)
+    if widget is None:
+        raise Disabled(
+            "No editable widget for item at row {!r} and column {!r}".format(
+                index.row(), index.column()
+            )
+        )
+    QTest.keyClicks(widget, sequence, delay=delay)
```
aaronayres35, line 270: Perhaps we need to call `check_key_compat` here like we do in `key_sequence_textbox`?  I think I recall this being to keep features the same with wx (so tests can remain agnostic).

For TraitsUI tests this is good and make sense.  However, a downstream user may only care about qt so the features could be nice.  I believe qt could do things like `\b` for backspace whereas to preserve wx/qt compat we now use "Backspace", etc.

traitsui/testing/tester/_ui_tester_registry/qt4/_interaction_helpers.py
```diff
@@ -205,6 +205,127 @@ def mouse_click_item_view(model, view, index, delay):
...
+    index : int
+
+    Raises
+    ------
+    LookupError
+        If the index cannot be located.
+        Note that the index error provides more
+    """
+    check_q_model_index_valid(index)
+    return model.data(index, QtCore.Qt.DisplayRole)
```
aaronayres35, line 326: is `data` the right thing here for displayed text? (look at Qt docs for `QAbstractItemModel`)

aaronayres35:
On master the following TableEditor Tests fail on wx:
```
test_filtered_table_editor
test_table_editor
test_table_editor_select_cell
test_table_editor_select_cell_index
test_table_editor_select_cell_indices
test_table_editor_select_cells
test_table_editor_select_column
test_table_editor_select_column_index
test_table_editor_select_column_indices
test_table_editor_select_columns
test_table_editor_select_row
test_table_editor_select_row_index
test_table_editor_select_row_indices
test_table_editor_select_rows
```

(Note this is effectively every test - I run `python -m unittest traitsui/tests/editors/test_table_editor.py -v` in an env created with `python etstool.py install --toolkit=wx --editable` and I see 16 tests run with 14 errors and 2 skipped tests. 

Given this, I intend to skip all these tests on wx.  Note this is now a part of #1533.  Even further, I intend to rewrite the tests to use the new UI Tester support on Qt.  Thus, even if the TableEditor issues are fixed, we will need to follow through with the UI Tester support to simply stop skipping these tests on wx.  Either that or use this PR as a ref and revert the tests back to their old non UITester using state. I am doing this based off the following section of the developer guide in the traitsui docs: https://docs.enthought.com/traitsui/traitsui_dev_guide/testing.html#where-are-the-tests

traitsui/tests/editors/test_table_editor.py
```diff
@@ -280,216 +300,282 @@ def setUp(self):
...
 
         self.assertIs(selected, object_list.values[5])
+        self.assertIs(object_list.selected, selected)
 
-    @requires_toolkit([ToolkitName.qt, ToolkitName.wx])
     def test_table_editor_select_rows(self):
         object_list = ObjectListWithSelection(
             values=[ListItem(value=str(i ** 2)) for i in range(10)]
         )
         object_list.selections = object_list.values[5:7]
```
aaronayres35, line 343: would be nice to have shift click functionality enabled to do this type of thing...

aaronayres35, line 343: It would be nice to have shift click functionality to do this sort of thing with the tester...

traitsui/tests/editors/test_table_editor.py
```diff
@@ -280,216 +300,279 @@ def setUp(self):
...
-                selected = editor.selected_indices
-            elif is_wx():
-                selected = editor.selected_row_index
-
-            process_cascade_events()
+        tester = UITester()
+        with tester.create_ui(object_list, dict(view=select_row_index_view)) \
+                as ui:
+            editor = tester.find_by_name(ui, "values")._target
+            selected = editor.selected_indices
```
aaronayres35, line 362: Perhaps we should add a `SelectedIndices` query as well?

traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/table_editor.py
```diff
@@ -0,0 +1,117 @@
...
+    registry.register_interaction(
+        target_class=SimpleEditor,
+        interaction_class=Selected,
+        handler=lambda wrapper, _: wrapper._target.selected
+    )
+    registry.register_interaction(
+        target_class=SimpleEditor,
+        interaction_class=SelectedIndices,
+        handler=lambda wrapper, _: wrapper._target.selected_indices
+    )
```
aaronayres35, line 117: I've just added a `SelectedIndices` query class.
For the current implementation, for `TableEditor`, `Selected` and `SelectedIndices` effectively just serve as proxies to the `selected` and `selected_indices` traits on the table editor (see code above.)
In theory it may be better for these to dive down to the Qt level (?)

aaronayres35: This is one of the weird things with UITester support though IMO.  We want to test traitsui editors are working correctly.
So having tests be down at qt level is really what we want. Ie I do something in traitsUI and we verify that does what we expect down in qt land.
However, say I were implementing the handler for the TableEditor to do a `SelectedIndices` query. I would end up basically rewriting:
https://github.com/enthought/traitsui/blob/347f0b5df25b4432b1163c34ebfd8a99c5a26ae0/traitsui/qt4/table_editor.py#L626-L643
It creates sort of a weird chasing your own tail scenario 🤔 

For downstream users they just want to see that when they run _their_ code, ____ is selected.  For which this (just using traits on the editor) is perfectly fine.

traitsui/tests/editors/test_table_editor.py
aaronayres35: EDIT: done

traitsui/testing/tester/_ui_tester_registry/qt4/_interaction_helpers.py
aaronayres35: Given that this support is now going to be on qt any way this concern doesn't really seem important

aaronayres35:
I believe once this PR gets the green light, a lot of the code can be pulled to add support for TabularEditor which also ends up wrapping a `QTableView`.  Likewise that will probably get only qt support first as wx tests are currently all failing.

traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/table_editor.py
```diff
@@ -0,0 +1,117 @@
...
+from traitsui.testing.api import (
+    Cell,
+    DisplayedText,
+    MouseClick,
+    MouseDClick,
+    KeyClick,
+    KeySequence,
+    Selected,
+    SelectedIndices,
+)
```
aaronayres35, line 22: I shouldn't be using the api module for these imports

traitsui/testing/api.py
```diff
@@ -73,11 +77,14 @@
...
     IsEnabled,
     IsVisible,
+    Selected,
+    SelectedIndices,
     SelectedText
 )
 
 # Locations (for locating GUI elements)
 from .tester.locator import (
+    Cell,
```
rahulporuri, line 87: can you add this and `MouseDClick` to the module docstring?

traitsui/testing/tester/query.py
```diff
@@ -16,6 +16,30 @@
 """
 
 
+class Selected:
+    """ An object representing an interaction to obtain the object which is
+    currently selected.
```
rahulporuri, line 21:
```suggestion
    """ Represents an interaction to obtain the currently selected object.
```

traitsui/testing/tester/query.py
```diff
@@ -16,6 +16,30 @@
...
+
+    Implementations should return the selected object, or None if nothing is
+    selected.
+    """
+    pass
+
+
+class SelectedIndices:
+    """ An object representing an interaction to obtain the indices of objects
+    which are currently selected.
```
rahulporuri, line 31:
```suggestion
    """ Represents an interaction to obtain the indices of the currently
    selected objects.
```

traitsui/testing/tester/query.py
```diff
@@ -16,6 +16,30 @@
...
+
+    Implementations should return either a list of indicies of the selected
+    objects, a single index of a lone selected object, or -1 if nothing is
+    selected.
```
rahulporuri, line 35: What's the rationality behind this API? Why not just return `None` if nothing is selected? Also, why not always return a list - which might contain one or more elements depending on the size of the selection.

traitsui/testing/tester/query.py
```diff
@@ -16,6 +16,30 @@
...
+
+    Note that an index could be an integer, or it could be a tuple e.g.
+    corresponding to a specific cell at some (row, column) in a table.
```
rahulporuri, line 38: under what circumstances do we return an integer vs a tuple.

traitsui/testing/tester/query.py
aaronayres35: TBH it was because I am currently just mirroring the implementation of `selected_indices` in the `TableEditor` itself. See comment: https://github.com/enthought/traitsui/pull/1707#discussion_r661660617 and code:
https://github.com/enthought/traitsui/blob/762f66b146b676aaa3b7f1dfcc3d65118976822f/traitsui/qt4/table_editor.py#L627-L643

I agree this is confusing though and lazy.  I can at the very least wrap this with a special handler to give a more consistent return type.  The `SelectedIndices` query class has no reason to be at all tied with a specific Editor (in fact it definitely shouldn't be).

Nonetheless I do still have the question of whether or not I should be reimplementing/rewriting the above code in the UITester TableEditor support code as a specifc handler for the `SelectedIndices` query.  As mentioned in my comment I think that is a little unnecessary / we have this strange chase your own tail scenario... 🤔

aaronayres35: I need to carefully think about the best api for Selected and SelectedIndices before pushing this through 🤔 

It i tricky because with a TableEditor, one could have rows, columns, cells, etc. selected.  The api needs to accommodate this use case but also be general and not overly complex for other Editors to use

aaronayres35: I am going to update `Selected` so that it also always returns a list (either a singleton list for one selection, list of all objects selected, or an empty list if nothing is selected).
It is also a bit strange with `SelectedText` as in that case it doesn't really make sense to want to make that query if multiple objects are selected (at least I can't imagine a use case).  Hence for that I think I will not update it to enforce that it returns a list.

traitsui/testing/tester/query.py
aaronayres35: Currently, it would be an int if selecting entire rows or columns, whereas it would be a tuple if selecting specific cells.  However, if used with say a ListEditor, InstanceEditor, or EnumEditor, a selected index would likely be an integer.  Currently those implementations do not exist (in current tests querying what is selected we just use `SelectedText` rather than inspecting the object itself that is selected).  These implementations can be added in the future though.
I've updated the doc string to try to make this more clear.

traitsui/tests/editors/test_table_editor.py
```diff
@@ -280,216 +319,219 @@ def setUp(self):
...
+        tester = UITester()
+        with tester.create_ui(object_list, dict(view=select_row_view)) as ui:
+            # click the first cell in the 6th row to select the row
+            values_table = tester.find_by_name(ui, "values")
+            row6_cell = values_table.locate(Cell(5, 0))
+            row6_cell.perform(MouseClick())
+            selected = values_table.inspect(Selected())
 
-            process_cascade_events()
+        self.assertEqual(selected, [object_list.values[5]])
```
aaronayres35, line 355: I find the list here _slightly_ clunkier than before... However, i think being consistent with the api of `SelectedIndices` is more important.

traitsui/testing/tester/_ui_tester_registry/qt4/_interaction_helpers.py
corranwebster: Yes, `data` with `DisplayRole` should provide the text to display.

traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/table_editor.py
corranwebster: `UITester` strikes me as being useful for two things:
- testing that TraitsUI editors are integrating correctly with the toolkit
- as a framework for integration tests for TraitsUI apps

For unit tests of applications you can usually get away with tests at the traits level (does changing this trait in a UI change the corresponding thing in the model? or vice-versa) where you don't really care/can trust that the UI then does the right thing.

corranwebster (approved):
Looks good at a high level.  From examples of usage, looks like it is reasonably user-friendly.

PR #1709, opened 2021-06-30
Add  ``update_column_widths`` Event trait to force calls to resizeColumnsToContents

### LINKED ISSUE #1434: TabularEditor triggering column size recompute
Would like to link a trait listener to the `TabularEditor` that triggers a recompute of the column widths. I tried `refresh` and `update`, and both seemed not to recompute the column widths.

Specifically, I want to call `_TableView` `resizeColumnsToContents` within `traitsui/qt4/tabular_editor.py`  whenever `update_editor`  of `TabularEditor` (`traitsui/qt4/tabular_editor.py`) is triggered using the update trait listener on traitsui’s `TabularEditor` .

Or there could be a new listener `update_column_widths` in the traitsui `TabularEditor` that triggers `resizeColumnsToContents`?

aaronayres35:
closes #1434 

This PR adds a new `update_column_widths` `Event` trait to the qt4 `TabularEditor`, along with a listener that will call `self.control.resizeColumnsToContents` whenever this event is fired.  

Code I am using to manually test this:
<details>

```
from numpy import sqrt
from numpy.random import random

from traits.api import HasTraits, Property, Array, Any, Instance, Str
from traitsui.api import View, Item, TabularAdapter, TabularEditor, ModelView, TextEditor
from traitsui.testing.api import UITester


# -- Tabular Adapter Definition -------------------------------------------
class ArrayAdapter(TabularAdapter):

    columns = [('i', 'index'), ('x', 0), ('y', 1), ('z', 2)]

    font = 'Courier 10'
    alignment = 'right'
    format = '%.4f'  # Str('%.4f', update=True)

    index_text = Property()
    index_image = Property()

    def _get_index_text(self):
        return str(self.row)

    def _get_index_image(self):
        x, y, z = self.item
        if sqrt((x - 0.5) ** 2 + (y - 0.5) ** 2 + (z - 0.5) ** 2) <= 0.25:
            return '@icons:red_ball'

        return None



class ShowArray(ModelView):

    data = Array

    form = Str('%.4f')

    def object_form_changed(self, info):
        info.data.adapter.format = self.form
        info.data.update_column_widths = True

    traits_view = View(
        Item(
            'form',
            editor=TextEditor(auto_set=False)
        ),
        Item(
            'data',
            show_label=False,
            editor=TabularEditor(
                adapter=ArrayAdapter(),
                auto_resize=True,
                # Do not allow any kind of editing of the array:
                editable=False,
                operations=[],
                drag_move=False
            )
        ),
        title='Array Viewer',
        width=0.3,
        height=0.8,
        resizable=True
    )


# Create the demo:
demo = ShowArray(model=HasTraits(), data=random((2, 3)))

# Run the demo (if invoked from the command line):
if __name__ == '__main__':
    demo.configure_traits()
```

</details>

**Checklist**

### DIFF
--- a/traitsui/qt4/tabular_editor.py
+++ b/traitsui/qt4/tabular_editor.py
@@ -30,6 +30,7 @@ from traits.api import (
     Int,
     List,
     NO_COMPARE,
+    observe,
     Property,
     TraitListEvent,
 )
@@ -71,6 +72,9 @@ class TabularEditor(Editor):
 
     #: The event fired when a simple repaint is needed:
     refresh = Event()
+
+    #: The event fired when a recompute of column widths is needed
+    update_column_widths = Event()
 
     #: The current set of selected items (which one is used depends upon the
     #: initial state of the editor factory 'multi_select' trait):
@@ -534,6 +538,10 @@ class TabularEditor(Editor):
         self.control.scrollTo(
             self.model.index(max(self.selected_row, 0), column), scroll_hint
         )
+
+    @observe("update_column_widths", post_init=True)
+    def _resize_columns_on_event(self, event):
+        self.control.resizeColumnsToContents()
 
     # -- Table Control Event Handlers -----------------------------------------
 


aaronayres35:
Actually thinking about this more I do not like this approach... In my example code, one could replace the line `info.data.update_column_widths = True` with `info.data.control.resizeColumnsToContents()` and then the new Event trait is useless.

Any time a user is trying to fire the `update` Event Trait, they must have access to the editor.  I am unsure why they couldn't just manually call `editor.control.resizeColumnsToContents()` then at their own leisure.


Additionally, there exists an `auto_resize` trait that _seems_ like it should be dictating this sort of behavior.

aaronayres35:
@rahulporuri This PR isn't really ready for review if it is ready to merge, but I've opened it to get your ideas on some different approaches / so we can further discuss what makes sense

PR #1713, opened 2021-07-01
[WIP] UITester support for qt TreeEditor

aaronayres35:
This PR adds UITester support for the qt `TreeEditor`.  Note that some of the changes made here are identical to those made in #1707 (ie _interaction_helpers that work with `QAbstractItemModel`). This PR also currently adds a simple test for the `TreeEditor_demo` but I have not yet updated actual Tree Editor tests.

**Checklist**

### DIFF
--- a/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/tree_editor.py
+++ b/traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/tree_editor.py
@@ -10,8 +10,16 @@
 #
 
 from traitsui.qt4.tree_editor import SimpleEditor
-from traitsui.testing.tester.qt4 import helpers
-from traitsui.testing.tester import command, locator, registry_helper, query
+
+
+from traitsui.testing.tester.command import (
+    MouseClick, MouseDClick, KeyClick, KeySequence
+)
+from traitsui.testing.tester.locator import TreeNode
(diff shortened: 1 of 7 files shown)


traitsui/testing/tester/_ui_tester_registry/qt4/_traitsui/tree_editor.py
```diff
@@ -0,0 +1,125 @@
...
+        i_rows = iter(self.location.row)
+        item = tree_widget.topLevelItem(next(i_rows))
+        for i_row in i_rows:
+            item = item.child(i_row)
+        q_model_index = tree_widget.indexFromItem(item, i_column)
+        return dict(
+            model=tree_widget.model(),
+            view=tree_widget,
+            index=q_model_index,
+        )
```
aaronayres35, line 84: This logic and the `TreeNode` object having a `row`/`column` I need to think about more carefully.  I had this code in a branch locally from a long time ago which I believe had originally been pulled from one of Kit's draft PRs.

There may be a simple way to do this / at the very least I need to better document the TreeNode class to say what row and column actually specify.  In this case (see test_TreeEditor_demo.py) row is a tuple containing the index of the node of interest at subsequent levels of the tree, and column is the index at that last level (?).  I have just been using column as 0, ... it may be possible we only need one single tuple to do this TreeNode location.

PR #1714, opened 2021-07-02
Use ubuntu-latest instead of ubuntu-18.04

### LINKED ISSUE #1660: Use Ubuntu latest in integration-tests cron job
see PR #1652 and comment: https://github.com/enthought/traitsui/pull/1652#discussion_r629049992

We currently need to use swig 3 because of https://github.com/enthought/enable/issues/360
Ideally though, we could do so while still using the latest ubuntu version.  In the short term, we followed the lead of enable by just using ubuntu-18.04 which uses swig 3 by default. 

A similar update would be nice in enable, or better yet, fixing 360 so that this can be avoided altogether (or the more likely, migration to celiagg instead of agg so this becomes obsolete)

rahulporuri:
We were unable to use `ubuntu-latest` earlier because it shipped with swig 4 by default and enable didnt support swig 4. But the issue got fixed in enable so we should be able to use `ubuntu-latest` now.

Fixes #1660 
Xref https://github.com/enthought/enable/issues/360

**Checklist**

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -12,9 +12,7 @@ jobs:
   test:
     strategy:
       matrix:
-        # Note that Ubuntu 20 cannot be used because its SWIG version is 4.0
-        # (enthought/enable#360)
-        os: [ubuntu-18.04]
+        os: [ubuntu-latest]
         toolkit: ['pyside2']
         # Dependencies include chaco/enable, whose installation fails
         # on Python 3.8 and 3.9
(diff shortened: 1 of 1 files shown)


rahulporuri:
Once this PR is merged, we can make the same changes in other ETS repositories which aren't using `ubuntu-latest`

aaronayres35 (approved):
Changes LGTM

Although integration-tests workflow was not run on this PR.  I see you added `workflow_disptach` so the workflow can be manually triggered, but I'm not seeing it as having been run on the actions tab for the repo.  It is possible I am missing it, but if it hasn't been run we should trigger it just to make sure it comes out 🟢

PR #1717, opened 2021-07-06
Stop using "enthought_sphinx_theme" from git/github source

rahulporuri:
This PR updates the ci utility to stop using `enthought_sphinx_theme` from git/github and instead just installs it from edm instead. In fact, it looks like we are already installing `sphinx` and `enthought_sphinx_theme` via edm and then trying to pip install them again later.

**Checklist**


aaronayres35 (approved):
LGTM

Although I'm not sure where we were trying to pip install these deps again later as you mentioned.  It seems to me doc-requirements.txt was previously unused?

I built and served docs locally and everything looks correct.

rahulporuri:
> Although I'm not sure where we were trying to pip install these deps again later as you mentioned. It seems to me doc-requirements.txt was previously unused?

Yeah. It looks like all of the src requirements files can now be removed. I'll do that separately.

PR #1719, opened 2021-07-08
fix Dynamic_EnumEditor_demo.py on qt

### LINKED ISSUE #1676: Dynamic Enum Editor not working as expected
I recently upgraded to macOS Big Sur and when running the `Dynamic_EnumEditor_demo.py` things are not working as expected on pyqt5 (I don't remember seeing this issue before I had upgraded, but it is certainly possible I missed it).

On wx, running the example and unselecting some options from the checklist, when I click the combo box I see:
Screen Shot 2021-05-21 at 10 19 57 AM

As expected. However, with pyqt5 I see:
Screen Shot 2021-05-21 at 10 24 29 AM
If I then select one of the items in the combobox which has been unselected from the checklist, it shows up red:
Screen Shot 2021-05-21 at 10 20 42 AM

aaronayres35:
closes #1676 

For thee full saga see this comment:
https://github.com/enthought/traitsui/issues/1676#issuecomment-876585259

This PR fixes the example, by fixing a bug in EnumEditor itself.  In order to fully fix the issue, it also additionally updates the example to use an `Instance` rather than a `Constant` trait, since `Constant` doesn't seem to work with observe (?)
I will write up a MWE and open an issue (which can quickly be closed if this is in fact intentional).

**Checklist**

### DIFF
--- a/traitsui/examples/demo/Advanced/Dynamic_EnumEditor_demo.py
+++ b/traitsui/examples/demo/Advanced/Dynamic_EnumEditor_demo.py
@@ -57,7 +57,7 @@ Notes:
 """
 
 
-from traits.api import Enum, HasPrivateTraits, List, Constant
+from traits.api import Enum, HasPrivateTraits, Instance, List
 
 from traitsui.api import (
     View, Item, VGroup, HSplit, EnumEditor, CheckListEditor
@@ -97,7 +97,7 @@ class OrderMenu(HasPrivateTraits):
     entree = Enum(values='capabilities.available')
 
     # Reference to the restaurant's current entree capabilities:
-    capabilities = Constant(kitchen_capabilities)
+    capabilities = Instance(kitchen_capabilities, args=())
 
     # The user interface view:
     view = View(
--- a/traitsui/examples/demo/Advanced/Dynamic_EnumEditor_demo.py
+++ b/traitsui/examples/demo/Advanced/Dynamic_EnumEditor_demo.py
@@ -97,7 +97,10 @@ class OrderMenu(HasPrivateTraits):
     entree = Enum(values='capabilities.available')
 
     # Reference to the restaurant's current entree capabilities:
-    capabilities = Instance(kitchen_capabilities, args=())
+    capabilities = Instance(KitchenCapabilities)
+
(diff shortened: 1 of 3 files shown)


traitsui/qt4/enum_editor.py
```diff
@@ -139,7 +139,7 @@ def _get_inverse_mapping(self):
 
     # Trait change handlers --------------------------------------------------
 
-    def _values_changed(self, event=None):
+    def _values_updated(self, event=None):
```
aaronayres35, line 142: naming of methods here is ugly - there was a `values_changed` and `_values_changed` method, neither of which was a magically named listener since `values` is a trait on the factory, not on this object.  
I've made this update here, but we likely want to do some more renaming.  I'm happy to add that here or in a follow up PR - whichever a reviewer prefers.

traitsui/examples/demo/Advanced/Dynamic_EnumEditor_demo.py
```diff
@@ -97,7 +97,7 @@ class OrderMenu(HasPrivateTraits):
     entree = Enum(values='capabilities.available')
 
     # Reference to the restaurant's current entree capabilities:
-    capabilities = Constant(kitchen_capabilities)
+    capabilities = Instance(kitchen_capabilities, args=())
```
rahulporuri, line 100: Shouldn't we be using the class instead of the instance here because we're setting `args=()`? A meaningful alternative would be to return `kitchen_capabilities` in a `_capabilities_default` method.

```suggestion
    capabilities = Instance(KitchenCapabilities, args=())
```

traitsui/qt4/enum_editor.py
rahulporuri: im not sure why we need to set `event=None` here

```suggestion
    def _update_values_and_rebuild_editor(self, event=None):
```

rahulporuri (approved):
🎉  LGTM with a couple of comments

traitsui/qt4/enum_editor.py
aaronayres35: ah you are right it is only `values_changed` that is called stand alone

PR #1726, opened 2021-07-12
add an expand_all method to TreeEditor

### LINKED ISSUE #673: Expanding tree nodes in qt tree editor causes attribute error
When attempting to call `expandAll` on the Qt tree editor widget, the minimal example below fails with traceback:
```bash
  File ".../lib/python3.6/site-packages/traitsui/qt4/tree_editor.py", line 1974, in paint
    expanded, node, instance = self.editor._get_node_data(item)
  File ".../lib/python3.6/site-packages/traitsui/qt4/tree_editor.py", line 893, in _get_node_data
    return nid._py_data
AttributeError: 'QTreeWidgetItem' object has no attribute '_py_data'

```

```python
from traits.api import *
from traitsui.api import *


class Cheese(HasTraits):
    name = Str


class Taco(HasTraits):
    name = Str
    cheeses = List(Cheese)


class Meal(HasTraits):
    tacos = List(Taco)

    def _tacos_default(self):
        return [
            Taco(name='crunchy',
                 cheeses=[Cheese(name='cheddar'), Cheese(name='cotija')]),
            Taco(name='greasy',
                 cheeses=[Cheese(name='parmesan'), Cheese(name='colby')]),
            Taco(name='grumpy',
                 cheeses=[Cheese(name='fresco')]),
        ]


class MenuHandler(Handler):
    def object_expand_all_changed(self, info):
        editor = info.ui.get_editors('meal')[0]
        editor._tree.expandAll()


class Menu(HasTraits):
    meal = Instance(Meal, ())
    expand_all = Button

    traits_view = View(
        UItem(
            'meal',
            editor=TreeEditor(
                editable=False,
                nodes=[
                    TreeNode(
                        node_for=[Meal],
                        label='=Meal',
                        children='tacos',
                        auto_open=False,
                    ),
                    TreeNode(
                        node_for=[Taco],
                        label='name',
                        children='cheeses',
                    ),
                    TreeNode(
                        node_for=[Cheese],
                        label='name',
                    ),
                ]
            ),
        ),
        UItem('expand_all'),
        handler=MenuHandler(),
    )


if __name__ == '__main__':
    Menu().configure_traits()
```

aaronayres35:
closes #673 
This PR took inspiration from the `expand_levels` method here:
https://github.com/enthought/traitsui/blob/fc3b8ac92eea79efba923ec2387f292c281eddf3/traitsui/qt4/tree_editor.py#L263-L273
and the second response here https://stackoverflow.com/questions/28829192/how-to-get-the-number-of-items-of-a-qtreewidget suggesting to make use of https://doc.qt.io/qt-5/qtreewidgetitemiterator.html#details for easy traversal of a `QTreeWidget`.

I found a lot of the TreeEditor internals hard to follow.  TraitsUI does its own thing setting `node_data` which takes the form `(expanded, node, object)`.  I am unclear on why expanded is included here as it could be directly accessed on the nid object via `nid.isExpanded()`.  However, this is very prevalent throughout the code so I left things as is for now rather than attempting a large refactor.  In the implementation in this PR I follow the lead of `expand_levels` to try to make use of the traitsUI way to expand nodes.  ie. calling the editors `_expand_node` method _and_ calling `setExpanded(True)` on the `nid` (which is a `QTreeWidgetItem`.  

To test this code, I used the code provided on the issue, with a very slight modification - I changed the line `editor._tree.expandAll()` to `editor.expand_all()`.

**Checklist**

### DIFF
--- a/traitsui/qt4/tree_editor.py
+++ b/traitsui/qt4/tree_editor.py
@@ -272,6 +272,21 @@ class SimpleEditor(Editor):
                 for cnid in self._nodes_for(nid):
                     self.expand_levels(cnid, levels - 1)
 
+    def expand_all(self):
+        """ Expands all expandable nodes in the tree.
+
+        Warning: If the Tree contains a large number of items, this function
+        will be very slow.
+        """
+        iterator = QtGui.QTreeWidgetItemIterator(self._tree)
+        while iterator.value():
+            item = iterator.value()
+            expanded, node, object = self._get_node_data(item)
+            if self._has_children(node, object):
+                self._expand_node(item)
+                item.setExpanded(True)
+            iterator += 1
+
     def update_editor(self):
         """ Updates the editor when the object trait changes externally to the
             editor.
--- a/traitsui/qt4/tree_editor.py
+++ b/traitsui/qt4/tree_editor.py
@@ -278,6 +278,8 @@ class SimpleEditor(Editor):
         Warning: If the Tree contains a large number of items, this function
         will be very slow.
         """
+        # For more info on the QtGui.QTreeWidgetItemIterator object,
+        # see https://doc.qt.io/qt-5/qtreewidgetitemiterator.html
         iterator = QtGui.QTreeWidgetItemIterator(self._tree)
         while iterator.value():
             item = iterator.value()
--- a/traitsui/tests/editors/test_tree_editor.py
+++ b/traitsui/tests/editors/test_tree_editor.py
@@ -11,8 +11,9 @@
 import unittest
 
 from pyface.api import GUI
-from traits.api import Bool, HasTraits, Instance, List, Str
+from traits.api import Bool, Button, HasTraits, Instance, List, Str
 from traitsui.api import (
+    Handler,
     Item,
     ObjectTreeNode,
     TreeEditor,
@@ -20,6 +21,7 @@ from traitsui.api import (
     TreeNodeObject,
     View,
 )
+from traitsui.testing.api import MouseClick, UITester
 
 from traitsui.tests._tools import (
     BaseTestMixin,
@@ -49,6 +51,12 @@ class Bogus(HasTraits):
         return BogusWrap(bogus_list=[Bogus()])
 
 
+class BogusHandler(Handler):
+    def object_expand_all_changed(self, info):
+        editor = info.ui.get_editors('bogus')[0]
+        editor.expand_all()
+
+
 class BogusTreeView(HasTraits):
     """ A traitsui view visualizing Bogus objects as trees. """
 
@@ -60,9 +68,12 @@ class BogusTreeView(HasTraits):
 
     nodes = List(TreeNode)
 
+    expand_all = Button()
+
     def _nodes_default(self):
         return [
-            TreeNode(node_for=[Bogus], children="bogus_list", label="=Bogus")
+            TreeNode(node_for=[Bogus], children="bogus_list", label="=Bogus"),
+            TreeNode(node_for=[BogusWrap], label='name')
         ]
 
     def default_traits_view(self):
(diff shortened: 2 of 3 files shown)


rahulporuri:
i'm noticing another error if i run the example in #673 on this branch - 

steps to reproduce the error : 
- launch the dialog with the tree by running the script
- click "Expand All"
- click "Expand All" again after the tree has already expanded.

```
  File "c:\users\rporuri\work\github\ets\traitsui\traitsui\qt4\tree_editor.py", line 852, in _on_item_expanded
    self._expand_node(nid)
  File "c:\users\rporuri\work\github\ets\traitsui\traitsui\qt4\tree_editor.py", line 495, in _expand_node
    for child in node.get_children(object):
TypeError: 'NoneType' object is not iterable
```

traitsui/qt4/tree_editor.py
```diff
@@ -272,6 +272,21 @@ def expand_levels(self, nid, levels, expand=True):
...
 
+    def expand_all(self):
+        """ Expands all expandable nodes in the tree.
+
+        Warning: If the Tree contains a large number of items, this function
+        will be very slow.
+        """
+        iterator = QtGui.QTreeWidgetItemIterator(self._tree)
+        while iterator.value():
+            item = iterator.value()
```
rahulporuri, line 283: I'm not sure how this iterator works but these lines threw me off completely. Code like this is usually broken because for normal python iterators we'll be iterating twice in a single loop, for each call to `iterator.value` but having tested the code in #673 using the fix you mentioned, I guess this works for this Qt iterator.

Please add any further links/documentation/comments you have regarding this iterator here - that'll hopefully help with understanding later on.

traitsui/tests/editors/test_tree_editor.py
```diff
@@ -49,6 +51,12 @@ def _wrapped_bogus_default(self):
         return BogusWrap(bogus_list=[Bogus()])
 
 
+class BogusHandler(Handler):
+    def object_expand_all_changed(self, info):
```
rahulporuri, line 55: It took me a moment to figure out the fact that this is a special method - hooked up to listen to the `expand_all` button trait on the `object` of this handler. Is that correct?

rahulporuri (approved):
LGTM with one minor question.

rahulporuri:
Also - please add a news fragment

rahulporuri:
@aaronayres35 can you take a look at the review/comments and add the news fragment? Once those things are done, this PR can be merged.

traitsui/tests/editors/test_tree_editor.py
aaronayres35: Yep, see this section of the docs which I used as reference https://docs.enthought.com/traitsui/traitsui_user_manual/handler.html#reacting-to-trait-changes

But magically named methods are generally bad so maybe I should avoid doing this... We need the `UIInfo` object though and the inner magic makes it so these magically named methods are passed it as an argument. (I should dig into the code that does this more, I was just going off the docs)

traitsui/qt4/tree_editor.py
aaronayres35: It is a weird qt thing, that does things its own way.  The iteration is actually performed by the `iterator += 1` line below.  See docs https://doc.qt.io/qt-5/qtreewidgetitemiterator.html#operator-2b-eq

aaronayres35: I will add a link to https://doc.qt.io/qt-5/qtreewidgetitemiterator.html in a comment

aaronayres35:
> @aaronayres35 can you take a look at the review/comments and add the news fragment? Once those things are done, this PR can be merged.

Sounds good, thanks for the review!

PR #1725, opened 2021-07-09
Fix selectable InstanceEditor combobox updates

aaronayres35:
closes #1641 
For full saga see comment https://github.com/enthought/traitsui/issues/1641#issuecomment-877321165

This solution works but I am not sure when one could unhook these observers (perhaps it isn't needed?)

~I will work on adding a regression test, although this _may_ require currently not existent UI Tester support for combobox of Instance Editor, I am not sure yet~

**Checklist**
- [x] Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)

PR #1728, opened 2021-07-13
Fix Qt InstanceEditor appearance when None selected

### LINKED ISSUE #1134: Qt InstanceEditor initially appears as if the first item is selected despite the value being None
The Qt custom `InstanceEditor` initially appears as if the first item is selected, however the selection is not actually made (the underlying value is `None`). Expected behavior is for the combo box to appear empty until a selection is made, which is what `EnumEditor` does.

aaronayres35:
closes #1134 

I've been testing this locally using:
```
from traits.api import Button, HasStrictTraits, Instance, List, Str, observe
from traitsui.api import EnumEditor, InstanceEditor, Item, VGroup, View


class Option(HasStrictTraits):
    name = Str()


class Model(HasStrictTraits):
    selected_option = Instance(Option)
    all_options = List(Instance(Option))
    reset_to_none = Button()
    change_options = Button()

    selected_name = Str()
    all_names = List(Str)

    view = View(
        VGroup(
            Item('selected_option', label='InstanceEditor',
                 editor=InstanceEditor(name='all_options', editable=False)),
            Item('selected_name', label='EnumEditor',
                 editor=EnumEditor(name='all_names')),
            Item('reset_to_none'),
            Item('change_options')
        ),
    )

    @observe('reset_to_none')
    def do_stuff(self, event):
        self.selected_option = None

    @observe('change_options')
    def do_other_stuff(self, event):
        self.all_options = [Option(name='Option A')]


if __name__ == "__main__":
    model = Model(
        all_options=[Option(name='Option A'), Option(name='Option B')],
        all_names=['Option A', 'Option B'],
    )
    model.configure_traits()
    print(f'Selected option: {model.selected_option}')
    print(f'Selected name: {model.selected_name}')
```
The `reset_to_none` and `change_options` buttons were to ensure both the `rebuild_items` and `update_editor` code paths could be run.  I believe the changes to both methods are necessary for this behavior to be correct.

~I will work on writing real unit tests for this~ EDIT: done. I basically just copied the above and used UITester

**Checklist**

### DIFF
--- a/traitsui/qt4/instance_editor.py
+++ b/traitsui/qt4/instance_editor.py
@@ -218,9 +218,10 @@ class CustomEditor(Editor):
         if name >= 0:
             choice.setCurrentIndex(name)
         else:
-            # Otherwise, current value is no longer valid, try to discard it:
+            # Otherwise, current value is no longer valid, set combobox empty
             try:
                 self.value = None
+                choice.setCurrentIndex(name)
             except:
                 pass
 
@@ -275,17 +276,21 @@ class CustomEditor(Editor):
         # Update the selector (if any):
         choice = self._choice
         item = self.item_for(self.value)
-        if (choice is not None) and (item is not None):
-            name = item.get_name(self.value)
-            if self._object_cache is not None:
-                idx = choice.findText(name)
-                if idx < 0:
-                    idx = choice.count()
-                    choice.addItem(name)
+        if choice is not None:
+            if item is not None:
+                name = item.get_name(self.value)
+                if self._object_cache is not None:
+                    idx = choice.findText(name)
+                    if idx < 0:
+                        idx = choice.count()
+                        choice.addItem(name)
 
-                choice.setCurrentIndex(idx)
+                    choice.setCurrentIndex(idx)
+                else:
+                    choice.setText(name)
             else:
-                choice.setText(name)
+                choice.setCurrentIndex(-1)
+
 
     def resynch_editor(self):
         """ Resynchronizes the contents of the editor when the object trait
--- a/traitsui/qt4/instance_editor.py
+++ b/traitsui/qt4/instance_editor.py
@@ -221,7 +221,7 @@ class CustomEditor(Editor):
             # Otherwise, current value is no longer valid, set combobox empty
             try:
                 self.value = None
-                choice.setCurrentIndex(name)
+                choice.setCurrentIndex(-1)
             except:
                 pass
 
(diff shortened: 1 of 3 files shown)


traitsui/qt4/instance_editor.py
```diff
@@ -218,9 +218,10 @@ def rebuild_items(self, event=None):
         if name >= 0:
             choice.setCurrentIndex(name)
         else:
-            # Otherwise, current value is no longer valid, try to discard it:
+            # Otherwise, current value is no longer valid, set combobox empty
             try:
                 self.value = None
+                choice.setCurrentIndex(name)
```
rahulporuri, line 224: So this is effectively `choice.setCurrentIndex(-1)`. Right?

traitsui/tests/editors/test_instance_editor.py
```diff
@@ -185,16 +205,12 @@ def test_custom_editor_with_selection_change_option_name(self):
             # test that the current object is None
             self.assertIsNone(obj.inst)
 
-            # test that the displayed text is correct
-            instance = tester.find_by_name(ui, "inst")
-            text = instance.inspect(SelectedText())
-            self.assertEqual(text, obj.inst_list[0].name)
```
rahulporuri, line 191: im not sure why this was removed

traitsui/tests/editors/test_instance_editor.py
```diff
@@ -146,10 +166,10 @@ def test_custom_editor_with_selection(self):
             # test that the current object is None
             self.assertIsNone(obj.inst)
 
-            # test that the displayed text is correct
+            # test that the displayed text is empty
             instance = tester.find_by_name(ui, "inst")
             text = instance.inspect(SelectedText())
-            self.assertEqual(text, obj.inst_list[0].name)
+            self.assertEqual(text, '')
```
rahulporuri, line 172: i guess this is the right behavior - and previously, we were testing/checking for the buggy/wrong behavior.

rahulporuri (approved):
LGTM with a few questions.

traitsui/qt4/instance_editor.py
aaronayres35: yes, perhaps just doing `choice.setCurrentIndex(-1)` is clearer / better.
I will update

traitsui/tests/editors/test_instance_editor.py
aaronayres35: It was effectively just redundant / already tested elsewhere.  the original issue #1725
was really about the text being correct after changing the name of the selectable object

traitsui/tests/editors/test_instance_editor.py
aaronayres35: Yeah, exactly.  The old test I was more concerned with UITester support than behavior of the editor 😅.  I had basically manually tested a UI and then wrote UITester test that matched what I saw.

rahulporuri:
Still LGTM

PR #1731, opened 2021-07-15
Prevent RangeTextEditor from allowing values outside range

### LINKED ISSUE #737: RangeEditor allows out of range values when mode='text'
In the following example, it is possible to set the value of `x` to values out side of the range specified by the range editor:
```python
from traits.api import *
from traitsui.api import *


class A(HasTraits):

    x = Float()

    traits_view = View(
        Item('x', editor=RangeEditor(mode='text', low=0.5, high=1)),
    )


A().configure_traits()
```

aaronayres35:
closes #737

This PR simply forces the value to be `self.low` if a value outside of the range is given.  This is similar to what the `SimpleSliderEditor` / other range editors do. 

I've tested manually with the example in the issue and added a comparable regression test.

~I still need to investigate wx - hence the WIP title for now~ EDIT: looks like wx was already doing the right thing

**Checklist**

### DIFF
--- a/traitsui/qt4/range_editor.py
+++ b/traitsui/qt4/range_editor.py
@@ -693,6 +693,12 @@ class RangeTextEditor(TextEditor):
     #  Trait definitions:
     # -------------------------------------------------------------------------
 
+    #: Low value for the slider range
+    low = Any()
+
+    #: High value for the slider range
+    high = Any()
+
     #: Function to evaluate floats/ints
     evaluate = Any()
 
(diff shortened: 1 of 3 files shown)


traitsui/qt4/range_editor.py
```diff
@@ -693,6 +693,12 @@ class RangeTextEditor(TextEditor):
     #  Trait definitions:
     # -------------------------------------------------------------------------
 
+    #: Low value for the slider range
+    low = Any()
+
+    #: High value for the slider range
+    high = Any()
```
rahulporuri, line 700: I'm not sure why these are any traits. Is that because users can set `low` to a value or another trait?

rahulporuri (approved):
LGTM with one minor comment.

I tested this locally and the behavior observed in #737 doesn't exist on this branch.

rahulporuri:
I'm going to go ahead and merge this PR @aaronayres35 . Please take another look and see if there was anything else you wanted to do on this PR.

traitsui/qt4/range_editor.py
aaronayres35: TBH II'm not sure either... I was simply following the `SimpleSliderEditor`.  The factory class has them as `Property`, whose setter makes use of this method:
https://github.com/enthought/traitsui/blob/7e1d9ed6e91e2fe4cc356f4966653a2e889909a4/traitsui/editors/range_editor.py#L169-L176

PR #1732, opened 2021-07-17
Simply import "version" information from "traitsui.__init__"

rahulporuri:
Previously, we manually traversed the file system to extract the `version` and `release` information from `traitsui._version` file. Now, we simply expect `traitsui` to be installed in the python environment when we try building the docs and we just import `version` from `traitsui.__init__`. Note that this PR also removes the `release` variable in the configuration file.

Note that these changes should fix the broken RtD docs build. Ref https://readthedocs.org/projects/traitsui/builds/14256791/

**Checklist**

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -12,10 +12,9 @@
 # serve to show the default value.
 
 
-from io import open
 import os
 
-base_path = os.path.dirname(__file__)
+from traitsui import __version__ as version
 
 # If your extensions are in another directory, add it here. If the directory
 # is relative to the documentation root, use os.path.abspath to make it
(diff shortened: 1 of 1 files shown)


docs/source/conf.py
```diff
@@ -50,26 +49,6 @@
...
-# other places throughout the built documents.
-#
-# Pull from the actual release number without imports
-d = {}
-try:
-    version_path = os.path.join(base_path, '..', '..', 'traitsui', '_version.py')
-    with open(version_path, 'r', encoding='utf8') as fp:
-        exec(compile(fp.read(), version_path, 'exec'), d)
-    release = d['version']
-    version = '.'.join(d['version'].split('.', 2)[:2])
```
aaronayres35, line 63: Other ets packages still seem to be following this approach.

aaronayres35 (approved):
The changes LGTM given your PR description.  I am unsure how to test if the readthedocs build gets fixed though.

I built the docs locally and the version is pulled in correctly. eg I see `TraitsUI 7.3.0.dev555 User Manual`. However, previously it looks like we had intentionally cut off the dev stuff with `.split('.', 2)[:2])`.  When we actually publish the docs though using a released version this won't matter. 

Also is this something we will want to do in other ets projects? I notice many of them still follow an approach similar to the old way done here

docs/source/conf.py
rahulporuri: Yeah. I need to fix that. I am not sure why we were following this approach - i'll ask first on the internal slack channel and then gut it from ETS projects.

rahulporuri:
> I am unsure how to test if the readthedocs build gets fixed though.

I was planning on triggering a new RtD build after merging this PR - and i'm going off the fact that these changes exist in traits-futures where there are no RtD docs build errors.

rahulporuri:
this PR did in fact fix the RtD build - https://readthedocs.org/projects/traitsui/builds/14278933/

rahulporuri:
Note that the API documentation is broken because RtD doesn't have the custom build step where api docs are generated. We need to take the approach that traits-futures uses to have a single sphinx build step that builds both api docs and normal docs - https://github.com/enthought/traits-futures/pull/348

PR #1733, opened 2021-07-19
Fix droppable instance editor

### LINKED ISSUE #1478: InstanceEditor with droppable=True crashes on Qt
Setting `droppable=True` on an InstanceEditor when running under Qt crashes with the following error:
```
  File "/Users/cwebster/src/experiments/internal/cwebster/instance_editor/instance_choice_editor.py", line 122, in <module>
    analysis_view.configure_traits(kind='live')
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/handler.py", line 477, in configure_traits
    **args
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traits/has_traits.py", line 2015, in configure_traits
    args,
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/toolkit.py", line 233, in view_application
    context, view, kind, handler, id, scrollable, args
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/view_application.py", line 84, in view_application
    context, view, kind, handler, id, scrollable, args
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/view_application.py", line 126, in __init__
    args=self.args,
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/view.py", line 462, in ui
    ui.ui(parent, kind)
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/ui.py", line 244, in ui
    self.rebuild(self, parent)
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/toolkit.py", line 163, in ui_live
    ui_live.ui_live(ui, parent)
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/ui_live.py", line 43, in ui_live
    _ui_dialog(ui, parent, BaseDialog.NONMODAL)
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/ui_live.py", line 65, in _ui_dialog
    BaseDialog.display_ui(ui, parent, style)
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/ui_base.py", line 286, in display_ui
    ui.owner.init(ui, parent, style)
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/ui_live.py", line 225, in init
    self.add_contents(panel(ui), bbox)
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/ui_panel.py", line 259, in panel
    panel = _GroupPanel(content[0], ui).control
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/ui_panel.py", line 615, in __init__
    layout = self._add_groups(content, inner)
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/ui_panel.py", line 696, in _add_groups
    panel = _GroupPanel(subgroup, self.ui).control
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/ui_panel.py", line 617, in __init__
    layout = self._add_items(content, inner)
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/ui_panel.py", line 891, in _add_items
    editor.prepare(inner)
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/editor.py", line 236, in prepare
    self.init(parent)
  File "/Users/cwebster/.edm/envs/edm/lib/python3.6/site-packages/traitsui/qt4/instance_editor.py", line 127, in init
    self.control.installEventFilter(drop_event_filter)
AttributeError: 'NoneType' object has no attribute 'installEventFilter'
```

The proximate cause seems to be that this code:
https://github.com/enthought/traitsui/blob/fb8ea4a4e7df575a707e64cae8eb882a68ef995c/traitsui/qt4/instance_editor.py#L134-L137
occurs before `self.control` is set, so the fix may be as simple as moving the code further down the method.  It is also possible that the code should read `self._choice`  instead of `self.control`.

aaronayres35:
closes #1478 

**Checklist**

### DIFF
--- a/traitsui/qt4/instance_editor.py
+++ b/traitsui/qt4/instance_editor.py
@@ -122,10 +122,9 @@ class CustomEditor(Editor):
             self._choice.setReadOnly(True)
             self.set_tooltip(self._choice)
 
-        if droppable:
-            # Install EventFilter on control to handle DND events.
-            drop_event_filter = _DropEventFilter(self.control)
-            self.control.installEventFilter(drop_event_filter)
+            self.factory.observe(
+                self.rebuild_items, "values.items", dispatch="ui"
+            )
 
         orientation = OrientationMap[factory.orientation]
         if orientation is None:
@@ -154,6 +153,11 @@ class CustomEditor(Editor):
             layout = QtGui.QBoxLayout(orientation, parent)
             layout.setContentsMargins(0, 0, 0, 0)
             self.create_editor(parent, layout)
+
+        if droppable:
+            # Install EventFilter on control to handle DND events.
+            drop_event_filter = _DropEventFilter(self.control)
+            self.control.installEventFilter(drop_event_filter)
 
         # Synchronize the 'view' to use:
         # fixme: A normal assignment can cause a crash (for unknown reasons) in
@@ -289,7 +293,10 @@ class CustomEditor(Editor):
                 else:
                     choice.setText(name)
             else:
-                choice.setCurrentIndex(-1)
+                # choice can also be a QLineEdit in which case we just leave
+                # text as empty
+                if isinstance(choice, QtGui.QComboBox):
+                    choice.setCurrentIndex(-1)
 
 
     def resynch_editor(self):
--- a/traitsui/qt4/instance_editor.py
+++ b/traitsui/qt4/instance_editor.py
@@ -122,10 +122,6 @@ class CustomEditor(Editor):
             self._choice.setReadOnly(True)
             self.set_tooltip(self._choice)
 
-            self.factory.observe(
-                self.rebuild_items, "values.items", dispatch="ui"
-            )
-
         orientation = OrientationMap[factory.orientation]
         if orientation is None:
             orientation = self.orientation
@@ -372,9 +368,15 @@ class CustomEditor(Editor):
                     dispatch="ui"
                 )
 
-            self.factory.observe(
-                self.rebuild_items, "values.items", remove=True, dispatch="ui"
-            )
+            # _choice can also be a QLineEdit in which case we never set up
+            # this observer.
+            if isinstance(self._choice, QtGui.QComboBox):
+                self.factory.observe(
+                    self.rebuild_items,
+                    "values.items",
+                    remove=True,
+                    dispatch="ui"
+                )
 
         super().dispose()
 
--- a/traitsui/qt4/instance_editor.py
+++ b/traitsui/qt4/instance_editor.py
@@ -360,17 +360,17 @@ class CustomEditor(Editor):
             self._ui.dispose()
 
         if self._choice is not None:
-            if self._object is not None:
-                self._object.observe(
-                    self.rebuild_items,
-                    self._name + ".items",
-                    remove=True,
-                    dispatch="ui"
-                )
-
             # _choice can also be a QLineEdit in which case we never set up
             # this observer.
             if isinstance(self._choice, QtGui.QComboBox):
+                if self._object is not None:
+                    self._object.observe(
+                        self.rebuild_items,
+                        self._name + ".items",
+                        remove=True,
+                        dispatch="ui"
+                    )
+
                 self.factory.observe(
                     self.rebuild_items,
                     "values.items",
--- a/traitsui/tests/editors/test_instance_editor.py
+++ b/traitsui/tests/editors/test_instance_editor.py
@@ -97,6 +97,29 @@ none_view = View(
     Item('change_options'),
     buttons=["OK"],
 )
+non_editable_droppable_view = View(
+    Item(
+        "inst",
+        editor=InstanceEditor(
+            editable=False,
+            droppable=True
+        ),
+        style='custom',
+    ),
+    buttons=["OK"],
+)
+non_editable_droppable_selectable_view = View(
+    Item(
+        "inst",
+        editor=InstanceEditor(
+            name='inst_list',
+            editable=False,
+            droppable=True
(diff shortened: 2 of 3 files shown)


traitsui/tests/editors/test_instance_editor.py
```diff
@@ -389,3 +412,17 @@ def test_none_selected(self):
...
+        obj_with_list = ObjectWithList()
+        tester = UITester()
+
+        with tester.create_ui(obj, {'view': non_editable_droppable_view}) as ui:
+            pass
+
+        with tester.create_ui(
+            obj_with_list, {'view': non_editable_droppable_selectable_view}
+        ) as ui:
+            pass
```
aaronayres35: We previously both of these caused a failure / crash.

traitsui/tests/editors/test_instance_editor.py
```diff
@@ -97,6 +97,29 @@ def _modify_inst_list(self, event):
...
+    Item(
+        "inst",
+        editor=InstanceEditor(
+            editable=False,
+            droppable=True
+        ),
+        style='custom',
+    ),
+    buttons=["OK"],
+)
```
aaronayres35, line 110: I have no idea why one would want to do this... but it is a possible configuration one could make.  This would be when you are "editing" a single instance, but the single instance itself is not editable, so all that is displayed in a QLineEdit displaying the current instance.  The instance can be changed by dragging and dropping a different instance. The user can not select a different instance as theoretically there is no list to select from.

traitsui/qt4/instance_editor.py
```diff
@@ -122,10 +122,9 @@ def init(self, parent):
             self._choice.setReadOnly(True)
             self.set_tooltip(self._choice)
 
-        if droppable:
-            # Install EventFilter on control to handle DND events.
-            drop_event_filter = _DropEventFilter(self.control)
-            self.control.installEventFilter(drop_event_filter)
+            self.factory.observe(
+                self.rebuild_items, "values.items", dispatch="ui"
+            )
```
aaronayres35, line 127: The more I think about this it doesn't really make sense.  This was added because without it we were getting a `NotifierNotFound` error in the `non_editable_droppable_view` case.  However, the only time this occurs is the use case described here: https://github.com/enthought/traitsui/pull/1733/files#r672368160

In which case `values.items` should never change. (There isn't a list of instances, just the one!). 

Thinking about this more I think it probably makes more sense to check for this case when we arer removing the listener (and not try to remove it), rather than just adding it here to avoid the failure.

This code path is rather confusing.  Using `_choice` in this case _doesn't really make sense_.  At least not to me anyway, as we are not really choosing an instance from a list of options...

rahulporuri:
@aaronayres35 is this PR ready for review?

traitsui/qt4/instance_editor.py
```diff
@@ -365,9 +368,15 @@ def dispose(self):
...
-            )
+            # _choice can also be a QLineEdit in which case we never set up
+            # this observer.
+            if isinstance(self._choice, QtGui.QComboBox):
+                self.factory.observe(
+                    self.rebuild_items,
+                    "values.items",
+                    remove=True,
+                    dispatch="ui"
+                )
```
rahulporuri, line 379: Do we also want to remove the other observer hooked up if `factory.name != ""`? Ref https://github.com/enthought/traitsui/blob/6c0e71797a4efa6d74c317c1971e5e61d2944306/traitsui/qt4/instance_editor.py#L109-L112

traitsui/tests/editors/test_instance_editor.py
```diff
@@ -97,6 +97,29 @@ def _modify_inst_list(self, event):
     Item('change_options'),
     buttons=["OK"],
 )
+non_editable_droppable_view = View(
+    Item(
+        "inst",
+        editor=InstanceEditor(
+            editable=False,
+            droppable=True
```
rahulporuri, line 105:
```suggestion
            droppable=True,
```

traitsui/tests/editors/test_instance_editor.py
```diff
@@ -97,6 +97,29 @@ def _modify_inst_list(self, event):
...
+    ),
+    buttons=["OK"],
+)
+non_editable_droppable_selectable_view = View(
+    Item(
+        "inst",
+        editor=InstanceEditor(
+            name='inst_list',
+            editable=False,
+            droppable=True
```
rahulporuri, line 117:
```suggestion
            droppable=True,
```

rahulporuri (approved):
LGTM with a couple of nitpicky suggestions and one question about another observer that we might want to unhook.

traitsui/qt4/instance_editor.py
aaronayres35: I believe this is already getting unhooked above

aaronayres35:
@rahulporuri does the latest change with unhooking the observer still look okay?

rahulporuri:
still LGTM but this PR doesn't have a news fragment

PR #1734, opened 2021-07-20
Fix double error dialogs

### LINKED ISSUE #1550: Editor error dialog yields 2 dialogs / subsequent segfault on MacOS and Pyqt5
see PR #1546 for original inspiration.

To reproduce run:
```

class Foo(HasTraits):
    x = Range(low=0.0, high=1.0, value=0.5, exclude_low=True)

Foo().configure_traits()
```
and drag the slider all the way to `0.0`.  Result is:
111837968-a77fba00-88c6-11eb-8f17-284309402c97
and closing the dialogs triggers a segfault

aaronayres35:
closes #1550 

~WIP because I still need to add regression tests~ EDIT: tests have been added

**Checklist**


traitsui/tests/test_editor.py
```diff
@@ -937,10 +937,6 @@ def test_sync_value_both(self):
     # regression test for enthought/traitsui#1543
     @requires_toolkit([ToolkitName.qt])
     @unittest.skipIf(no_modal_dialog_tester, "ModalDialogTester unavailable")
-    @unittest.skipIf(
-        is_mac_os,
-        "There is a separate issue on OSX. See enthought/traitsui#1550"
-    )
     def test_editor_error_msg(self):
```
aaronayres35, line 940: This test is _very_ similar to the newly added range editor test `test_modify_out_of_range`.  However, this test was motivated by the text format of the error dialog.  The other is about the range editor not triggering two error dialogs.  I think it make sense to leave them both...

traitsui/qt4/range_editor.py
```diff
@@ -739,7 +749,7 @@ def update_object(self):
             else:
                 col = OKColor
 
-            self.value = value  
+            self.value = value
```
aaronayres35, line 752: I didn't remember touching this line, but remembered I recently updated VSCode settings to kill trailing whitespace whenever I save a file 😄

traitsui/qt4/range_editor.py
```diff
@@ -208,8 +215,12 @@ def update_object_on_enter(self):
                 )
             finally:
                 self.control.slider.blockSignals(blocked)
-        except TraitError as excp:
-            pass
+        except TraitError:
+            # They entered something invalid, pretend it didn't happen
+            value = self.value
+            self.control.text.setText(str(value))
```
aaronayres35, line 221: Without this my test was failing because we would set text to 0.0, hit enter, error message pops up, we click ok on the error dialog.  The slider value would be unchanged (eg still show as 0.5 because the error occurred in the `self.value = value` line above so we never got to the `self.control.slider.setValue` line.  However, the textbox would still contain what the user had been typing just before hitting enter.  In other words it would still show 0.0.  Then, when the main ui was closed (we exit the `create_ui` context manager), it was as if we were trying to enter that value in the text field again.  This led to _another_ error dialog causing the test to hang.

I think reseting the text box to the previously set _valid_ value before you hit enter makes sense from a UX perspective.  Perhaps a reviewer will disagree though?  If so, I can try to modify the code / test to work with that UX.  If it doesn't get reset then it is expected that the user will modify the value again to an acceptable value.

corranwebster: Where resetting the value can be problematic is when you are transitioning through an invalid value to enter a valid value, eg. if the minimum was `0.05` and you started typing `0.0` on the way to that, having a reset makes it hard to hit the desired value.

However if this is only ever triggered by ending editing, resetting is OK, as by definition once editing is done the user can't be transitioning to another value.

corranwebster (approved):
This looks good to me.

I am not fond of any of the errors states which pop up a modal dialog - they are very interruptive of the user's interactions and can likely be handled better by other signals (eg. bg color, error icons, etc.). But given the current state this fixes the issue.

PR #1735, opened 2021-07-20
Update to GitHub Actions CI configurations

rahulporuri:
This PR does two things

- sets `PYTHONUNBUFFERED` environment variable to `True` on all jobs except for the PyPI upload job.
- sets up `concurrency` so that existing CI runs are cancelled when a new commit is pushed to a PR branch. This is done only for the jobs that are run on PRs.

One more step towards enthought/ets#71

**Checklist**

### DIFF
--- a/.github/workflows/bleeding-edge.yml
+++ b/.github/workflows/bleeding-edge.yml
@@ -8,6 +8,9 @@ on:
   # Make it possible to manually trigger the workflow
   workflow_dispatch:
 
+env:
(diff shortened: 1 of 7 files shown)


rahulporuri:
it looks like this is the wrong solution as the style check and test jobs were cancelled - https://github.com/enthought/traits/actions/runs/1049390008 and https://github.com/enthought/traitsui/actions/runs/1049408598

PR #1737, opened 2021-07-26
Replace all use of magic named _fired methods with observe

aaronayres35:
This PR simply replaces all uses of the `_fired` magically named methods with their observe equivalent.

**Checklist**

### DIFF
--- a/traitsui/qt4/table_editor.py
+++ b/traitsui/qt4/table_editor.py
@@ -1374,7 +1374,7 @@ class TableFilterEditor(HasTraits):
         ]
         self.selected_template = self.templates[0]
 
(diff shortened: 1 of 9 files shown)


traitsui/qt4/table_editor.py
```diff
@@ -1374,7 +1374,8 @@ def _editor_changed(self):
         ]
         self.selected_template = self.templates[0]
 
-    def _add_button_fired(self):
+    @observe('_add_button')
```
rahulporuri, line 1377:
```suggestion
    @observe('add_button')
```

rahulporuri (approved):
Almost LGTM

PR #1741, opened 2021-08-09
Default branch has been renamed from "master" to "main"

rahulporuri:
This PR replaces the use of `master` in the repo with `main` as the default branch name has been updated.

**Checklist**

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -175,7 +175,7 @@ except ImportError as exc:
 
 # Useful aliases to avoid repeating long URLs.
 extlinks = {'github-demo': (
-    'https://github.com/enthought/traitsui/tree/master/traitsui/examples/demo/%s',
+    'https://github.com/enthought/traitsui/tree/main/traitsui/examples/demo/%s',
     'github-demo')
 }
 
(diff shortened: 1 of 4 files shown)


docs/source/conf.py
```diff
@@ -175,7 +175,7 @@
 
 # Useful aliases to avoid repeating long URLs.
 extlinks = {'github-demo': (
-    'https://github.com/enthought/traitsui/tree/master/traitsui/examples/demo/%s',
+    'https://github.com/enthought/traitsui/tree/main/traitsui/examples/demo/%s',
```
rahulporuri, line 178: I'll fix this in a separate PR - we shouldn't be linking to the example of `main` branch. Instead, I think we should be linking to the specific version/tag.

mdickinson (approved):
LGTM

rahulporuri:
Update - i was afraid that the links to demos in the docs were broken because of the rename - but github seems to have taken care of that for us! All of the links in https://docs.enthought.com/traitsui/demos/index.html continue to work as expected and we get a banner on top of the github page stating that `master` was renamed to `main`.

PR #1757, opened 2021-09-22
Fix/flake8 code issues

corranwebster:
Low-hanging fixes for various flake8 issues.  These are the more serious and/or less frequent thinks that need fixing.

This is work towards #1754.

Reviewing may be easier commit-by-commit, as each commit deals with one particular error code.

Edit:

Fixing the comparisons to True/False exposed an issue where many custom `Handler` subclasses were not returning anything from their `init()` methods (the API calls for returning True/False, with False raising an error in the `UI` class).  As a result we are checking for `handler.init(...) is False` rather than `not handler.init(...)` and have added a warning for 3rd party users if the result is not a `bool` (since many of the missing return values were in example code, raising an error seems problematic even if the fix is easy).  Additionally all handlers in TraitsUI were fixed to conform to the API.

There is a question if the API is good (should `handler.init()` just raise directly?), and further work is needed in future releases, so a separate issue will be opened for this.

**Checklist**

### DIFF
--- a/integrationtests/ui/instance_editor_test2.py
+++ b/integrationtests/ui/instance_editor_test2.py
@@ -26,7 +26,7 @@ class Person(HasStrictTraits):
 
     name = Str()
     age = Int()
-    phone = Regex(value='000-0000', regex='\d\d\d[-]\d\d\d\d')
+    phone = Regex(value='000-0000', regex=r'\d\d\d[-]\d\d\d\d')
 
     #-------------------------------------------------------------------------
     #  Traits view definition:
--- a/integrationtests/ui/instance_editor_test2.py
+++ b/integrationtests/ui/instance_editor_test2.py
@@ -9,8 +9,7 @@
 # Thanks for using Enthought open source!
 
 from traits.api import HasStrictTraits, Instance, Int, List, Regex, Str
-from traitsui.api import InstanceEditor, Item, View
-from traitsui.instance_choice import InstanceFactoryChoice
+from traitsui.api import InstanceEditor, InstanceFactoryChoice, Item, View
 
 #-------------------------------------------------------------------------
 #  'Person' class:
--- a/integrationtests/ui/instance_editor_test2.py
+++ b/integrationtests/ui/instance_editor_test2.py
@@ -8,10 +8,9 @@
 #
 # Thanks for using Enthought open source!
 
-from traits.api import *
-from traitsui.api import *
-from traitsui.instance_choice \
-    import InstanceChoice, InstanceFactoryChoice
+from traits.api import HasStrictTraits, Instance, Int, List, Regex, Str
+from traitsui.api import InstanceEditor, Item, View
+from traitsui.instance_choice import InstanceFactoryChoice
 
 #-------------------------------------------------------------------------
 #  'Person' class:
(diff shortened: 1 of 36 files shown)


integrationtests/ui/instance_editor_test2.py
```diff
@@ -8,10 +8,9 @@
 #
 # Thanks for using Enthought open source!
 
-from traits.api import *
-from traitsui.api import *
-from traitsui.instance_choice \
-    import InstanceChoice, InstanceFactoryChoice
+from traits.api import HasStrictTraits, Instance, Int, List, Regex, Str
+from traitsui.api import InstanceEditor, Item, View
+from traitsui.instance_choice import InstanceFactoryChoice
```
rahulporuri, line 13:
```suggestion
from traitsui.api import InstanceEditor, InstanceFactoryChoice, Item, View
```

integrationtests/ui/instance_editor_test5.py
```diff
@@ -8,10 +8,9 @@
 #
 # Thanks for using Enthought open source!
 
-from traits.api import *
-from traitsui.api import *
-from traitsui.instance_choice \
-    import InstanceChoice, InstanceFactoryChoice
+from traits.api import HasStrictTraits, Instance, Int, List, Regex, Str
+from traitsui.api import InstanceEditor, Item, View
+from traitsui.instance_choice import InstanceFactoryChoice
```
rahulporuri, line 13:
```suggestion
from traitsui.api import InstanceEditor, InstanceFactoryChoice, Item, View
```

integrationtests/ui/instance_editor_test6.py
```diff
@@ -8,10 +8,9 @@
 #
 # Thanks for using Enthought open source!
 
-from traits.api import *
-from traitsui.api import *
-from traitsui.instance_choice \
-    import InstanceChoice, InstanceFactoryChoice
+from traits.api import HasStrictTraits, Instance, Int, List, Regex, Str
+from traitsui.api import InstanceEditor, Item, View
+from traitsui.instance_choice import InstanceFactoryChoice
```
rahulporuri, line 13:
```suggestion
from traitsui.api import InstanceEditor, InstanceFactoryChoice, Item, View
```

traitsui/ui.py
```diff
@@ -407,8 +407,16 @@ def prepare_ui(self):
...
         # failure:
-        if handler.init(info) == False:
+        started = handler.init(info)
+        if started is False:
             raise TraitError("User interface creation aborted")
+        elif not isinstance(started, bool):
+            warn(
+                "Handler.init() must return True or False, but instead "
+                f"returned {started}.  "
+                "This will become an error in a future release.",
```
rahulporuri, line 417: This works but it feels off. Without the line length rule, this would be an f-string on a single-line. Not an actionable comment. I think i'd just prefer if it all of them were f-strings. But that might be confusing.

rahulporuri (approved):
Mostly LGTM apart from a few comments.

integrationtests/ui/instance_editor_test2.py
corranwebster: Right, I forgot we'd made them part of the API!

traitsui/ui.py
corranwebster: Yes, I wasn't sure if it would work like this, or if they all needed to be f-strings, or if an f-string with no substitutions was even allowed.  I tried this and it worked.

If you like I could include explicit `+` operations which might make it less magical.

rahulporuri: no. it's fine. like i said, not an actionable comment.

PR #1762, opened 2021-10-06
Fix datetime initialization problem

### LINKED ISSUE #1761: Fix DatetimeEditor initialization when minimum and maximum dates are provided
Currently in traitsui there is a bug where the listener on the DatetimeEditor will update the value of the editor when initializing the `minimum_datetime` and `maximum_datetime` values.  In short, if the either are set (`minimum_datetime` or `maximum_datetime`) then the initial value will be lost and will be set to the `minimum_datetime` or `maximum_datetime` in that order.

Minimum working example of bug:
```python
from datetime import datetime, timedelta

from traits.api import Datetime, HasTraits
from traitsui.api import DatetimeEditor, Item, View

MAX_DATE = datetime.now()
MIN_DATE = MAX_DATE - timedelta(days=10)


class DatetimeTest(HasTraits):

    start = Datetime()
    stop = Datetime()

    def _start_default(self):
        return MIN_DATE

    def _stop_default(self):
        return MAX_DATE

        return View(
            Item('start',
                 editor=DatetimeEditor(
                     maximum_datetime=MAX_DATE,
                     minimum_datetime=MIN_DATE)),
            Item('stop',
                 editor=DatetimeEditor(
                     maximum_datetime=MAX_DATE,
                     minimum_datetime=MIN_DATE)),
        )


if __name__ == "__main__":
    test = DatetimeTest()
    test.configure_traits()
```

ctangell:
Closes #1761. Fixes the data initialization problem by moving the code the connects the update of the widget to after when the minimum and maximum date values are initialized.

### DIFF
--- a/traitsui/qt4/datetime_editor.py
+++ b/traitsui/qt4/datetime_editor.py
@@ -38,9 +38,9 @@ class SimpleEditor(Editor):
         self.maximum_datetime = self.factory.maximum_datetime
 
         self.control = QtGui.QDateTimeEdit()
-        self.control.dateTimeChanged.connect(self.update_object)
         self.update_minimum_datetime()
         self.update_maximum_datetime()
+        self.control.dateTimeChanged.connect(self.update_object)
 
     def dispose(self):
         """Disposes of the contents of an editor."""


ctangell:
Added a regression test.  I didn't see the same problem in the wx editor, so this test only runs for qt.

PR #1765, opened 2021-10-13
Allow read-only TextEditor text to be selected by users by default

Corwinpro:
Changes the default behaviour of `TextEditor` (only for Qt backend) in read-only mode: the displayed text can now be selected in the UI by default (the `readonly_allow_selection` attribute defaults to `True`).

The content of read-only `TextEditor`s is usually some useful information so users need to be able to copy those values quite often.

### DIFF
--- a/traitsui/editors/text_editor.py
+++ b/traitsui/editors/text_editor.py
@@ -78,7 +78,7 @@ class TextEditor(EditorFactory):
     view = AView
 
     #: In a read-only text editor, allow selection and copying of the text.
-    readonly_allow_selection = Bool(False)
+    readonly_allow_selection = Bool(True)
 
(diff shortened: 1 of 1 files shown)


corranwebster (approved):
This only affects `readonly`-style `TextEditor`s. An addition is perhaps to set the underlying `QtCore.Qt.TextSelectableByMouse` flag in the base `ReadonlyEditor` and then un-set it in the `TextEditor` if `readonly_allow_selection` is `False`.  This is othogonal to changing the default however, so it can be a separate PR.

Corwinpro:
@corranwebster I am not sure why it is failing: I can only see "This check failed" message in the Details page.

Corwinpro:
Only Windows is failing with no obvious reason.

PR #1769, opened 2021-11-01
Call Bool before evaluating {enabled/visible}_when condition for labels

### LINKED ISSUE #1768: visible_when on Label doesn't implicitly take the boolean of the specified trait
Consider this code:
```python
from traits.api import HasTraits, Str
from traitsui.api import View, Item, Label, UItem

class Foo(HasTraits):

    a_string = Str()
    fixed_string = Str("a_string is true (according to UItem)")

    traits_view = View(
        Item("a_string"),
        Label("a_string is true (according to Label)", visible_when="a_string"),
        UItem("fixed_string", style="readonly", visible_when="a_string"),
        Label("len(a_string) is nonzero (according to Label)", visible_when="len(a_string)"),
        width=420, height=200
    )

Foo().configure_traits()
```

When `a_string` is the empty string (therefore False): 
- the `Label(visible_when="a_string")` does not get hidden
- for comparison, the `UItem` with the same `visible_when`, and the `Label(visible_when="len(a_string)")`, correctly do get hidden

Peek 2021-11-01 11-13

Tried on `main` with `pyside2`.

aaronayres35:
closes #1768 

WIP label because the still needs a test.  support was introduced in #1544, but that needs a test as well.  Not sure if I will have the time to write a test on this, but wanted to open a PR with the simple fix now so it isn't forgetten. (previous comment about difficulty I had writing tests for Labels {enabled/visible}_when https://github.com/enthought/traitsui/pull/1544#issuecomment-823551479)

Note that for "normal" enabled/visible when we have `cond_value` come up in an `if` statement: https://github.com/enthought/traitsui/blob/87dceb219b8b7b547907f031aaca70e1a66b7945/traitsui/ui.py#L793

In the code here we really do want to always pass a bool to `setVisible` or `setEnabled`.

**Checklist**

### DIFF
--- a/traitsui/qt4/ui_panel.py
+++ b/traitsui/qt4/ui_panel.py
@@ -726,7 +726,7 @@ class _GroupPanel(object):
             method_to_call = getattr(label, method_dict[kind])
             try:
                 cond_value = eval(when, globals(), context)
-                method_to_call(cond_value)
+                method_to_call(bool(cond_value))
             except Exception:
                 # catch errors in the validate_when expression
                 from traitsui.api import raise_to_debug
--- a/traitsui/wx/ui_panel.py
+++ b/traitsui/wx/ui_panel.py
@@ -732,7 +732,7 @@ class FillPanel(object):
             method_to_call = getattr(label, method_dict[kind])
             try:
                 cond_value = eval(when, globals(), context)
-                method_to_call(cond_value)
+                method_to_call(bool(cond_value))
             except Exception:
                 # catch errors in the validate_when expression
                 from traitsui.api import raise_to_debug


aaronayres35:
@corranwebster this PR was labelled as a WIP because it does not contain a test but it is a trivial change that fixes the associated issue.  I had trouble writing a test for Label {enabled/visible}_when previously (https://github.com/enthought/traitsui/pull/1544#issuecomment-823551479) and am not sure I'll have the time to figure that out now.  Wanted to get your thoughts on if you think it is safe / worth it to throw this fix into the next release or if we should just hold off.

corranwebster (approved):
LGTM

PR #1791, opened 2021-12-23
Fix NotebookEditor initial selected

### LINKED ISSUE #1790: NotebookEditor initial selected
If you have a notebook style list editor, by default, the first tab is shown / selected, however the `selected` trait remains `None` initially.

This happens because `NotebookEditor.init` is called, which connects `currentChanged` qt signal to the method `_tab_activated` which sets the `selected` trait by looking through `_uis`.  Then, `update_editor` gets called, which goes through each item in the list (ie `self.value`), calls `_create_page` and then adds that page to `_uis`.  _However_, that first call to `_create_page` will hit a `self.control.addTab` call which, since before that the QTabWidget was empty, will cause `currentChanged` to fire and `_tab_activated` to be called but this happens _before_ we add this page to `_uis` so `_tab_activated` is effectively a no-op when it shouldn't be.

aaronayres35:
closes #1790 

**Checklist**

### DIFF
--- a/traitsui/qt4/list_editor.py
+++ b/traitsui/qt4/list_editor.py
@@ -526,8 +526,7 @@ class NotebookEditor(Editor):
 
             # Remember the page for later deletion processing:
             self._uis.append([ui.control, ui, view_object, monitoring])
-            if len(self._uis) == 1:
-                self._tab_activated(0)
+        self._tab_activated(0)
         if self.selected:
             self._selected_changed(self.selected)
 
--- a/traitsui/qt4/list_editor.py
+++ b/traitsui/qt4/list_editor.py
@@ -526,7 +526,8 @@ class NotebookEditor(Editor):
 
             # Remember the page for later deletion processing:
             self._uis.append([ui.control, ui, view_object, monitoring])
-
+            if len(self._uis) == 1:
+                self._tab_activated(0)
         if self.selected:
             self._selected_changed(self.selected)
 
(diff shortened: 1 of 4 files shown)


traitsui/qt4/list_editor.py
```diff
@@ -526,7 +526,8 @@ def update_editor(self):
 
             # Remember the page for later deletion processing:
             self._uis.append([ui.control, ui, view_object, monitoring])
-
+            if len(self._uis) == 1:
+                self._tab_activated(0)
```
aaronayres35, line 530: These fixes are a little clunky, but they get the job done.  Maybe it is worth thinking about it more and coming up with something cleaner?  

Goal here was really a quick fix

rahulporuri (approved):
The code changes LGTM but i'd recommend waiting for @siddhantwahal s approval as well.

traitsui/qt4/list_editor.py
```diff
@@ -526,7 +526,8 @@ def update_editor(self):
...
         if self.selected:
```
siddhantwahal, line 531: Can `self._tab_activated` be placed outside the for loop? That makes it clearer that we always want the first tab to be selected. 

The tests seem do seem to pass if it's outside.

```suggestion
        self._tab_activated(0)
        if self.selected:
```

siddhantwahal (approved):
Looks good to me to, one question though:

traitsui/qt4/list_editor.py
aaronayres35: I think this makes sense / is safe 👍

PR #1797, opened 2022-02-01
Convert Qt Enums to PyQt6 compatible forms

### LINKED ISSUE #1798: Add Enum class names to Enums for Qt
PyQt6 is using Python Enums for Qt Enum types, which means that you need to insert the Enum class name when referring to the Enum. See https://stackoverflow.com/questions/66235661/how-to-check-mousebuttonpress-event-in-pyqt6 for example.

So this means something like QEvent.Close becomes QEvent.Type.Close. This is a change with huge impacts in many places , but the construct is compatible with other Qt versions that we care about (PySide2, PySide6, PyQt5).

See https://github.com/enthought/pyface/issues/1092 for the Pyface version

corranwebster:
This should still work with all other Qt backends.  Fixes #1798.

One manual change was needed for the VideoEditor code. There is a change of enum names between Qt5 and Qt6 for QMediaPlayer, and I fixed to use the Qt5 names.  We will sort out how to handle the different names when we properly add support for Qt6 (which will likely require a number of other similar changes). 

**Checklist**


traitsui/qt4/table_editor.py
```diff
@@ -1246,7 +1246,7 @@ def _update_header_sizing(self):
             hheader.setStretchLastSection(True)
         else:
             # hheader.setSectionResizeMode(
-            #     stretchable_columns[-1], QtGui.QHeaderView.Stretch)
+            #     stretchable_columns[-1], QtGui.QHeaderView.ResizeMode.Stretch)
```
aaronayres35, line 1249: just pointing out the commented code

traitsui/qt4/tabular_editor.py
```diff
@@ -670,9 +670,9 @@ def paint(self, painter, option, index):
 
         # FIXME: 'styleHint' is returning bogus (negative) values. Why?
         # style = QtGui.QApplication.instance().style()
-        # color = style.styleHint(QtGui.QStyle.SH_Table_GridLineColor, option)
+        # color = style.styleHint(QtGui.QStyle.StyleHint.SH_Table_GridLineColor, option)
         # painter.setPen(QtGui.QColor(color))
```
aaronayres35, line 674: same here

traitsui/qt4/video_editor.py
```diff
@@ -9,6 +9,7 @@
 # Thanks for using Enthought open source!
 
 """Traits UI 'display only' video editor."""
+from pyface.qt import is_qt5
```
aaronayres35, line 12: Doesn't seem like this is being used yet?

aaronayres35 (approved):
Changes LGTM

Do you have a reference for how you went about making the automated changes? A combo of PyQtEnumConverter and re.sub ?

corranwebster:
> Do you have a reference for how you went about making the automated changes? A combo of PyQtEnumConverter and re.sub ?

Yes, the same script that I used for Pyface.

traitsui/qt4/table_editor.py
corranwebster: OK with changes to commented out code, since that code may eventually be not commented out.

PR #1804, opened 2022-02-02
Fix SimpleSpinEditor "enter set" on qt

aaronayres35:
This PR adds a simple fix for providing enter set behavior for the simple spin editor.  As with some other editors, this behavior is enabled by setting `auto_set` to False (the value of `enter_set` doesn't actually matter).

ref: https://doc.qt.io/archives/qt-4.8/qabstractspinbox.html#keyboardTracking-prop

**Checklist**


corranwebster (approved):
LGTM

PR #1807, opened 2022-02-08
Add a dynamically created ArrayImage to the ImageEditor example.

corranwebster:
Illustrates the new capabilities from Pyface 7.4.0.  Modified UI looks like this:

image

The gradient comes from an RGB array built dynamically in the example.

**Checklist**


traitsui/examples/demo/Extras/Image_editor_demo.py
```diff
@@ -62,18 +65,34 @@ class Employee(HasTraits):
                     ),
```
aaronayres35, line 65: I was surprised that this worked previously without defining the picture trait on the class - just passing `picture=ImageResource('e-logo-rev', search_path=search_path)` on instantiation.

aaronayres35 (approved):
LGTM, I ran the example locally and it worked as expected

PR #1809, opened 2022-02-09
Use two-phase HeadingText creation from Pyface 7.4.0.

corranwebster:
This removes a deprecation warning coming from a change in Pyface behaviour. Additionally it adds proper disposal of the HeadingText Widget to the TitleEditor.  It would be nice to do the same for the ui_panel code, but there is no available hook and deeper architectural issues with the code.

**Checklist**
- [ ] ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~ (Not needed - small internal change).

### DIFF
--- a/traitsui/qt4/ui_panel.py
+++ b/traitsui/qt4/ui_panel.py
@@ -1349,7 +1349,7 @@ class HTMLHelpWindow(QtGui.QDialog):
 HeadingText = None
 
 
-def heading_text(*args, create=False,  **kw):
+def heading_text(*args, create=False, **kw):
     """Create a Pyface HeadingText control."""
     global HeadingText
 
--- a/traitsui/qt4/ui_panel.py
+++ b/traitsui/qt4/ui_panel.py
@@ -168,7 +168,7 @@ class _Panel(BasePanel):
 
             # Handle any view title.
             if title != "":
-                layout.addWidget(heading_text(None, text=view.title).control)
+                layout.addWidget(heading_text(parent=w, text=view.title).control)
 
             if isinstance(self.control, QtGui.QWidget):
                 layout.addWidget(self.control)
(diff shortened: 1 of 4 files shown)


traitsui/qt4/ui_panel.py
```diff
@@ -1349,11 +1349,13 @@ def __init__(self, parent, html_content, scale_dx, scale_dy):
 HeadingText = None
 
 
-def heading_text(*args, **kw):
+def heading_text(*args, create=False, **kw):
```
rahulporuri, line 1352: should the new `create` kwarg be passed along to the `widget.create` method call?

traitsui/wx/ui_panel.py
```diff
@@ -1219,11 +1219,13 @@ def _on_ok(self, event):
 HeadingText = None
 
 
-def heading_text(*args, **kw):
-    """Creates a Pyface HeadingText control."""
+def heading_text(*args, create=False, **kw):
```
rahulporuri, line 1222: same comment here as with the qt `ui_panel` code

traitsui/qt4/ui_panel.py
corranwebster: Indeed.  Thank you!

traitsui/qt4/ui_panel.py
```diff
@@ -1349,11 +1349,13 @@ def __init__(self, parent, html_content, scale_dx, scale_dy):
...
     """Create a Pyface HeadingText control."""
     global HeadingText
 
     if HeadingText is None:
         from pyface.api import HeadingText
 
-    return HeadingText(*args, **kw)
+    widget = HeadingText(create=create, *args, **kw)
```
rahulporuri, line 1359: I think we need to make this change as well -

```suggestion
    widget = HeadingText(*args, create=create, **kw)
```

traitsui/wx/ui_panel.py
```diff
@@ -1219,11 +1219,13 @@ def _on_ok(self, event):
...
+    """Create a Pyface HeadingText control."""
     global HeadingText
 
     if HeadingText is None:
-        from pyface.heading_text import HeadingText
+        from pyface.api import HeadingText
 
-    return HeadingText(*args, **kw)
+    widget = HeadingText(create=create, *args, **kw)
```
rahulporuri, line 1229:
```suggestion
    widget = HeadingText(*args, create=create, **kw)
```

PR #1819, opened 2022-02-16
Improvements for Font traits

corranwebster:
This adds support for Pyface Font traits, as well as tests.

This does some fussing around with the way font families are handled - they seem to have little effect on qt4 and OS X, while on wx they are very powerful and may override the user's preferred font.  I've adjusted the way things work slightly such that I hope it makes a little better sense, in particular to make sure that the font family gets included in the string version of the code (since that is used for pickling, it seemed important!) - the tests show the behaviour.

This code is likely to be deprecated/replaced in the next release, so the tests are probably the most important piece as they document current behaviours.

**Checklist**

### DIFF
--- a/traitsui/qt4/font_trait.py
+++ b/traitsui/qt4/font_trait.py
@@ -25,8 +25,9 @@
 
 
 from pyface.qt import QtGui
-
+from pyface.font import Font as PyfaceFont
 from traits.api import Trait, TraitHandler, TraitError
+
 
 
 # -------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


traitsui/qt4/font_trait.py
```diff
@@ -92,9 +106,7 @@ def create_traitsfont(value):
     for word in value.split():
         lword = word.lower()
         if lword in font_families:
-            f = QtGui.QFont()
-            f.setStyleHint(font_families[lword])
-            family = f.defaultFamily()
```
corranwebster, line 97: This was changed because it always returned an empty string on OS X.

aaronayres35 (approved):
LGTM just a few nit picky comments

traitsui/wx/font_trait.py
```diff
@@ -62,8 +77,10 @@ def font_to_str(font):
 
 def create_traitsfont(value):
     """Create a TraitFont object from a string description."""
```
corranwebster, line 79:
```suggestion
    """Create a TraitsFont object from a string description."""
```

PR #1820, opened 2022-02-16
Fix and refactor KeyBinding editor.

corranwebster:
This refactor adds a `ModelView` handler for the editor and separates out some of the `KeyBindings` functionality into there.  It also removes the focus tracking, as that turned out to be non-functional (it was never called from anywhere).  This allowed significant simplification of the code.

**Checklist**


traitsui/key_bindings.py
```diff
@@ -294,3 +260,48 @@ def _match_binding(self, binding, skip=frozenset()):
...
+        return None
+
+    def _match_binding(self, binding, skip=frozenset()):
+        """Return all KeyBinding instances that match the given binding.
+        """
+        return (
+            a_binding
+            for a_binding in self.bindings
+            if a_binding not in skip and a_binding.match(binding)
+        )
```
aaronayres35, line 284: do we need the duplication here? or can we simply call the method on the model of the same name?

traitsui/qt4/key_binding_editor.py
```diff
@@ -57,47 +58,34 @@ def init(self, parent):
...
-                != QtGui.QMessageBox.StandardButton.Yes
-            ):
+                ),
+                title="Duplicate Key Definition",
+            )
+            if result != YES:
                 return
 
         self.value = key_name
+        # Need to manually update editor because loopback protection doesn't
```
aaronayres35, line 86: what does loopback protection mean here?

traitsui/qt4/key_binding_editor.py
```diff
@@ -57,47 +58,34 @@ def init(self, parent):
...
         self.control = KeyBindingCtrl(self)
 
-    def update_object(self, event):
-        """Handles the user entering input data in the edit control."""
-        try:
-            self.value = value = key_event_to_name(event)
-            self._binding.text = value
-        except:
-            pass
-
```
aaronayres35, line 67: This is being removed because it is unused correct?

traitsui/key_bindings.py
```diff
@@ -294,3 +260,48 @@ def _match_binding(self, binding, skip=frozenset()):
...
+    traits_view = View(
+        [
+            Item(
+                "bindings",
+                style="readonly",
+                show_label=False,
+                editor=ListEditor(style="custom"),
+            ),
+            "|{Click on an entry field, then press the key to "
+            "assign. Double-click a field to clear it.}<>",
```
aaronayres35, line 300: double clicking to clear doesn't seem to work for me, but I may be doing something wrong

aaronayres35 (approved):
LGTM just a few comments

traitsui/qt4/key_binding_editor.py
```diff
@@ -27,6 +27,7 @@
 
 
 from pyface.qt import QtCore, QtGui
+from pyface.api import YES, confirm
 
 from traits.api import Bool, Event
```
rahulporuri, line 32: nit - grouping the ETS imports together.

```suggestion
from pyface.qt import QtCore, QtGui

from pyface.api import YES, confirm
from traits.api import Bool, Event
```

traitsui/key_bindings.py
```diff
@@ -294,3 +260,48 @@ def _match_binding(self, binding, skip=frozenset()):
...
+        """Return all KeyBinding instances that match the given binding.
+        """
+        return (
+            a_binding
+            for a_binding in self.bindings
+            if a_binding not in skip and a_binding.match(binding)
+        )
+
+    def _bindings_default(self):
+        bindings = list(set(self.model.root._get_bindings([])))
```
rahulporuri, line 287: This API feels a little awkward. Why make `bindings` a necessary argument for `_get_bindings` in the first place? why not make `bindings=None` the default and make `_get_bindings` handle the case internally?

traitsui/qt4/key_binding_editor.py
rahulporuri: I think it means trait change listeners are disabled when updating the `Editor.value` trait - IIRC it's done to prevent recursion of the listeners or something of that sort.

traitsui/tests/test_key_bindings.py
```diff
@@ -269,3 +225,54 @@ def test_do_no_match_complete(self):
 
         self.assertFalse(result)
         self.assertIsNone(controller.called)
+
+
+class TestKeyBindingsHandler(UnittestTools, unittest.TestCase):
```
rahulporuri, line 230: it doesn't look like we need to inherit from `UnittestTools` here.

traitsui/key_bindings.py
corranwebster: It's needed because the `bindings` list on this object is sorted differently from the one on the underlying model.

traitsui/qt4/key_binding_editor.py
corranwebster: Yes - it is never called.

traitsui/qt4/key_binding_editor.py
corranwebster: I think my comment here is a bit garbled.  I'll update.

As Rahul mentioned, you frequently run into trouble in GUI code with infinite event loops where change in UI -> change in model -> change in UI -> ... etc.  So TraitsUI has code to prevent this sort of loopback.  When you set the `value`, it assumes that it is coming from the toolkit, and so the toolkit objects are in the desired state, and so it generally blocks pushing back changes to the toolkit object from this change.

However in this case, the content of the toolkit object isn't matching the value, so we need to manually update that as well.

This isn't needed on wx, since it overrides drawing completely and gets the context from the editor value.

traitsui/qt4/key_binding_editor.py
```diff
@@ -57,47 +58,34 @@ def init(self, parent):
...
-            ):
+                ),
+                title="Duplicate Key Definition",
+            )
+            if result != YES:
                 return
 
         self.value = key_name
+        # Need to manually update editor because loopback protection doesn't
+        # catch the change.
```
corranwebster, line 87:
```suggestion
        # Need to manually update editor because update to value won't
        # get transferred to toolkit object due to loopback protection.
```

traitsui/key_bindings.py
corranwebster: Yep, not fond of it myself (it feels old-school), but it works and I wanted to change as little as possible, given the state of testing on the code.

traitsui/tests/test_key_bindings.py
corranwebster, line 230:
```suggestion
class TestKeyBindingsHandler(unittest.TestCase):
```

traitsui/key_bindings.py
corranwebster: Looks like it doesn on Wx, but was never hooked up in Qt.  I'll open an issue.

corranwebster: See #1854.

PR #1824, opened 2022-02-17
Fix issue with resetting the `ProgressEditor`

### LINKED ISSUE #1787: ProgressDialog does not support reset to 0
When using the ProgressEditor with a range from 0 to 100, one cannot reset the progress back to 0 after changing it to a value > 0, since `traitsui.qt4.progress_editor.SimpleEditor.update_editor` restricts values to be positive.
IMHO using 0 as value is a quite normal use case which should be supported.

Monkeypatching from `if self.value:` to `if self.value >= 0:` does the job for me, but i don't know which other effects this could have anywhere else.

Here an example to quickly see the behavior:
```
from traits.api import HasTraits, Int, Button, Str, observe, Float
from traitsui.editors import ProgressEditor


class App(HasTraits):

    percentage_complete = Int()

    view = View(
        Item("inc"),
        Item("dec"),
        Item("res"),
        Item('percentage_complete',
             show_label=False,
             editor=ProgressEditor(min=0, max=100)
             ),
        resizable=True
    )

    inc = Button("Inc")
    dec = Button("Dec")
    res = Button("Reset")

    @observe("inc")
    def incr(self, event):
        self.percentage_complete += 10

    @observe("dec")
    def decr(self, event):
        self.percentage_complete -= 10

    @observe("res")
    def reset(self, event):
        self.percentage_complete = 0


if __name__ == '__main__':
    app = App()
    app.configure_traits()
```

Using traitsui 7.2.1, Win10

corranwebster:
Fixes #1787

This just pushes any sanity checking of values down to the underlying code in Pyface.

**Checklist**
- [x] ~Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~

### DIFF
--- a/traitsui/qt4/progress_editor.py
+++ b/traitsui/qt4/progress_editor.py
@@ -93,9 +93,7 @@ class SimpleEditor(Editor):
         self.progress.min = self.min
         self.progress.max = self.max
         self.progress.message = self.message
-        if self.value:
-            self.progress.update(self.value)
-        return
+        self.progress.update(self.value)
 
     def _min_changed(self):
         self.update_editor()
--- a/traitsui/wx/progress_editor.py
+++ b/traitsui/wx/progress_editor.py
@@ -100,9 +100,7 @@ class SimpleEditor(Editor):
         self.progress.min = self.min
         self.progress.max = self.max
         self.progress.change_message(self.message)
-        if self.value:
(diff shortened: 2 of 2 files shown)


aaronayres35 (approved):
LGTM

Tested using the code from the issue on pyside2 and pyside6 and things worked as expected

Issue #1788, opened 2021-12-15
QtView modal dialog cannot properly use a style_sheet


yves-surrel:
There are two problems when trying to use a style sheet with QtView modal dialogs, as in:

```
from traitsui.qt4.extra.qt_view import QtView
from traitsui.message import Message

style_sheet =  """
* {font: 12pt "Verdana";
	background: #191919;
	color: #DDDDDD;
	border: 1px solid #5A5A5A;	
}
"""

def message(message="hello", title="Message", buttons=["OK"]):
    msg = Message(message=message)
    ui = msg.edit_traits(
        view=QtView(
            ["message~", "|<>"], title=title, buttons=buttons, kind="modal",
        style_sheet= style_sheet)
    )
    return ui
```
First, the style_sheet is not taken into account, and second there is an error on exit:

```
In [2]: message()
Traceback (most recent call last):

  File "<ipython-input-2-c7233cf6135f>", line 1, in <module>
    message()

  File "/Volumes/FAT/Developpement/Python/core/untitled22.py", line 14, in message

  File "/Users/yves/miniconda3/lib/python3.8/site-packages/traits/has_traits.py", line 1736, in edit_traits
    return view.ui(

  File "/Users/yves/miniconda3/lib/python3.8/site-packages/traitsui/qt4/extra/qt_view.py", line 65, in ui
    ui.control.setStyleSheet(self.style_sheet)

AttributeError: 'NoneType' object has no attribute 'setStyleSheet'
```

yves-surrel:
A simple fix would be to add
```
if not ui.control: return ui
```
after the call to `ui()` in `QtView.ui()`.

corranwebster:
The immediate problem here is that this is `kind='modal'` so the call to `ui` blocks until the user clicks OK and then the control flow returns after the ui has been disposed, so no control exists and setting the stylesheet is too late at that point.

So as written, this is class is incompatible with modal `kind` parameters (`modal` and `livemodal`) but should work with others.

I'll make the suggested change so that it does not error, and add a note to the docstring, but I don't see any easy fixes for this.

Remember that you can add stylesheets to groups and items which may allow you to achieve what you want.

corranwebster:
> Seems a hidden feature.

Yes, because there is no equivalent on Wx it isn't strongly advertised.

PR #1825, opened 2022-02-17
QtView shouldn't crash for modal kinds.

corranwebster:
This doesn't mean that stylesheets work, just that there is no crash.  Also update the docstring to indicate why it won't work.

See #1788 - this is only a partial fix.

**Checklist**


aaronayres35 (approved):
LGTM

tested on `main` / this branch and the Error no longer occurs as expected

PR #1827, opened 2022-02-18
Enable the use of Schemas in Views

### LINKED ISSUE #1826: Allow the use of Pyface `Schema` for menubars and toolbars
In Pyface 7.4.0 these have been extracted from Tasks and would make a lot of sense when using TraitsUI declaratively.

Intended usage is something like:
```
View(
     ...
     menubar=MenuBarSchema(
        MenuSchema(
              ...
              id='File',
        ),
     ),
     toolbar=ToolBarSchema(
         ...
     ),
)
```
It should still work to use the regular `MenuBarManager` and `ToolBarManager` classes directly.

corranwebster:
This allows users to use the Pyface `Schema` system instead of using `MenuBarManager`, `ToolBarManager` and similar objects directly.  This permits declarative use of these without needing to have a toolkit available.  Additionally, there are hooks on the `ActionManagerBuilder` to permit extensions, should they be desired.

This is experimental.  There is perhaps an argument that the `ActionManagerBuilder` shouldn't be directly exposed, but instead a factory function expected, as per `TasksActionManagerBuilder`, potentially permitting other controllers.

Fixes #1826.

**Checklist**


traitsui/tests/test_actions.py
```diff
@@ -204,6 +241,93 @@ def _wx_trigger_button_action(ui):
...
+            click_event = wx.CommandEvent(
+                wx.wxEVT_COMMAND_TOOL_CLICKED, control_id
+            )
+
+            # send the event to the toolbar
+            toolbar_control.ProcessEvent(click_event)
+
+        self._test_actions(_wx_trigger_toolbar_action)
+
+    # TODO: I couldn't find a way to press menu items programmatically for wx
```
aaronayres35, line 328: I tested manually and it worked as expected

aaronayres35: I had to add `DialogWithSchema().configure_traits()` in the `if __name__ == "__main__"` below.  Maybe it is worth adding this for more explicit manual testing since there isn't a unittest?

traitsui/tests/test_actions.py
```diff
@@ -204,6 +241,93 @@ def _wx_trigger_button_action(ui):
...
+            toolbar_item_wrapper = toolbar_item._wrappers[0]
+            control_id = toolbar_item_wrapper.control_id
+
+            # build event that clicks the button
+            click_event = wx.CommandEvent(
+                wx.wxEVT_COMMAND_TOOL_CLICKED, control_id
+            )
+
+            # send the event to the toolbar
+            toolbar_control.ProcessEvent(click_event)
```
aaronayres35, line 324: given the duplication maybe this can be pulled out to sit next to `_qt_trigger_action`?

traitsui/tests/test_actions.py
```diff
@@ -149,10 +189,7 @@ def test_qt_button_action(self):
 
     # ----- wx tests
 
-    @unittest.skipIf(
-        not is_mac_os,
-        "Problem with triggering toolbar actions on Linux and Windows. Issue #428.",  # noqa: E501
-    )
+    @unittest.skip("Problem with triggering toolbar actions. Issue #428.")
```
aaronayres35, line 192: why did we remove macOS testing here?

aaronayres35 (approved):
LGTM

just a few comments / questions

traitsui/view.py
```diff
@@ -172,12 +173,18 @@ class View(ViewElement):
     #: The top-level Group object for the view:
     content = Content
 
-    #: The menu bar for the view. Usually requires a custom **handler**:
+    #: The menu bar for the view. Usually requires a custom **handler**.
+    #: This may be either a MenuBarManager or a MenuBarSchema
     menubar = Any  # Instance( pyface.action.MenuBarManager )
```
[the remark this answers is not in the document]
corranwebster:
```suggestion
    menubar = Any
```

traitsui/view.py
```diff
@@ -172,12 +173,18 @@ class View(ViewElement):
...
 
-    #: The toolbar for the view. Usually requires a custom **handler**:
+    #: The toolbar for the view. Usually requires a custom **handler**.
+    #: This may be either a ToolBarManager or a ToolBarSchema
     toolbar = Any  # Instance( pyface.action.ToolBarManager )
```
corranwebster, line 182:
```suggestion
    toolbar = Any
```

traitsui/tests/test_actions.py
```diff
@@ -204,6 +241,93 @@ def _wx_trigger_button_action(ui):
...
+        qt_trigger_menu_action = partial(
+            _qt_trigger_action, pyface.ui.qt4.action.menu_manager._Menu
+        )
+
+        self._test_actions(qt_trigger_menu_action)
+
+    # ----- wx tests
+
+    @unittest.skipIf(
+        not is_mac_os,
```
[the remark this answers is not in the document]
corranwebster: That appears to be the case (my suspicion was that the old way had been broken for some time) - but let me just double check that I didn't accidentally break the old way somehow.

traitsui/tests/test_actions.py
corranwebster: Digging into this a little to see what is going on, and looking at the UI generated, there is a toolbar, but no toolbar buttons show.

corranwebster:
Digging into what was going on with toolbars in Wx ended up with some weird issues where the buttons aren't displaying.  This is likely related to the tests failing.  A days worth of digging hasn't resolved, or even indicated what the problem might be.  I have opened #1843 and I think at this point the plan is:

- push this PR back to the next release
- leave the issue as a known issue and revisit as needed (likely in MayaVi) when there are more concrete use-cases.

corranwebster:
Looked at his a bit more: the issue with Wx toolbars in #1843 is independent of this PR.

PR #1828, opened 2022-02-21
Use Pyface's message dialogs in errors instead of rolling our own

corranwebster:
This allows us to use the same error-handling code cross-platform, modulo a little tweaking for editors which wrap `QLayout` instead of `QWidget`.

This should be transparent to tests - the underlying widgets are the same, we just create them differently.

**Checklist**

### DIFF
--- a/traitsui/wx/editor.py
+++ b/traitsui/wx/editor.py
@@ -61,17 +61,6 @@ class Editor(UIEditor):
         if self.control.GetValue() != new_value:
             self.control.SetValue(new_value)
 
-    def error(self, excp):
-        """Handles an error that occurs while setting the object's trait value."""
-        dlg = wx.MessageDialog(
-            self.control,
-            str(excp),
-            self.description + " value error",
-            wx.OK | wx.ICON_INFORMATION,
-        )
-        dlg.ShowModal()
-        dlg.Destroy()
-
     def set_tooltip_text(self, control, text):
         """Sets the tooltip for a specified control."""
         control.SetToolTip(text)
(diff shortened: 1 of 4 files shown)


traitsui/wx/editor.py
```diff
@@ -61,17 +61,6 @@ def update_editor(self):
         if self.control.GetValue() != new_value:
             self.control.SetValue(new_value)
 
-    def error(self, excp):
-        """Handles an error that occurs while setting the object's trait value."""
-        dlg = wx.MessageDialog(
-            self.control,
-            str(excp),
-            self.description + " value error",
-            wx.OK | wx.ICON_INFORMATION,
```
aaronayres35, line 70: information from pyface will now also have `| wx.STAY_ON_TOP` for the style.

traitsui/wx/editor.py
```diff
@@ -61,17 +61,6 @@ def update_editor(self):
...
-        )
-        dlg.ShowModal()
-        dlg.Destroy()
```
aaronayres35, line 73: pyface also seems to just call `open` on the `MessageDialog` which does https://github.com/enthought/pyface/blob/8adf2a1db26e57cafafd4d2a884d1deff5f35a3d/pyface/i_dialog.py#L149

as opposed to calling wx.MessageDialog methods:
```
dlg.ShowModal()
dlg.Destroy()
```

seems like the right thing to do though.

aaronayres35 (approved):
LGTM

traitsui/wx/editor.py
corranwebster: I think Pyface should be doing the equivalent of that..

traitsui/wx/editor.py
corranwebster: I think that's good in this case.

PR #1833, opened 2022-02-22
Remove Enable and Chaco dependencies

### LINKED ISSUE #1139: Consider replacing chaco and enable with lighter dependency in example
There is a single example in TraitsUI demo that depends on chaco (and enable, which is a dependency of chaco already):
https://github.com/enthought/traitsui/blob/a97287600c64f7e382f5bdfaa0f83271b3c94832/traitsui/examples/demo/Useful/demo_group_size.py#L1-L22

The purpose of the example can be presented without chaco. `chaco` (or actually, `enable`) is a nontrivial dependency: it requires cython and swig etc.

It makes it a fair bit harder to install TraitsUI + its examples. Also it is a substantial barrier for what is actually a basic example about sizing groups and panels.

The example can be rewritten using a different model class.

### LINKED ISSUE #1793: Move integration tests that depend on enable to enable
This is https://github.com/enthought/traitsui/blob/main/integrationtests/ui/table_editor_color_test.py in particular.

corranwebster:
This PR removes the various uses of Enable and Chaco from the TraitsUI codebase.  This simplifies the testing dependencies significantly.

Specifically:

- removes an integration test that was not working and had not been for some considerable time
- moves another integration test to Enable that ensures the Enable ColorTrait values work in TableEditors with ColorColumns (see https://github.com/enthought/enable/pull/892 for the corresponding Enable PR).
- changes an example that used a Chaco plot to demonstrate controlling the size of a group to instead use a `CodeEditor`.
- removes Chaco and Enable as dependencies in the relevant places in tests.

Fixes #1793 and fixes #1139 

**Checklist**


traitsui/examples/demo/Misc/demo_group_size.py
```diff
@@ -20,11 +20,9 @@
 We separate the left and right groups by a splitter (HSplit), and also
```
aaronayres35, line 20: tested the example and it works as expected

aaronayres35 (approved):
LGTM

PR #1836, opened 2022-02-23
Small fix for segfault when non-int values given to `QColor`

corranwebster:
When testing on Python 3.8 with certain versions of PySide6 we were getting a segfault when a tuple with a string value was passed as a color.  This should raise a `TraitError`, not segfault.

**Checklist**

### DIFF
--- a/traitsui/qt4/color_trait.py
+++ b/traitsui/qt4/color_trait.py
@@ -49,10 +49,10 @@ def convert_to_color(object, name, value):
             tup = channels_to_ints(color_table[value])
 
     if isinstance(tup, tuple):
-        if 3 <= len(tup) <= 4:
+        if 3 <= len(tup) <= 4 and all(isinstance(x, int) for x in tup):
             try:
                 color = QtGui.QColor(*tup)
-            except TypeError:
+            except Exception:
                 raise TraitError
         else:
             raise TraitError


traitsui/qt4/color_trait.py
```diff
@@ -49,10 +49,10 @@ def convert_to_color(object, name, value):
             tup = channels_to_ints(color_table[value])
 
     if isinstance(tup, tuple):
-        if 3 <= len(tup) <= 4:
+        if 3 <= len(tup) <= 4 and all(isinstance(x, int) for x in tup):
             try:
                 color = QtGui.QColor(*tup)
-            except TypeError:
+            except Exception:
                 raise TraitError
```
aaronayres35, line 56: note: a test already exists
```
    def test_bad_tuple_not_int(self):
        with self.assertRaises(TraitError):
            ObjectWithColor(color=("not an int", 0xff, 0xff))
```

aaronayres35 (approved):
Code changes LGTM

> ... Python 3.8 with certain versions of PySide6 ...

Do you know which versions specifically?  I was unable to reproduce the segfault locally, but we have a test that exercises this and the code changes are simple enough / make sense

traitsui/qt4/color_trait.py
corranwebster: Yes, it was failure of that test which prompted this PR.

corranwebster: Yes, it was failure of that test which prompted this PR.

corranwebster: Yes, it was failure of that test which prompted this PR.

corranwebster: Yes, it was failure of that test which prompted this PR.

corranwebster:
> Do you know which versions specifically?

The data points I have are that it was working on pip-installed PySide6, but failing on the version from EDM, on Mac OS - didn't check other systems.

corranwebster:
> Do you know which versions specifically?

The data points I have are that it was working on pip-installed PySide6, but failing on the version from EDM, on Mac OS - didn't check other systems.

Issue #1848, opened 2022-03-01
TreeEditor factory callable should support None-type return value


johannesloibl:
I have a use case where the factory trait of a TreeEditor may yield None, which causes my application to crash because **traitsui.qt4.tree_editor.SimpleEditor._menu_new_node** expects factory to return a HasTraits instance.
In my case the factory opens a dialog when the user tries to add a new tree node item and the user may close the dialog and abort the add operation. In this case the factory returns nothing and the app crashes.
Exiting _menu_new_node if new_object is None works fine for me. Could you take this over?

Original:
```
    def _menu_new_node(self, factory, prompt=False):
        """ Adds a new object to the current node.
        """
        try:
            node, object, nid = self._data
        except TypeError:
            raise
        self._data = None
        new_object = factory()
        if (not prompt) or new_object.edit_traits(
            parent=self.control, kind="livemodal"
        ).result:
```

My Version:
```
    def _menu_new_node(self, factory, prompt=False):
        """Adds a new object to the current node."""
        try:
            node, object, nid = self._data
        except TypeError:
            raise
        self._data = None
        new_object = factory()
        if new_object is None:
            # Added return to support None-type returns from factory
            return

        if (not prompt) or new_object.edit_traits(parent=self.control, kind="livemodal").result:
```

pyhannes:
I agree your point. Nevertheless if `None` is a valid value in some situations, this function would crash since edit_traits is called on `None`.

corranwebster:
Good point.

This requires some more thought about what the "right" thing to here is - maybe there needs to be a check for whether the object is `HasTraits`, or adaptation needs to be brought in.

It's also worth noting that if the user cancels the TraitsUI dialog then nothing is added.

The short-term work-around is perhaps to define your own "New" `Action` which calls a method on the appropriate TraitsUI `Handler` subclass - it will get passed the information you need to create a new object as the `info` argument (see https://github.com/enthought/traitsui/blob/4c9601ef523d7ed3f671fcf5284bd4db0d6e70b9/traitsui/qt4/tree_editor.py#L1197-L1200 where the sorting out of what to call and actually calling it is done)

pyhannes:
My short-term workaround is monkey-patching, but thanks for the hint with creating my own New action :)

PR #1849, opened 2022-03-02
TreeEditor factory callable should support None-type return (#1848)

johannesloibl:
Fix for #1848


corranwebster (approved):
Thanks for the PR!

Issue #1839, opened 2022-02-24
Tree editor extended selection mode bug


pyhannes:
This bug occurs when using a TreeEditor with selection_mode="extended", when multiple items are selected holding the CTRL key.
This only occurs if first a group node and then a child node is selected.

Following code shows the issue with traitsui (7.2.1) on Windows, when first **Business** is selected and then **Jason** while holding CTRL.

It's just the TreeEditor_demo.py example of traitsui with extended selection enabled.

```
from traits.api import HasTraits, Str, Regex, List, Instance

from traitsui.api import Item, View, TreeEditor, TreeNode


class Employee(HasTraits):
    """ Defines a company employee. """

    name = Str('<unknown>')
    title = Str()
    phone = Regex(regex=r'\d\d\d-\d\d\d\d')

    def default_title(self):
        self.title = 'Senior Engineer'


class Department(HasTraits):
    """ Defines a department with employees. """

    name = Str('<unknown>')
    employees = List(Employee)


class Company(HasTraits):
    """ Defines a company with departments and employees. """

    name = Str('<unknown>')
    departments = List(Department)
    employees = List(Employee)


# Create an empty view for objects that have no data to display:
no_view = View()

# Define the TreeEditor used to display the hierarchy:
tree_editor = TreeEditor(
    nodes=[
        # The first node specified is the top level one
        TreeNode(
            node_for=[Company],
            auto_open=True,
            # child nodes are
            children='',
            label='name',  # label with Company name
            view=View(['name'])
        ),
        TreeNode(
            node_for=[Company],
            auto_open=True,
            children='departments',
            label='=Departments',  # constant label
            view=no_view,
            add=[Department],
         ),
        TreeNode(
            node_for=[Company],
            auto_open=True,
            children='employees',
            label='=Employees',   # constant label
            view=no_view,
            add=[Employee]
        ),
        TreeNode(
            node_for=[Department],
            auto_open=True,
            children='employees',
            label='name',   # label with Department name
            view=View(['name']),
            add=[Employee]
        ),
        TreeNode(
            node_for=[Employee],
            auto_open=True,
            label='name',   # label with Employee name
            view=View(['name', 'title', 'phone'])
        )
    ],
    selection_mode="extended",
)


class Partner(HasTraits):
    """ Defines a business partner."""

    name = Str('<unknown>')
    company = Instance(Company)

    traits_view = View(
        Item(name='company', editor=tree_editor, show_label=False),
        title='Company Structure',
[...]

PR #1852, opened 2022-03-03
Fix error with multiple selection

### LINKED ISSUE #1850: "selected" trait the tree editor does not contain all selected items
Just saw that due to fix #1846, the "selected" trait of the tree editor does not contain the full list of selected items anymore, since we break out of the loop after the first item. So i think this has to be re-opened?

This code fixes the problem by calculation of the selected variable already before the for loop.
So the bug is fixed and selection still contains all selected items:
```
        selected = []
        if len(nids) > 0:
            selected = [self._get_node_data(nid)[2] for nid in nids]
            for nid in nids:
                # If there is a real selection, get the associated object:
                expanded, node, sel_object = self._get_node_data(nid)

                # Try to inform the node specific handler of the selection, if
```

_Originally posted by @pyhannes in https://github.com/enthought/traitsui/issues/1839#issuecomment-1055793805_

corranwebster:
Fixes #1850 - switch looping strategy from breaking to instead using a flag for the first element.

### DIFF
--- a/traitsui/qt4/tree_editor.py
+++ b/traitsui/qt4/tree_editor.py
@@ -874,22 +874,21 @@ class SimpleEditor(Editor):
         nids = self._tree.selectedItems()
 
         selected = []
+        first = True
         if len(nids) > 0:
             for nid in nids:
                 # If there is a real selection, get the associated object:
-                expanded, node, sel_object = self._get_node_data(nid)
+                expanded, sel_node, sel_object = self._get_node_data(nid)
                 selected.append(sel_object)
 
                 # Try to inform the node specific handler of the selection, if
                 # there are multiple selections, we only care about the first
                 # (or maybe the last makes more sense?)
-
-                # QTreeWidgetItem does not have an equal operator, so use id()
-                if id(nid) == id(nids[0]):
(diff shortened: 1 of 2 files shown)


pyhannes (approved):
Look's good from my view

traitsui/qt4/tree_editor.py
```diff
@@ -874,22 +874,21 @@ def _on_tree_sel_changed(self):
...
         if len(nids) > 0:
             for nid in nids:
                 # If there is a real selection, get the associated object:
-                expanded, node, sel_object = self._get_node_data(nid)
+                expanded, sel_node, sel_object = self._get_node_data(nid)
                 selected.append(sel_object)
 
                 # Try to inform the node specific handler of the selection, if
                 # there are multiple selections, we only care about the first
                 # (or maybe the last makes more sense?)
```
aaronayres35, line 886: nit
```suggestion
```

aaronayres35 (approved):
LGTM, tested this using a modified version of the example code from https://github.com/enthought/traitsui/issues/1839 and the `selected` trait was updated as expected with multiple selections

Issue #1681, opened 2021-06-08
Test sefaults with python 3.10


opoplawski:
Fedora Rawhide has updated to python 3.10.  We are now seeing a test sefault:
```
+ xvfb-run nosetests-3.10 -v
...
test_readonly_editor_value_changed (traitsui.tests.editors.test_image_enum_editor.TestReadOnlyImageEnumEditor) ... ok
test_simple_editor_combobox (traitsui.tests.editors.test_image_enum_editor.TestSimpleImageEnumEditor) ... ERROR
test_simple_editor_more_cols (traitsui.tests.editors.test_image_enum_editor.TestSimpleImageEnumEditor) ... ok
test_simple_editor_popup_editor (traitsui.tests.editors.test_image_enum_editor.TestSimpleImageEnumEditor) ... SKIP: Test requires one of these toolkits: [ToolkitName.wx]
test_custom_editor (traitsui.tests.editors.test_instance_editor.TestInstanceEditor) ... ok
test_custom_editor_resynch_editor (traitsui.tests.editors.test_instance_editor.TestInstanceEditor) ... ok
test_simple_editor (traitsui.tests.editors.test_instance_editor.TestInstanceEditor) ... ERROR
test_simple_editor_modal (traitsui.tests.editors.test_instance_editor.TestInstanceEditor) ... ok
test_simple_editor_parent_closed (traitsui.tests.editors.test_instance_editor.TestInstanceEditor) ... ERROR
test_simple_editor_resynch_editor (traitsui.tests.editors.test_instance_editor.TestInstanceEditor) ... ERROR
test_index_out_of_range (traitsui.tests.editors.test_list_editor.TestCustomListEditor) ... /usr/bin/xvfb-run: line 181: 3741440 Aborted                 (core dumped) DISPLAY=:$SERVERNUM XAUTHORITY=$AUTHFILE "$@" 2>&1
error: Bad exit status from /var/tmp/rpm-tmp.JFiHcK (%check)
```
I'll try to capture a backtrace when I have more time.

corranwebster:
Can you let us know platform (I'm guessing some linux flavour) and backend (Wx, PyQt, PySide) version that you're seeing this with.

opoplawski:
This is Fedora Linux x86_64.  Pretty sure this is using the PyQt backend.

opoplawski:
Running it locally it produces more helpful output:
```
[...]

PR #1862, opened 2022-03-11
Enforce use of int for Qt code.  Final fix for #1681

opoplawski:
**Checklist**


corranwebster (approved):
Thanks - this looks good.

PR #1885, opened 2022-04-21
Allow separators to be added in EnumEditor's dropdown menu

oeyh:
This allows user to add special separator strings (default to empty string) in `values` that is passed to EnumEditor and EnumEditor will translate them to separators if dropdown menu style is used. For Qt backend only.

**Checklist**

### DIFF
--- a/traitsui/editors/enum_editor.py
+++ b/traitsui/editors/enum_editor.py
@@ -15,7 +15,7 @@ traits user interface toolkits.
 import os
 import sys
 
-from traits.api import Any, Range, Enum, Bool
+from traits.api import Any, Range, Enum, Bool, Str
 
 from traitsui.editor_factory import EditorWithListFactory
 from traitsui.toolkit import toolkit_object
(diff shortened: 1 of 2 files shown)


oeyh:
A small demo:
```python
from traits.api import HasTraits, List, Str, Enum, Property
from traitsui.api import Item, View, EnumEditor


class Foo(HasTraits):
    fruits = List(Str)
    diaries = List(Str)
    meat = List(Str)
    items = Enum(values='_items')
    _items = Property

    def _get__items(self):
        return self.fruits + self.diaries + self.meat

    def _fruits_default(self):
        return ['apple', 'orange']

    def _meat_default(self):
        return ['beef', 'chicken', 'pork']

    def _diaries_default(self):
        return ['milk', 'yogurt']

        separator = ''
        return View(
            Item(
                'items',
                style='simple',
                editor=EnumEditor(
                    values=(
                        self.fruits
                        + [separator]
                        + self.diaries
                        + [separator]
                        + self.meat
                    ),
                    use_separator=True,
                ),
            ),
            resizable=True
        )

Foo().configure_traits()
```


image

corranwebster:
Closing and re-opening to trigger CI again.

corranwebster (approved):
Looks good to me.  Thanks for the contribution.

PR #1907, opened 2022-06-21
If `Image` is `None`, use default `paintEvent` handler only in Qt `ImageEditor`

### LINKED ISSUE #1882: `None` value not handled correctly by ImageEditor
In some circumstances I am getting the following traceback (it seems to be on start-up in my code).  This is a Qt-specific issue, and means that there is a code path that should allow for `None` but doesn't.
```
  File "/Users/cwebster/.edm/envs/test-matplotlib/lib/python3.8/site-packages/traitsui/qt4/image_editor.py", line 71, in paintEvent
    pm_size = pixmap.size()
AttributeError: 'NoneType' object has no attribute 'size'
```

Further investigation: it happens when the `IImage` is `None`, I think.

corranwebster:
Fixes #1882.

**Checklist**

### DIFF
--- a/docs/releases/upcoming/1907.bugfix.rst
+++ b/docs/releases/upcoming/1907.bugfix.rst
@@ -1,0 +1,1 @@
+Fix ImageEditor paintEvent when image is None. (#1907)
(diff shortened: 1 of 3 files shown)


aaronayres35 (approved):
change LGTM

However, the modified test does not seem to fail without the fix (it may or may not be worth the time trying to write a new failing test). Downstream where we ran into this issue, it seems to be resolved though so the fix appears to work as expected.

Thank you for this!

PR #1908, opened 2022-06-27
Fixes for Qt6 VideoEditor

### LINKED ISSUE #1905: `VideoEditor` does not work in Qt6
The `QAbstractVideoBuffer` is no longer part of the public QtMultimedia API:

> The back-end API of Qt Multimedia is private in Qt 6. This improves response time for supporting new multimedia use cases. Any classes that contain the words "Control" or "Abstract" in the class name in Qt 5 are now private in Qt 6.

This may mean a significant re-write.

corranwebster:
Get basic `VideoEditor` functionality up and working for Qt6.  There have been considerable changes to the way that QtMultimedia works.

Can't (yet) handle editing the video frames along the way, but it is technically feasible: the problem is that the mechanism is completely different in Qt6 and still needs some investigation.

Fixes #1905.

**Checklist**

### DIFF
--- a/integrationtests/test_all_examples.py
+++ b/integrationtests/test_all_examples.py
@@ -198,16 +198,6 @@ SEARCHER.skip_file_if(
     "Only support wx",
 )
 SEARCHER.skip_file_if(
-    os.path.join(DEMO, "Misc", "demo_group_size.py"),
-    is_qt6,
-    "enable tries to import a missing constant. See enthought/enable#891",
-)
-SEARCHER.skip_file_if(
-    os.path.join(DEMO, "Standard_Editors", "VideoEditor_demo.py"),
-    lambda: not is_qt5(),
-    "Only supported on Qt5",
-)
-SEARCHER.skip_file_if(
     os.path.join(TUTORIALS, "view_multi_object.py"),
     lambda: True,
     "Require wx and is blocking.",
--- a/traitsui/qt4/video_editor.py
+++ b/traitsui/qt4/video_editor.py
@@ -131,7 +131,9 @@ if is_qt5:
                 cloned_frame.width(),
                 cloned_frame.height(),
                 cloned_frame.bytesPerLine(),
-                QVideoFrame.imageFormatFromPixelFormat(cloned_frame.pixelFormat()),
+                QVideoFrame.imageFormatFromPixelFormat(
+                    cloned_frame.pixelFormat()
+                ),
             )
             self.frameAvailable.emit(image)
             return True
@@ -389,7 +391,7 @@ class VideoEditor(Editor):
         if is_qt5:
             self.buffer = self.media_player.bufferStatus()
         else:
-            self.buffer = self.media_player.bufferProgress()
+            self.buffer = int(self.media_player.bufferProgress() * 100)
 
     def _notify_interval_changed_emitted(self, interval):
         self.notify_interval = interval / 1000.0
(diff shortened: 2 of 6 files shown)


integrationtests/test_all_examples.py
```diff
@@ -197,16 +197,6 @@ def get_python_files(self):
     lambda: not is_wx(),
     "Only support wx",
 )
-SEARCHER.skip_file_if(
-    os.path.join(DEMO, "Misc", "demo_group_size.py"),
-    is_qt6,
-    "enable tries to import a missing constant. See enthought/enable#891",
-)
```
jwiggins, line 204: Is this a drive-by fix?

integrationtests/test_all_examples.py
```diff
@@ -197,16 +197,6 @@ def get_python_files(self):
...
-SEARCHER.skip_file_if(
-    os.path.join(DEMO, "Standard_Editors", "VideoEditor_demo.py"),
-    lambda: not is_qt5(),
-    "Only supported on Qt5",
-)
```
jwiggins, line 209: And this goes away because Qt4 is no longer in the matrix?

traitsui/qt4/video_editor.py
```diff
@@ -54,73 +60,83 @@
...
+            self.painter = None
+            self.resizeEvent(None)
+            if image_func is None:
 
-            self.image_func = I_fun
-        else:
-            self.image_func = image_func
+                def I_fun(image, bbox):
+                    # Don't bother with creating an ndarray version
+                    return image, self._np_image
```
jwiggins, line 87: I know it's unrelated to this change, but why not use a lambda here?
```python
# Don't bother with creating an ndarray version
self.image_func = lambda image, bbox: (image, self._np_image)
```

traitsui/qt4/video_editor.py
```diff
@@ -233,10 +251,19 @@ def init(self, parent):
...
 
-        self.media_player = QMediaPlayer(None, QMediaPlayer.VideoSurface)
+        if is_qt5:
+            self.media_player = QMediaPlayer(None, QMediaPlayer.VideoSurface)
+        else:
+            self.media_player = QMediaPlayer()
         self._set_video_url()
         self.media_player.setVideoOutput(self.control)
-        self.media_player.setMuted(self.muted)
+        if not is_qt5:
```
jwiggins, line 260: I would recommend `if is_qt5:` just to keep things consistent with the other conditional paths

traitsui/qt4/video_editor.py
```diff
@@ -233,10 +251,19 @@ def init(self, parent):
...
+            from pyface.qt.QtMultimedia import QAudioOutput
+            self._audio = QAudioOutput()
+            self._audio.setMuted(self.muted)
```
jwiggins, line 263: `Editor` doesn't derive from `HasStrictTraits` I guess?

jwiggins (approved):
Looks "OK" to me.

I can't decide if it's better to just have two implementations of the video player that we can choose between at toolkit resolution time. How do you feel about the various conditional paths scattered about?

integrationtests/test_all_examples.py
corranwebster: Yes, it's a drive-by fix now that Enable understands Qt6.

integrationtests/test_all_examples.py
corranwebster: Yep

traitsui/qt4/video_editor.py
corranwebster: No, almost nothing does in TraitsUI (you're lucky if it doesn't derive from `HasPrivateTraits`). But yes, I should add a trait for this (probably just an `Any` for now).

corranwebster:
> How do you feel about the various conditional paths scattered about?

I'm OK with it - that's why we introduced the `is_qt5` etc. flags.  With two editors we're much more likely to have to fix things twice because something breaks in Qt5 and needs to be copied to the Qt6 version, for example.

PR #1910, opened 2022-06-28
Fix updates for ImageEnumEditor

### LINKED ISSUE #844: QAbstractItemModel no longer has a reset() method (qt4->qt5 update)
Rebuilding of `ImageEnumEditor` fails with an attribute error `'ImageEnumModel' object has no attribute 'reset'` because the `reset` method has been removed from QAbstractItemModel in qt5 (https://doc.qt.io/qt-5/qabstractitemmodel.html).

Documentation of the old `reset` method for reference: https://doc.qt.io/archives/qt-4.8/qabstractitemmodel.html#reset

### LINKED ISSUE #1890: ImageEnumEditor: error when changing value list
I am used to have an Item in my code like this
`Item('Colormap',editor=EnumEditor(name="object.ColormapList"))`
ColormapList is obviously a list of available Colormaps that can be dynamically updated, works totally fine

Lately I switched to ImageEnumEditor like
`Item('Colormap',editor=ImageEnumEditor(name="object.ColormapList", path=CM.get_image_dir()), style='simple')`
As long as ColormapList does not change, everything works fine, but when attempting to add a new element it raises an error (I hid abs paths): 

```Exception occurred in traits notification handler for event object: TraitChangeEvent(object=<multiphysics.Plot3D.V3D.items.dataitem_grid2d.Grid2DDataItem object at 0x0000018F388D14F0>, name='ColormapList', old=['blue-red', 'black-white'], new=['blue-red', 'black-white', 'turbo'])
  File "C:\WinPython-64bit-3.9.8.0\python-3.9.8.amd64\Lib\site-packages\traits\observation\_trait_event_notifier.py", line 122, in __call__
    self.dispatcher(handler, event)
  File "C:\WinPython-64bit-3.9.8.0\python-3.9.8.amd64\Lib\site-packages\traits\trait_notifiers.py", line 48, in ui_dispatch
    handler(*args, **kw)
  File "C:\WinPython-64bit-3.9.8.0\python-3.9.8.amd64\Lib\site-packages\traitsui\qt4\enum_editor.py", line 147, in _values_changed
    self.rebuild_editor()
  File "C:\WinPython-64bit-3.9.8.0\python-3.9.8.amd64\Lib\site-packages\traitsui\qt4\image_enum_editor.py", line 106, in rebuild_editor
    self.control.model().reset()
AttributeError: 'ImageEnumModel' object has no attribute 'reset' 
```
If is set `style='custom'` the widget works as supposed with the custom style, but as I need a dropdown list, I assume I still need to use simple style

System:
windows 10
Python 3.9.8
traitsui-7.2.1
traits-6.3.2

Thanks

corranwebster:
This fixes #844 and fixes #1890.

**Checklist**

### DIFF
--- a/traitsui/qt4/image_enum_editor.py
+++ b/traitsui/qt4/image_enum_editor.py
@@ -99,7 +99,17 @@ class SimpleEditor(BaseEditor, SimpleEnumEditor):
         """Rebuilds the contents of the editor whenever the original factory
         object's **values** trait changes.
         """
-        self.control.model().reset()
+        # This forces the model to completely rebuild itself.
+        self.control.model().endResetModel()
+
+    # Trait change handlers --------------------------------------------------
+
+    def _update_values_and_rebuild_editor(self, event):
+        """Handles the underlying object model's enumeration set or factory's
+        values being changed.
+        """
+        self.control.model().beginResetModel()
+        super()._update_values_and_rebuild_editor(event)
 
 
 class CustomEditor(BaseEditor, CustomEnumEditor):
--- a/traitsui/qt4/image_enum_editor.py
+++ b/traitsui/qt4/image_enum_editor.py
@@ -99,8 +99,8 @@ class SimpleEditor(BaseEditor, SimpleEnumEditor):
         """Rebuilds the contents of the editor whenever the original factory
         object's **values** trait changes.
         """
-        # This forces the model to completely rebuild itself.
-        self.control.model().endResetModel()
+        # unlike other EnumEditor subclasses, this doesn't ever need rebuilding
+        pass
 
     # Trait change handlers --------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


traitsui/qt4/image_enum_editor.py
```diff
@@ -99,7 +99,17 @@ def rebuild_editor(self):
         """Rebuilds the contents of the editor whenever the original factory
         object's **values** trait changes.
         """
-        self.control.model().reset()
+        # This forces the model to completely rebuild itself.
+        self.control.model().endResetModel()
```
itziakos, line 103: I am not sure when the rebuild_editor is called but I feel that the beginResetModel() and endResetModel() should always pair together so maybe they should appear on a try..finally in _update_values_and_rebuild_editor

itziakos (commented):
I am not totally sure about the placement of the endResetModel and beginResetModel calls.

traitsui/qt4/image_enum_editor.py
itziakos: Do we need to call _update_values_and_rebuild_editor during the rebuild_editor?

itziakos: We might want to redsign the _update_values_and_rebuild_editor so that the begin..end reset are called in the rebuild_editor

corranwebster: The `begin...` really needs to happen before the call to `values_changed` to avoid the Qt model being out of sync with the `Editor`.  I'm OK with the way I changed it; I'd be happier if it wasn't in a traits listener, but I can live with it.

itziakos (commented):
PR looks ok

itziakos (approved):

PR #1931, opened 2022-08-30
Fix range text editor bug

### LINKED ISSUE #1923: RangeTextEditor errors out when low or high limit in Range trait is not set
RangeTextEditor checks if value is between low and high limit (PR #1731) but it doesn't consider the case where either low or high is not set (equals to `None`) in the corresponding `Range` trait, which is used to indicate no limit on that end. This change breaks some existing project code.

Wondering if this is an intended change to enforce both low and high limit to be set explicitly. 

A minimum reproducible example:
```python


class Experiment(HasTraits):
    # sigma = Range(low=0, high=5, value=3)  # works
    sigma = Range(low=0, value=3)  # no longer works
    # sigma = Range(high=5, value=3)  # no longer works

        return View(
            Item('sigma', style='text')
        )


if __name__ == '__main__':
    Experiment().configure_traits()

```

ctangell:
Closes #1923.  Makes the qt range text editor accept a none value for either the high or low value. 

**Checklist**

### DIFF
--- a/traitsui/qt4/range_editor.py
+++ b/traitsui/qt4/range_editor.py
@@ -719,8 +719,10 @@ class RangeTextEditor(TextEditor):
         self.sync_value(factory.high_name, "high", "from")
 
         # force value to start in range
-        if not (self.low <= self.value <= self.high):
+        if self.low is not None and self.low > self.value:
             self.value = self.low
+        elif self.high is not None and self.high < self.value:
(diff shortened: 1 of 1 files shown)


corranwebster:
Test failure seems orthogonal to the PR.  I have opened https://github.com/enthought/traitsui/issues/1932 for it and am re-running tests.

corranwebster (approved):
Looks good.

PR #1937, opened 2022-09-16
Fix progress column rendering on linux and windows

### LINKED ISSUE #1936: progress_column skips a row
I tested the demo for the table_editor_with_progress_column but there is an issue with a row skipped so the progress bar is not on the right row, except for the first one. I tested on Linux and Windows and got the same results. This was for:

- Python 3.6
- traitsui 7.4.0
- pyface 7.4.2

I also tried to fix a large `row_height` to the `table editor` and set `auto_size = False`, but it just increases the size of the row and there is still a row skipped.
progress_column

corranwebster:
Should fix #1936.

**Checklist**

### DIFF
--- a/docs/releases/upcoming/1927.bugfix.rst
+++ b/docs/releases/upcoming/1927.bugfix.rst
@@ -1,0 +1,1 @@
+Fix issue with progress column rendering on Linux and Windows. (#1927)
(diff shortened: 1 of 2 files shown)


sallenEnth:
Thank you @corranwebster. That fix my issue.
fix_progress_column

aaronayres35 (approved):
LGTM,
I confirmed the example runs the same and looks correct on both main and this branch on macos

PR #1945, opened 2022-10-03
Fix/more 7 4 1 backports

corranwebster:
It looks like we missed a couple of bugfixes in the previous PR.  This adds them.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -1,6 +1,18 @@
 Traits UI Changelog
 ===================
 
+Release 7.4.1
+-------------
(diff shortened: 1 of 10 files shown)


corranwebster:
Wrong target

rahulporuri:
@corranwebster You don't need to close and reopen a PR to change its target/base. You can do this from the "Edit" button.

image

corranwebster:
Ah - I didn't know that, I only use the dit button for changing titles.  Doesn't surprise me, though.

PR #1950, opened 2022-11-02
Fix RangeTextEditor handling of None for low/high

aaronayres35:
Follow up to https://github.com/enthought/traitsui/pull/1931 as we need that same fix in `update_object` not just `init`.

**Checklist**


corranwebster:
Yep - we need to pin PySide6 to < 6.4

corranwebster (approved):
LGTM

PR #1951, opened 2022-11-03
Pin PySide6 version to <6.4

corranwebster:
Also turn off EDM/Windows/PySide6/Python3.6 tests - can't get an appropriate PySide6 version and combination is tested with pip dependencies elsewhere.

**Checklist**


aaronayres35 (approved):
LGTM

PR #1957, opened 2022-11-04
Release 7.4.2

corranwebster:
This is a combined changelog update and release PR.


setup.py
```diff
@@ -22,7 +22,7 @@
 # into the package source.
 MAJOR = 7
 MINOR = 4
-MICRO = 1
+MICRO = 2
 PRERELEASE = ""
 IS_RELEASED = True
```
aaronayres35, line 27: Do we typically do a follow up PR to toggle IS_RELEASED back to False?  Looks like I forgot after #1943 😅

aaronayres35 (approved):

setup.py
corranwebster: Yes, but it doesn't really matter unless we needed to do a dev build for a client.  We'll probably move to a system like is being used in traits-futures eventually anyway.

PR #1966, opened 2022-11-23
Generate extra `TableEditor` menus dynamically

### LINKED ISSUE #1947: Table Editor Right Click Exception
### Summary

Present on:
- python v3.6
- `traitsui==7.4.0-2`
- `pyside2==5.14.2.3-2`
- Windows `win_x86_64`

Same issue happens on python 3.8 with traitsui and pyside6

If 3 conditions are present on a traitsui view:
- `scrollable` set to `True`
- Anything in the view that has `visible_when` kwarg
- Table editor that is right clickable (reorderable etc..)

Then when that `visible_when` clause is `True` , it causes the table editor to not be right clickable, and the error message shows:
```python
  File "C:\Users\jdragos\.edm\envs\tires\lib\site-packages\traitsui\qt4\table_editor.py", line 1046, in eventFilter
    editor.header_menu_up.setVisible(show_up)
RuntimeError: Internal C++ object (PySide2.QtWidgets.QAction) already deleted.
```

### Example
Right clicking the table initially is fine, but after clicking the button and making the text invisible, right clicking the table causes an error.
The same error occurs if you initially set `show_text_toggle` to `False` to initially have the text invisible too

```python
from traits.api import Bool, Button, HasTraits, Instance, Int, List, Str
from traitsui.api import Item, ObjectColumn, TableEditor, UItem, View


class Row(HasTraits):

    col1 = Int()
    col2 = Str()


class RightClickTable(HasTraits):

    rows = List(Instance(Row))

    invis_text = Str("Invisible!")

    show_text = Button()

    show_text_toggle = Bool(True)

    def _show_text_changed(self):
        self.show_text_toggle = not self.show_text_toggle

    view = View(
        UItem(
            "rows",
            style="custom",
            editor=TableEditor(
                columns=[
                    ObjectColumn(name="col1"),
                    ObjectColumn(name="col2")
                ],
                rows=True,
                reorderable=True,
            ),
        ),
        UItem("show_text"),
        Item("invis_text", style="readonly", visible_when="show_text_toggle"),
        scrollable=True
    )


if __name__ == "__main__":
    rows = [
        Row(col1=1, col2="hi"),
        Row(col1=2, col2="bye"),
    ]
    table_ui = RightClickTable(rows=rows)
    table_ui.configure_traits()
```

corranwebster:
This removes the `empty_menu`, `header_menu` and related traits.  As a result there is a small chance of backwards incompatibility, however:

- there are no uses of these traits found in a quick search of the Enthought codebase
- any code using these traits is likely buggy, as discovered in #1947

The fixes #1947

**Checklist**

### DIFF
--- a/traitsui/qt4/table_editor.py
+++ b/traitsui/qt4/table_editor.py
@@ -578,18 +578,14 @@ class TableEditor(Editor, BaseTableEditor):
         Returns a QMenu or None if no menu to display.
         """
         header_menu = QtGui.QMenu(self.table_view)
-        insertable = self.factory.row_factory is not None
-        editable = self.factory.editable
-        deletable = self.factory.deletable
-        reorderable = self.factory.reorderable
-        if editable:
-            if insertable:
+        if self.factory.editable:
+            if self.factory.row_factory is not None:
                 action = header_menu.addAction("Insert new item")
                 action.triggered.connect(self._on_context_insert)
-            if deletable:
+            if self.factory.deletable:
                 action = header_menu.addAction("Delete item")
                 action.triggered.connect(self._on_context_remove)
-        if reorderable:
+        if self.factory.reorderable:
             show_up = (self.header_row > 0)
             show_down = (self.header_row < self.model.rowCount() - 1)
             if not header_menu.isEmpty() and (show_up or show_down):
--- a/traitsui/qt4/table_editor.py
+++ b/traitsui/qt4/table_editor.py
@@ -141,16 +141,6 @@ class TableEditor(Editor, BaseTableEditor):
     #: The Traits UI associated with the table editor toolbar:
     toolbar_ui = Instance(UI)
 
-    #: The context menu associated with empty space in the table
-    empty_menu = Instance(QtGui.QMenu)
-
-    #: The context menu associated with the vertical header
-    header_menu = Instance(QtGui.QMenu)
-
-    #: The context menu actions for moving rows up and down
-    header_menu_up = Instance(QtGui.QAction)
-    header_menu_down = Instance(QtGui.QAction)
-
     #: The index of the row that was last right clicked on its vertical header
     header_row = Int()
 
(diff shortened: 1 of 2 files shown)


traitsui/qt4/table_editor.py
```diff
@@ -1031,21 +1043,20 @@ def eventFilter(self, obj, event):
...
             if row == -1:
-                factory = editor.factory
-                if factory.row_factory is not None:
-                    editor.empty_menu.exec_(event.globalPos())
+                empty_menu = editor._create_empty_menu()
+                if empty_menu is not None:
+                    event.accept()
+                    empty_menu.exec_(position)
             else:
                 editor.header_row = row
```
jwiggins, line 1055: Is there other code which is dependent on this mutation of `header_row`? Otherwise I would think it's better to pass `row` to `_create_header_menu`.

traitsui/qt4/table_editor.py
```diff
@@ -594,6 +559,53 @@ def _get_image(self, image):
...
+    def _create_header_menu(self):
+        """Create a QMenu to display in the vertical header.
+
+        Returns a QMenu or None if no menu to display.
+        """
+        header_menu = QtGui.QMenu(self.table_view)
+        insertable = self.factory.row_factory is not None
+        editable = self.factory.editable
+        deletable = self.factory.deletable
+        reorderable = self.factory.reorderable
```
jwiggins, line 584: Not sure how I feel about these given they're only referenced once.

jwiggins (approved):

traitsui/qt4/table_editor.py
corranwebster: Yes, unfortunately.  It's used by most of the handler methods for the default menu Actions (`_on_context_insert`, `_on_context_remove`, etc.)

traitsui/qt4/table_editor.py
corranwebster: I think these are there because the code originally came from the deleted code above, like this:
```
if factory.editable and (insertable or factory.deletable):
```
but with added `self.`.  But then I refactored.  It probably makes sense to remove them.

PR #1967, opened 2022-11-23
Display the selected date in the DateEditor in custom style.

### LINKED ISSUE #1942: Qt custom DateEditor should show selected dates when created
The DateEditor always starts showing the calendar page for the current month, even if the current date being displayed is much earlier or later in time.

Default behaviour should be to make sure that the current selection is visible to the user, at least in part.

corranwebster:
This change:

- shows the calendar page with the current value when started
- if multi-select is true, it shows the page with the earliest date
- when the date changes externally in single-select mode it also updates

This fixes #1942

**Checklist**

### DIFF
--- a/docs/releases/upcoming/1967.bugfix.rst
+++ b/docs/releases/upcoming/1967.bugfix.rst
@@ -1,0 +1,1 @@
+Display the selected date in the DateEditor in custom style (#1967)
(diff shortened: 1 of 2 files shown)


jwiggins (approved):
LGTM.

(Is it possible to test this? I see the tests are toolkit-agnostic)

corranwebster:
> (Is it possible to test this? I see the tests are toolkit-agnostic)

The test infrastructure hasn't been extended to the `DateEditor` and it's probably too much to add the testing infrastructure in this PR.

PR #1973, opened 2022-12-15
Protect against deleted nodes in Qt `TreeEditor`

corranwebster:
This change makes checks that the underlying Qt `QTreeViewItem` exists before attempting to access any data stored on it.  This hopefully resolves some issues observed in PySide6 which I suspect were caused by a change in API from returning `AttrbuteError` to `RuntimeError` when accessing that data.

**Checklist**


corranwebster:
I've added issue #1974 for the test segfault on MacOS.

tmreay (approved):
Code changes look good and make sense to me!

PR #1976, opened 2022-12-21
FIX: Replace unnecessary eval() calls with literal_eval()

jwiggins:
Drive-by while looking into `RangeEditor` strangeness.

### DIFF
--- a/traitsui/qt4/check_list_editor.py
+++ b/traitsui/qt4/check_list_editor.py
@@ -26,6 +26,7 @@ user interface toolkit.
 
 
 import logging
+from ast import literal_eval
 
 from pyface.qt import QtCore, QtGui, is_pyside
 
@@ -266,7 +267,7 @@ class TextEditor(BaseTextEditor):
         """Handles the user changing the contents of the edit control."""
         try:
             value = str(self.control.text())
-            value = eval(value)
+            value = literal_eval(value)
         except:
             pass
         try:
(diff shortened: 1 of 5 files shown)


traitsui/qt4/check_list_editor.py
```diff
@@ -266,7 +267,7 @@ def update_object(self, event=None):
         """Handles the user changing the contents of the edit control."""
         try:
             value = str(self.control.text())
-            value = eval(value)
+            value = literal_eval(value)
```
jwiggins, line 270: I'm not 100% sure about this one because I don't have much experience with the editor.

jwiggins:
I don't understand this failure. It doesn't seem related to the changes.
```
  File "D:\a\traitsui\traitsui\.edm\envs\traitsui-test-3.6-pyside2\lib\site-packages\traitsui\testing\tests\test_gui.py", line 152, in test_qt_process_events_process_all
```

corranwebster:
Failure looks to be unrelated, unless the change is causing tests to not clean up after themselves properly.

I'll open an issue.

corranwebster:
Sat down and had a more thorough look at this and I think that `literal_eval` isn't the right thing for a bunch of these since many editors have an `evaluate` trait or method - eg. `BaseRangeEditor` provides this so at least the `RangeEditor` examples should probably replace `eval` with `self.evaluate`.

That feels like the right solution to propagate for most of these.  However a sane default in some of these cases may be `literal_eval` and doing this in some cases may be an API change.

Issue #1924, opened 2022-07-14
Occasional segfaults on OS X video editor tests


corranwebster:
See https://github.com/enthought/traitsui/runs/7226536033?check_suite_focus=true for example:
```
test_video_editor_basics (traitsui.tests.editors.test_video_editor.TestVideoEditor) ... qt.qpa.drawing: Layer-backing is always enabled.  QT_MAC_WANTS_LAYER/_q_mac_wantsLayer has no effect.
qt.qpa.drawing: Layer-backing is always enabled.  QT_MAC_WANTS_LAYER/_q_mac_wantsLayer has no effect.
qt.qpa.drawing: Failed to create QWindow::MetalSurface. Metal is not supported by any of the GPUs in this system.
No MTLDevice
qt.qpa.drawing: Layer-backing is always enabled.  QT_MAC_WANTS_LAYER/_q_mac_wantsLayer has no effect.
qt.qpa.drawing: Layer-backing is always enabled.  QT_MAC_WANTS_LAYER/_q_mac_wantsLayer has no effect.
qt.qpa.drawing: Failed to create QWindow::MetalSurface. Metal is not supported by any of the GPUs in this system.
No MTLDevice
qt.qpa.backingstore.iosurface: Failed to lock QIOSurfaceGraphicsBuffer(0x7fa4f3088580, surface=0x0, size=QSize(0, 16), isLocked=false, isInUse=false) -536870206
Error: The process '/Users/runner/hostedtoolcache/Python/3.6.15/x64/bin/python' failed with exit code null
```

PR #1983, opened 2023-01-12
Update copyright notice in 2023 for non-python files.

corranwebster:
Missed a few copyright updates because I was filtering for Python files only.


corranwebster:
Test failure looks like #1924

mdickinson (approved):
LGTM

Issue #1988, opened 2023-03-09
'InstanceFactoryChoice' object has no attribute 'object'


hopeful0:
I noticed that the `InstanceFactoryChoice` class doesn't work properly when using the selection feature of `InstanceEditor`. The error message is as follows:

```
File ".../python3.9/site-packages/traitsui/qt4/instance_editor.py", line 185, in _get_items
    value.object.observe(
AttributeError: 'InstanceFactoryChoice' object has no attribute 'object'
```

The test examples (instance_editor_test2.py/instance_editor_test4.py/instance_editor_test5.py) in `integrationtests/ui` which use `InstanceFactoryChoice` also trigger this error.

I checked the source code and found that the problem was caused by pull request #1725 which fixed issue #1641. In that pull request, a listener was added for the `object` trait of `InstanceChoiceItem` instance `value` . However, this trait is only defined in `InstanceChoice` and not in other subclasses like `InstanceFactoryChoice`, which dynamically retrieves the `object` through the `get_object()` method.

Therefore, a type check should be added to avoid this problem:

```python
if isinstance(value, InstanceChoice):
    value.object.observe(
        ...
```

PR #1989, opened 2023-03-09
Fix error: 'InstanceFactoryChoice' object has no attribute 'object'

hopeful0:
Fixed the error mentioned in issue #1988.


corranwebster (approved):
Thanks for the contribution!  This looks like it fixes the immediate problem - in the longer term I suspect that the `InstanceChoiceItem` ABC needs to grow a way of signalling that an update is needed to rebuild the items.

PR #1829, opened 2022-02-22
Use Pyface file dialogs instead of creating custom ones

corranwebster:
This has the potential to allow more common code between the two editors, but not in this PR.

No new tests, unfortunately, as we don't have a way to drive the file dialogs once they are open.

**Checklist**
- ~[ ] Add a news fragment if this PR is news-worthy for end users. (see docs/releases/README.rst)~

PR #1993, opened 2023-03-15
Fix regression in wx version of FileEditor

David-Baddeley:
File (and directory editors) are broken in current head (and the last couple of releases). See traceback below.

```
  File "/Users/david/dev/traitsui/traitsui/wx/file_editor.py", line 147, in show_file_dialog
    dlg = self._create_file_dialog()
  File "/Users/david/dev/traitsui/traitsui/wx/file_editor.py", line 209, in _create_file_dialog
    default_path=self._file_name.text(),
AttributeError: 'TextCtrl' object has no attribute 'text'
```

This looks like it can be traced to #1829, which seems to have copy-pasted the `FileDialog` creation code verbatim from the QT version into the wx version. Replacing the `.text()` call with the wx equivalent `.GetValue()` seems to fix things. Thought I'd submit a PR directly seeing as I had a simple fix, but happy to create an issue instead/as well if desired.

**Checklist**

### DIFF
--- a/traitsui/wx/directory_editor.py
+++ b/traitsui/wx/directory_editor.py
@@ -37,7 +37,7 @@ class SimpleEditor(SimpleFileEditor):
         """Creates the correct type of file dialog."""
         dlg = DirectoryDialog(
             parent=self.get_control_widget(),
-            default_path=self._file_name.text(),
+            default_path=self._file_name.GetValue(),
         )
         return dlg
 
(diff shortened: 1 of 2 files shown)


corranwebster (approved):
Thanks for the contribution! This looks good to merge.

greschd:
@corranwebster looking at the change in `file_editor`, I've noticed it also has a `_get_value` method:

https://github.com/enthought/traitsui/blob/86927f5acd3a6e009f80ddeec51c48cf0bc5c6bd/traitsui/wx/file_editor.py#L234-L239

I'm wondering if that should be used instead? Or alternatively, if the `_get_value` method might be unused?

corranwebster:
Possibly - `_get_value` matches the pattern for a Traits property getter, but that may be a coincidence.  I'd need to dig in a bit, which I don't have time for right now - in the middle of getting Python 3.11 support working.

PR #2002, opened 2023-03-22
Remove deprecated `format` trait from `RangeEditor`

corranwebster:
Using it will now raise a trait error.

Also removes from the "extra" BoundsEditor.

**Checklist**


dpinte (approved):
👍

PR #2008, opened 2023-04-12
Be strict about `handler.init()` return values

corranwebster:
Switches a warning to a ValueError. This may break downstream code, but warnings have been in place for a while now.

**Checklist**


mdickinson (approved):
LGTM

PR #2028, opened 2023-06-11
Use QtGui.QStyle.SubControl.SC_ComboBoxArrow

opoplawski:
**Checklist**

This seems to be needed in Fedora Rawhide with PyQt6 6.5.1.


corranwebster:
Thanks for the contribution - PyQt6 migrated from custom Enums to subclasses of Python enum.Enum in a way which broke a lot of code from PyQt5 and are hard to search and replace.  There will be more issues like this.

corranwebster (approved):

PR #2041, opened 2024-01-15
TableEditor: initialise events with is_event=True

chnce:
**Checklist**
- Initialise Events in TableEditor with is_event=True in order to avoid the error below in "sync_value"

```Traceback (most recent call last):
  File "/Users/nino/Repos/traitsui/traitsui/editor.py", line 513, in raise_to_debug
    yield
  File "/Users/nino/Repos/traitsui/traitsui/editor.py", line 398, in sync_value
    editor_value = xgetattr(self, editor_name)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/nino/micromamba/envs/more_pyside/lib/python3.11/site-packages/traits/trait_base.py", line 319, in xgetattr
    return getattr(object, names[-1])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: The dclick trait of a TableEditor instance is an 'event', which is write only.
```

### DIFF
--- a/traitsui/qt/table_editor.py
+++ b/traitsui/qt/table_editor.py
@@ -282,8 +282,8 @@ class TableEditor(Editor, BaseTableEditor):
 
         # Set up the required externally synchronized traits
         is_list = mode in ("rows", "columns", "cells")
-        self.sync_value(factory.click, "click", "to")
-        self.sync_value(factory.dclick, "dclick", "to")
+        self.sync_value(factory.click, "click", "to", is_event=True)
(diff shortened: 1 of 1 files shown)


corranwebster:
Thanks for the error report and contribution. Assuming that the tests pass this looks basically OK.

It looks like there may be a corresponding error in the Wx version of the code: https://github.com/enthought/traitsui/blob/a00f0c34f51b48745d9fc086b2bc31012f4eb599/traitsui/wx/table_editor.py#L279-L282

corranwebster (approved):

